PR #19, opened 2024-05-08
feat(docs): changing the order of tabs in the example code block (#DS-2477)

pimenovoleg:
It is more convenient when the first tab is **HTML** `HTML, TS, CSS`


pimenovoleg:
https://koobiq-next--prs-7a81bb6-3hfvvxd2.web.app/

PR #28, opened 2024-05-15
fix(textarea): title & required attributes behavior (#DS-2449)

NikGurev:
## Summary
Aligned `required` attribute with native html realization

## List of notable changes:
- **updated** `required` attribute with `null` default value to align with native html declaration

### DIFF
--- a/packages/components/textarea/textarea.component.ts
+++ b/packages/components/textarea/textarea.component.ts
@@ -66,8 +66,8 @@ export const KbqTextareaMixinBase: CanUpdateErrorStateCtor & typeof KbqTextareaB
         '[attr.id]': 'id',
         '[attr.placeholder]': 'placeholder',
         '[attr.aria-invalid]': 'errorState',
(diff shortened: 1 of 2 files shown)


packages/components/textarea/textarea.component.ts
```diff
@@ -67,7 +67,7 @@ export const KbqTextareaMixinBase: CanUpdateErrorStateCtor & typeof KbqTextareaB
         '[attr.placeholder]': 'placeholder',
         '[attr.aria-invalid]': 'errorState',
         '[attr.disabled]': 'disabled || null',
-        '[attr.required]': 'required',
```
lskramarov, line 70: лучше так:
```
'[disabled]': 'disabled',
'[required]': 'required',
```

при false тогда аттрибут не будет добавляться

NikGurev: Упростил

PR #93, opened 2024-06-20
feat: added cspell support (#DS-2570)

artembelik:


artembelik:
https://github.com/koobiq/angular-components/pull/94

PR #175, opened 2024-07-29
fix(select,tree-select): item width calculation (#DS-2649)

NikGurev:
## Summary

Fixed item width calculation, so hidden items can be represented correctly.
Fixed bug for specific window size, when 1 item is hidden but "show more" button not represented

### Before

https://github.com/user-attachments/assets/f5cfddd6-ad2c-465b-8669-8aab7eb38b49

### After

https://github.com/user-attachments/assets/465a9cdd-ccf2-438d-a4fe-bcb894bab686

### DIFF
--- a/packages/components/select/select.component.ts
+++ b/packages/components/select/select.component.ts
@@ -961,7 +961,7 @@ export class KbqSelect
         const width: number = parseInt(computedStyle.width as string);
         const marginLeft: number = parseInt(computedStyle.marginLeft as string);
         const marginRight: number = parseInt(computedStyle.marginRight as string);
-        // item width + margins + flex gap between each item
+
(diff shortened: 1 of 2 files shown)


packages/components/select/select.component.ts
```diff
@@ -962,8 +961,8 @@ export class KbqSelect
         const width: number = parseInt(computedStyle.width as string);
         const marginLeft: number = parseInt(computedStyle.marginLeft as string);
         const marginRight: number = parseInt(computedStyle.marginRight as string);
-
-        return width + marginLeft + marginRight;
+        // item width + margins + flex gap between each item
```
artembelik: вроде и без коммента всё понятно

artembelik (approved):

lskramarov (approved):

PR #195, opened 2024-08-06
fix(docs): fixed calendar doubleclicks (#DS-2728)

sk8knut:


packages/docs-examples/components/datepicker/datepicker-overview/datepicker-overview-example.ts
```diff
@@ -10,5 +9,5 @@ import { DateTime } from 'luxon';
     styleUrls: ['datepicker-overview-example.css']
 })
 export class DatepickerOverviewExample {
-    date: DateTime;
+    date = null;
```
NikGurev:
```suggestion
    date: DateTime | null = null;
```

NikGurev (approved):

pimenovoleg (approved):

PR #537, opened 2025-02-19
fix(select): focus frame only for .cdk-keyboard-focused (#DS-2606)

lskramarov:


NikGurev (approved):
Работает!


https://github.com/user-attachments/assets/51baa758-7b57-4f66-a9ab-7dcec65cf955

artembelik (approved):

PR #540, opened 2025-02-20
fix(tabs): missing underline in tab-nav-bar (#DS-3405)

impankratov:
## Summary

tab-nav-bar renders same tab underline indicator (`.kbq-tab-list__active-tab-underline`) as in tab-header, but it misses specific CSS rules which should style & position indicator properly.

## List of notable changes:

Fix is to move styles related to tab underline indicator under `paginated-tab-header` mixin.

closes DS-3405

### DIFF
--- a/packages/components-dev/tabs/template.html
+++ b/packages/components-dev/tabs/template.html
@@ -11,6 +11,19 @@
 
 <hr />
 
+<h3>Nav-bar tabs underlined</h3>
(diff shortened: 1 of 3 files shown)


impankratov:
Don't know if I should, but I've added this specific case to components-dev

before:
image

after:
image

so fix is indeed works.

artembelik (approved):

NikGurev (approved):

PR #598, opened 2025-03-21
fix(dl): does not switch to vertical on initial render (#DS-3542)

impankratov:
## Summary

On initial render, there can be case when kbq-dl width is already less than `minWidth` property.

## List of notable changes:

Also added example case in components-dev to be able to verify that case.

Could be also fixed by changing `resizeStream` to `BehaviorSubject`, but for me `startWith(null)` better informs the intention of needing the initial event.

**Update**: 
Move to standalone in 8655b5da337506a7ae15fe1846a73a3324e61e69 [introduced](https://github.com/impankratov/angular-components/blob/d6378a7b/packages/components-dev/dl/module.ts#L12) "OnPush" strategy in `DlDev` component, thus initial fix stopped working. Had to add `markForCheck` to trigger change detection run and pick up updated `vertical` value.

Also I've noticed that all examples in `DlDev` will be switched to vertical after this fix. Thus tweaked them so they are horizontal and only new example is vertical.

### DIFF
--- a/packages/components-dev/dl/template.html
+++ b/packages/components-dev/dl/template.html
@@ -91,3 +91,10 @@
         authentication. For getting authentication details run the following command
     </kbq-dd>
 </kbq-dl>
+
+<br />
+
+<kbq-dl style="width: 590px" class="docs-border" [minWidth]="600">
+    <kbq-dt>Description</kbq-dt>
+    <kbq-dd>This should be rendered in vertical mode</kbq-dd>
+</kbq-dl>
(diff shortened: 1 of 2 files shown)


lskramarov (approved):

lskramarov:
@impankratov why did you transfer to the draft, are you planning to do something else?

impankratov:
@lskramarov somehow after rebase on current master fix doesn't work. 

Caused by moving to standalone in  8655b5da. `startWith` still works, but Angular doesn't re-render on `this.vertical` change.

Will figure this out, need a little more time.

impankratov:
@lskramarov sorry for delay, fix is ready now

lskramarov (approved):

PR #968, opened 2025-08-11
fix(overflow-items): overflow detection when container has justify-content property (#DS-4013)

artembelik:
- [x] check the work of `alwaysVisible`, apparently now it is ignored and unit tests do not fall


artembelik:
@NikGurev необходима помощь, упал тест - "should enforce the max limit of breadcrumb items displayed async"

NikGurev (approved):

lskramarov (approved):

PR #1081, opened 2025-10-28
fix(tags): deprecated multiple attribute (#DS-4224)

artembelik:


NikGurev (approved):

packages/components/tags/tag-list.component.ts
```diff
@@ -611,7 +625,7 @@ export class KbqTagList
 
         if (this.disabled || isNull(target)) return;
 
-        const _isSelectAll = this.multiple && this.selectable && isSelectAll(event);
+        const _isSelectAll = this.selectable && isSelectAll(event);
```
lskramarov: `_`

lskramarov (approved):

packages/components/tags/tag-list.component.ts
artembelik: поправил

PR #1305, opened 2026-01-22
feat: new search result highlighting style (#DS-4563)

lskramarov:


lskramarov:
реализовано в https://github.com/koobiq/angular-components/pull/1604

PR #1380, opened 2026-02-25
fix(input): block stepper on disabled (#DS-4723)

NikGurev:


packages/components/form-field/stepper.scss
```diff
@@ -22,5 +22,9 @@
     .kbq-stepper-step-down {
         width: 100%;
         text-align: center;
+
+        &[disabled='true'] {
```
artembelik:
```suggestion
        &.kbq-disabled {
```

artembelik (approved):

lskramarov (approved):

PR #1584, opened 2026-05-19
fix(select, tree-select): keyboard focus on single-option list (#DS-4913)

lskramarov:


artembelik (approved):

artembelik:
давай добавим unit или e2e тест

NikGurev (approved):

PR #1592, opened 2026-05-21
chore: update checkout action versions and add permissions (#DS-3557)

NikGurev:
## Summary

  Security hardening for GitHub Actions workflows

  Two categories of changes across 12 workflow files:

  1. Explicit permissions blocks — all workflows previously relied on the repository's default token permissions. Now each workflow declares only what it needs:
  - contents: read added to api, commitlint, license, linters, units
  - contents: read + checks: write added to docs-stable (Firebase deploy posts check runs)
  - permissions: {} added to pr-notify — it uses pull_request_target (elevated trigger) but makes no GitHub API calls, so it needs no token access at all

  2. Action pins to commit SHAs — all actions/checkout references were on floating version tags (@v6.0.2). All are now pinned with the version in a comment. Additionally three third-party actions were unpinned:
  - FirebaseExtended/action-hosting-deploy@v0 → pinned to (# v0.10.0)
  - algolia/algoliasearch-crawler-github-actions@v1.1.13 → pinned
  - mattermost/action-mattermost-notify@v2.1.0 → pinned 

  Also includes a minor fix in commitlint.yml: the PR title is now passed via an environment variable instead of being interpolated directly into the shell command, which prevents potential shell injection.

### DIFF
--- a/.github/workflows/actions/setup-node/action.yml
+++ b/.github/workflows/actions/setup-node/action.yml
@@ -4,7 +4,7 @@ runs:
   using: composite
   steps:
     - name: Setup Node
-      uses: actions/setup-node@v6
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version-file: .nvmrc
         cache: 'yarn'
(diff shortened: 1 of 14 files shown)


lskramarov (approved):

NikGurev:
dependabot supports hash pinned updates

https://github.com/dependabot/dependabot-core/pull/14349

artembelik (approved):

PR #1720, opened 2026-07-10
feat(docs): added AskAI function for docsearch (#DS-4628)

artembelik:


apps/docs/src/styles/_docsearch.scss
```diff
@@ -1,333 +1,6 @@
-/* stylelint-disable scss/at-extend-no-missing-placeholder */
```
NikGurev, line 1: Из apps/docs/src/app/components/navbar/navbar.scss тоже надо убрать кастомизацию стилей, иначе на мобилке некорректно отображается

Image

NikGurev (approved):

apps/docs/src/app/components/docsearch/docsearch.directive.ts
```diff
@@ -31,39 +120,58 @@ const CONFIG: DocsDocsearchProps = {
...
+                    indices: [
+                        { name: 'koobiq', searchParameters: { hitsPerPage: 40, facetFilters: [`lang:${locale}`] } }
+                    ],
+                    transformItems: this.transformItems,
+                    translations: docsIsRuLocale(locale) ? TRANSLATION_RU : {},
+                    askAi: {
+                        assistantId: 'f6ddad78-2e0d-4f11-8cf0-d7e34c354d0e',
+                        agentStudio: true
+                    }
+                });
```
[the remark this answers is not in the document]
artembelik: This is intentional, not a leak.

`docsearch()`'s internal `render(vnode, container)` call reconciles into the same container on every invocation (same as calling legacy `ReactDOM.render` repeatedly on the same root), rather than mounting a new tree. DOM nodes and event listeners are diffed/updated in place, not duplicated. I verified this against the compiled `@docsearch/js@4.6.3` output: the exported `docsearch()` resolves `container` (here `SELECTOR`, the directive's host element — stable across calls) and re-renders into it each time.

Calling `.destroy()` before every re-render would actually be worse: it fully unmounts the modal (loses open/scroll state) and causes a visible flicker on every theme/locale toggle — that's the exact behavior this change was fixing. `.destroy()` is intentionally reserved for the directive's actual teardown (`destroyRef.onDestroy`), not per-emission updates.

apps/docs/src/app/components/docsearch/docsearch.directive.ts
```diff
@@ -1,26 +1,115 @@
...
-const CONFIG: DocsDocsearchProps = {
-    container: SELECTOR,
-    appId: '7N2W9AKEM6',
-    apiKey: '0f0df042e7b349df5cb381e72f268b4d',
-    indexName: 'koobiq',
-    maxResultsPerGroup: 20,
-    disableUserPersonalization: false,
-    resultsFooterComponent: () => null
-} as const;
+type AnyFunction = (...args: any[]) => any;
```
lskramarov: anyFunction use any args and return any result :)

агент не уважает agents.md..

apps/docs/src/app/components/docsearch/docsearch.directive.ts
```diff
@@ -31,39 +120,58 @@ const CONFIG: DocsDocsearchProps = {
...
-                },
-                placeholder: _isRuLocale ? 'Поиск' : 'Search',
-                transformItems: this.transformItems,
-                translations: this.translations(_isRuLocale)
+    private init(): void {
+        combineLatest([
+            this.theme.current.pipe(map((t) => t?.className.replace('kbq-', ''))).pipe(distinctUntilChanged()),
+            this.docsLocaleService.changes.pipe(distinctUntilChanged())
+        ])
+            .pipe(takeUntilDestroyed(this.destroyRef))
```
lskramarov, line 162: два пайпа

lskramarov (approved):

apps/docs/src/app/components/docsearch/docsearch.directive.ts
artembelik: поправил

apps/docs/src/app/components/docsearch/docsearch.directive.ts
artembelik: поправил

apps/docs/src/app/components/docsearch/docsearch.directive.ts
artembelik: откатил, оставил как есть, добавил описание

rmnturov (approved):

PR #1762, opened 2026-07-21
feat(tabs): redesign (#DS-4761)

NikGurev:
## Summary

Tabs underlined variant - visual redesign

#### Layout

- Removed hardcoded `padding: 14px` overrides for underlined tab items; horizontal padding now derives uniformly from `--kbq-tabs-size-tab-item-padding-horizontal` (12px) minus the focus outline width (2px)
- Vertical padding on `.kbq-tab-list__content` tokenised from hardcoded `8px` to `var(--kbq-size-s)`
- `.kbq-tab-header` `border-radius` removed in underlined mode

#### Icon-only tabs

- `margin-inline` reduced globally from `6px` (`--kbq-size-xs`) to `4px` (`--kbq-size-xxs`)
- In underlined mode, icon-only width/height reduced from `32px` to `28px` via new `--kbq-tabs-size-tab-item-underlined-icon-only-size` token (`calc(32px - 4px)`), keeping the total header height at 48px

#### First tab alignment

- Replaced `padding-left: 0` on `:first-child` with `margin-inline-start: -12px` (text tabs) / `0` (icon-only). Negative margin preserves the full padding on the element so the focus ring can extend beyond the container edge, while `padding-left: 0` would have clipped it under `overflow: hidden`

#### Hover underline

- Added `::after` pseudo-element on each underlined text/icon+text tab: 3px bar pinned to the separator, coloured `var(--kbq-line-contrast-fade)` on hover, suppressed on the selected tab (active underline div takes over)
- `left/right` offsets are content-aligned — `calc(padding-horizontal - border-width)` — so hover and active underlines have identical width and position
- Icon-only underlined tabs reuse `::after` for the same hover bar (overriding the non-underlined click-expander role, which for underlined is already covered by `::before`)
- Four new hover underline color tokens added for all filled/transparent × on-background/on-surface combinations

#### Active underline positioning (TypeScript)

- `activeTabOffsetWidth` / `activeTabOffsetLeft` in `KbqTabHeader`: icon-only uses full `offsetWidth` / raw `offsetLeft`; text tabs subtract/add `TAB_PADDING` (12px) to align the underline with the content area, not the padding box
- Same recalculation applied to `KbqTabNavBar` (previously returned raw `offsetWidth`/`offsetLeft` with no adjustment)
- `selectedIndex === 0` special-casing removed — with negative `margin-inline-start` the first tab's `offsetLeft` is negative, so `left + TAB_PADDING` resolves correctly for all positions uniformly


rmnturov (approved):

lskramarov (approved):

packages/components/tabs/__screenshots__/01-light.png
artembelik, line 1: @rmnturov фокус для первого элемента обрезается, выглядит как баг, это ожидаемое поведение?

Image

artembelik (approved):

packages/components/tabs/__screenshots__/01-light.png
NikGurev: исправил дефолтное отображение, обновил скриншот тесты

PR #1781, opened 2026-07-24
feat: add reposition and close scroll strategy (#DS-1920)

NikGurev:
## Summary

Implemented separate scroll strategy that enhances Angular CDK scroll strategies.

- Behavior is simple: hide if originElement is outside of view (cdkScrollable) and reposition otherwise. Angular Cdk `RepositionScrollStrategy` only allows to close when outside of viewport (on `autoClose="true"`)

- It gets origin element from FlexibleConnectedPositionStrategy for built-in components using Angular Overlay, but gives the ability to provide in runtime for custom resolution.

- Strategy intentionally never calls `overlayRef.detach()` itself — hiding should go through the consumer's own `onHide` (e.g. `KbqPopUpTrigger.hide()`), since a direct detach would skip the exit animation/delay and leave `visible`/`instance` state out of sync.


packages/components/core/overlay/auto-hide-scroll-strategy.ts
artembelik, line 1: предлагаю не вводить новые суффиксы для имен файлов - `hide-on-scroll-strategy.ts`

https://angular.dev/style-guide#match-file-names-to-the-typescript-identifier-within

packages/components/core/overlay/hide-on-scroll.strategy.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /** @docs-private */
+    attach(overlayRef: OverlayRef): void {
+        this.overlayRef = overlayRef;
+
+        if (!this.originElement) {
+            // FlexibleConnectedPositionStrategy stores the origin as a private field.
+            // Reading it here avoids requiring callers to pass originElement explicitly.
+            this.originElement = this.coerceOriginElement((overlayRef.getConfig().positionStrategy as any)?._origin);
```
artembelik: без `any`не обойтись?

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,177 @@
...
+        const originRect = originElement.getBoundingClientRect();
+
+        return this.scrollDispatcher
+            .getAncestorScrollContainers(originElement)
+            .some((scrollable) =>
+                this._isOutsideBounds(originRect, scrollable.getElementRef().nativeElement.getBoundingClientRect())
+            );
+    }
+
+    private _isOverlayOutsideViewport(): boolean {
```
artembelik: у cdk есть метод, мб то что тебе нужно - `isElementScrolledOutsideView`

artembelik (approved):

packages/components/core/overlay/hide-on-scroll.strategy.ts
```diff
@@ -0,0 +1,177 @@
...
+ * Scroll strategy that repositions the overlay on scroll and calls the optional `onHide` callback
+ * when the tracked element moves outside its boundary:
+ *
+ * - With `originElement`: tracks the origin against each ancestor `CdkScrollable` container.
+ * - Without `originElement`: tracks the overlay panel against the viewport.
+ *
+ * Pass `onHide` via the factory returned by `kbqHideOnScrollStrategyFactory` so the strategy
+ * calls it when the trigger scrolls out of bounds.
+ */
+export class KbqHideOnScrollStrategy implements ScrollStrategy {
```
artembelik: текущее имя вводит в заблуждение, можно подумать, что стратегия скрывает оверлей при любом скролле, по сути это `RepositionScrollStrategy` + доп проверка на выход за границу видимости

может стоит отнаследовать `RepositionScrollStrategy` и сделать оверрайд метода закрытия?

NikGurev: изучу вопрос

packages/components/core/overlay/hide-on-scroll.strategy.ts
NikGurev: это приватное поле, так что да

UPD, вынес в отдельный интерфейс

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: функция не экспортируется из пакета, к сожалению

packages/components/core/overlay/hide-on-scroll.strategy.ts
NikGurev: UPD. Изучил, `RepositionScrollStrategy` закрытый класс, у него нет protected свойств, `enable` ведет себя абсолютно по-другому. Наследованные от него все равно выльется в полное переписывание логики

`RepositionScrollStrategy` vs `KbqHideOnScrollStrategy`

RepositionScrollStrategy с `autoClose = true` проверяет только выход элемента за границы viewport.
Если элемент вовне, то вызывается overlayRef.detach

А новая стратегия - проверяет выход за границы кастомного контейнера и делает скрытие по заданной в компоненте логике

По поводу имени согласен

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+                const isOutside = this.originElement
+                    ? this._isOriginOutsideAncestors(this.originElement)
+                    : this._isOverlayOutsideViewport();
+
+                if (isOutside) {
+                    this.ngZone.run(() => {
+                        this.hideSubject.next();
+                        this.hooks?.onHide?.();
+                    });
+                }
```
lskramarov, line 110: **No latch — this fires on every scroll tick while the origin stays out of bounds.**

CDK's `RepositionScrollStrategy` calls `this.disable()` **before** `_ngZone.run(...)`, precisely so the hide happens once. That's missing here.

I verified it on a harness: 5 scroll ticks with the origin out of bounds produce **5 `hide` emissions, 5 `onHide` calls, 5 `ngZone.run()` invocations (i.e. 5 change-detection cycles) and 5 `getAncestorScrollContainers()` walks**.

This isn't theoretical for the intended consumer. `KbqPopUpTrigger.hide()` goes through `leaveDelay` and only detaches after the exit animation (`afterHidden()` → `detach()`), so the overlay is **not** detached synchronously — the callback keeps firing every 20 ms for the whole delay window.

```suggestion
                if (isOutside) {
                    this.disable();

                    this.ngZone.run(() => {
                        this.hideSubject.next();
                        this.hooks?.onHide?.();
                    });
                }
```

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+ *   useFactory: kbqAutoHideScrollStrategyFactory
+ * }
+ * ```
+ */
+export function kbqAutoHideScrollStrategyFactory(
+    scrollDispatcher: ScrollDispatcher,
+    viewportRuler: ViewportRuler,
+    ngZone: NgZone
+): (hooks?: KbqScrollStrategyHooks) => KbqAutoHideScrollStrategy {
+    return (hooks?) => new KbqAutoHideScrollStrategy(scrollDispatcher, viewportRuler, ngZone, {}, hooks);
```
lskramarov: **`config` is hard-coded to `{}`, which makes `KbqAutoHideScrollStrategyConfig` unreachable.**

I verified that neither `originElement` nor `scrollThrottle` can be supplied through this — the only public entry point. The PR description says the strategy "gives the ability to provide in runtime for custom resolution", but that isn't reachable via the factory, so the config type is effectively dead API right now.

Looks like `config` should be a factory parameter too, e.g. `(hooks?, config?) => new KbqAutoHideScrollStrategy(scrollDispatcher, viewportRuler, ngZone, config ?? {}, hooks)`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+        private readonly viewportRuler: ViewportRuler,
+        private readonly ngZone: NgZone,
+        private readonly config: KbqAutoHideScrollStrategyConfig = {},
+        private hooks?: KbqScrollStrategyHooks
+    ) {
+        this.originElement = config.originElement ?? null;
+    }
+
+    /** @docs-private */
+    attach(overlayRef: OverlayRef): void {
```
lskramarov, line 65: **The resolved `originElement` is cached once here and never refreshed.**

`attach()` reads `_origin` exactly once. However:

* `FlexibleConnectedPositionStrategy.setOrigin()` is public CDK API and can be called later;
* `KbqPopUpTrigger.setExternalNativeElement()` swaps the anchor at runtime;
* `KbqPopUpTrigger.createOverlay()` returns the already-created `OverlayRef`, so a single strategy instance lives for the trigger's entire lifetime — `attach()` runs once, while `enable()`/`disable()` run per open/close.

As a result a changed origin is never picked up. I'd suggest resolving the origin lazily inside the scroll handler, or re-resolving it in `enable()`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * `FlexibleConnectedPositionStrategy` narrowed to its private `_origin` field, which the
+ * class doesn't expose publicly. `_origin` isn't part of the typed public API surface, so a
+ * future `@angular/cdk` upgrade could rename or remove it without a compile error here —
+ * if that happens, `attach()` below warns in dev mode instead of silently losing the
+ * ancestor-scroll-container tracking.
+ */
+interface PositionStrategyWithOrigin {
+    _origin?: FlexibleConnectedPositionStrategyOrigin;
+}
```
lskramarov: **This comment asserts the opposite of what the CDK types say, and the cast below creates the very risk the comment warns about.**

The doc says `_origin` is a *"private field"*, that the class *"doesn't expose [it] publicly"*, and that it *"isn't part of the typed public API surface, so a future `@angular/cdk` upgrade could rename or remove it without a compile error here."*

For the CDK version in use (20.2.14) that isn't the case — `_origin` is declared as a public member:

```ts
// node_modules/@angular/cdk/overlay-module.d.d.ts:451
/** The origin element against which the overlay will be positioned. */
_origin: FlexibleConnectedPositionStrategyOrigin;
```

I checked, and `positionStrategy._origin` type-checks with no cast at all after the `instanceof` narrowing. So the `as PositionStrategyWithOrigin` cast on line 80 is the only reason a rename in the CDK would **not** be a compile error, and the `isDevMode()` warning covers a hole that wouldn't otherwise exist.

I'd drop the interface, the cast and the warning, and let the compiler handle it. (Yes, `_origin` is underscore-prefixed and the CDK may change it outside a major release — but that's an argument **for** the compile-time check, not against it.)

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+ * - Without `originElement`: tracks the overlay panel against the viewport.
+ *
+ * Pass `onHide` via the factory returned by `kbqAutoHideScrollStrategyFactory` so the strategy
+ * calls it when the trigger scrolls out of bounds.
+ */
+export class KbqAutoHideScrollStrategy implements ScrollStrategy {
+    private readonly hideSubject = new Subject<void>();
+
+    /** Emits when the tracked element scrolls outside its boundary. */
+    readonly hide: Observable<void> = this.hideSubject.asObservable();
```
lskramarov: Two hide mechanisms are exposed and both fire: this observable and `hooks.onHide`. The class docs say to use `onHide` *"instead of relying on external `hide` subscriptions"*, so I'd keep just one.

Related: the `@example` on the factory shows `useFactory: kbqAutoHideScrollStrategyFactory` for `KBQ_POPOVER_SCROLL_STRATEGY`, which is typed `() => ScrollStrategy` — so the factory is invoked with no arguments and `onHide` ends up `undefined`. The example doesn't demonstrate the capability the surrounding prose describes.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+     * Element whose position is tracked against ancestor scroll container boundaries.
+     * When omitted, the overlay panel is checked against the viewport instead.
+     */
+    originElement?: HTMLElement;
+    /** Scroll event throttle in ms. Defaults to 20. */
+    scrollThrottle?: number;
+}
+
+/** Lifecycle hooks passed to a scroll strategy factory. New hooks can be added without breaking existing providers. */
+export interface KbqScrollStrategyHooks {
```
lskramarov: `KbqScrollStrategyHooks` is a very generic name to claim at the core level for a bag that exactly one strategy consumes. Something scoped to this strategy would leave the generic name free.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+        this.hideSubject.complete();
+        this.hooks = undefined;
+        this.overlayRef = null;
+    }
+
+    private _isOriginOutsideAncestors(originElement: HTMLElement): boolean {
+        const originRect = originElement.getBoundingClientRect();
+
+        return this.scrollDispatcher
+            .getAncestorScrollContainers(originElement)
```
lskramarov: `getAncestorScrollContainers()` is recomputed on **every** scroll tick — that's O(registered scrollables × DOM depth) `contains()` calls at roughly 50 Hz. The ancestor set doesn't change while the overlay is open, so it could be computed once in `enable()`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+        this.scrollSubscription = this.scrollDispatcher
+            .scrolled(scrollThrottle)
+            .pipe(
+                filter(
+                    (scrollable) =>
+                        !scrollable || !overlayRef.overlayElement.contains(scrollable.getElementRef().nativeElement)
+                )
+            )
+            .subscribe(() => {
+                overlayRef.updatePosition();
```
lskramarov: `updatePosition()` runs before the bounds check, so an overlay that's about to be hidden gets repositioned first (both do layout reads). Reordering the check ahead of the reposition is enough.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+
+        return this.scrollDispatcher
+            .getAncestorScrollContainers(originElement)
+            .some((scrollable) =>
+                this._isOutsideBounds(originRect, scrollable.getElementRef().nativeElement.getBoundingClientRect())
+            );
+    }
+
+    private _isOverlayOutsideViewport(): boolean {
+        const overlayRect = this.overlayRef!.overlayElement.getBoundingClientRect();
```
lskramarov: The non-null assertion could be avoided by taking the ref as a parameter, the way `_isOriginOutsideAncestors` takes `originElement`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+    ): boolean {
+        return (
+            rect.bottom < containerRect.top ||
+            rect.top > containerRect.bottom ||
+            rect.right < containerRect.left ||
+            rect.left > containerRect.right
+        );
+    }
+
+    private coerceOriginElement(raw?: FlexibleConnectedPositionStrategyOrigin) {
```
lskramarov: Two small things: no explicit return type (`HTMLElement | null`), and in `raw?.nativeElement` the optional chaining is already redundant after `raw instanceof ElementRef`.

Also worth picking one convention for private members — this class mixes `coerceOriginElement` with `_isOriginOutsideAncestors` / `_isOverlayOutsideViewport` / `_isOutsideBounds`. The repo mostly doesn't underscore TS-private members.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /** Unsubscribes from scroll events. */
+    disable(): void {
+        this.scrollSubscription?.unsubscribe();
+        this.scrollSubscription = null;
+    }
+
+    /** Disables the strategy and completes `hide`. */
+    detach(): void {
```
lskramarov, line 123: `detach()` clears `hooks` and `overlayRef` but not `originElement`, so a DOM reference is retained after dispose.

Unrelated but nearby: a second `attach()` silently rebinds (I verified). The CDK throws `getMatScrollStrategyAlreadyAttachedError()` in dev mode for that case.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
```diff
@@ -0,0 +1,206 @@
...
+ * class doesn't expose publicly. `_origin` isn't part of the typed public API surface, so a
+ * future `@angular/cdk` upgrade could rename or remove it without a compile error here —
+ * if that happens, `attach()` below warns in dev mode instead of silently losing the
+ * ancestor-scroll-container tracking.
+ */
+interface PositionStrategyWithOrigin {
+    _origin?: FlexibleConnectedPositionStrategyOrigin;
+}
+
+export interface KbqAutoHideScrollStrategyConfig {
```
lskramarov, line 15: The interface itself has no doc comment (its members do), so it landed in `core.api.md` as `// @public (undocumented)` — which conflicts with the AGENTS.md requirement to document public members. The `/** @docs-private */` on `attach()` isn't a description either.

lskramarov (changes requested):
The idea is right and fills a real gap in the CDK — `RepositionScrollStrategy` carries a literal TODO in its source: *"include all ancestor scroll containers here once we have a way of exposing the trigger element to the scroll strategy."* This PR is exactly that missing piece.

What I liked: filtering out scroll events that originate **inside** the overlay panel is a genuine improvement over the CDK version; `_isOutsideBounds` matches the semantics of the internal `isElementScrolledOutsideView` exactly (and that helper isn't exported, so reimplementing it is justified); zone handling is correct — `updatePosition()` outside the zone, only the emission wrapped in `ngZone.run()`; the `isDevMode()` guard matches the convention used elsewhere in the repo. Tests pass (18/18), eslint and prettier are clean.

Most of my notes are inline. Three of them I'd treat as blocking: the missing latch around the hide emission, the factory hard-coding `config = {}` (which makes `KbqAutoHideScrollStrategyConfig` unreachable), and the origin being cached once in `attach()`. The rest below isn't anchored to specific lines.

---

## The new public API has no consumers

There are zero usages across the branch — the wiring was removed in `feat: reverted popover and pop up changes`. Once released, the API guard freezes this shape. Meanwhile the three blocking points above are exactly the kind of thing that surfaces during the first real integration. Would it make sense to land this together with one consumer (popover/tooltip)?

## Naming

@artembelik's point (the name suggests hiding on any scroll, when it's really Reposition + a boundary check) applies just as much to `KbqAutoHideScrollStrategy` as it did to the previous `KbqHideOnScrollStrategy`. Something like `KbqRepositionOrHideScrollStrategy`, or `KbqRepositionScrollStrategy` with an `autoHide` config flag, would be closer to what it actually does.

## Tests

18 pass, but there are notable gaps:

* **`hooks.onHide` — the headline capability — has no coverage at all.** `buildStrategy` never passes hooks. I checked separately that it works, but the PR doesn't lock that in.
* `it('attach() stores the overlayRef')` asserts `expect(overlayRef.updatePosition).toBeDefined()` — that tests the mock, not the strategy.
* `it('disable() unsubscribes; calling again is a no-op')` never calls `disable()` twice.
* `it('detach() unsubscribes and completes hide$')` only checks completion.
* Not covered: `scrollThrottle` being passed to `scrolled()`, the use of `ngZone.run`, and the dev-mode warning branch. **An "emits exactly once" assertion would have caught the latch issue.**
* `describe('hide$ with origin')` and `completes hide$` refer to a `hide$` property that doesn't exist — it's called `hide`.
* `'uses _origin when it is an HTMLElement'` largely duplicates `'emits when origin scrolls above container'`.

## CI

The E2E job is red. The PR doesn't touch components, so this is most likely baseline drift — but it's worth getting green or explaining before merge.

lskramarov:
хоть какая то замена, пока копилот отдыхает :)

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — `disable()` now runs before the hide callback, so it only fires once per out-of-bounds excursion instead of on every scroll tick (mirrors `RepositionScrollStrategy`, which does the same). Added a regression test with 5 ticks → 1 call.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — the factory is now `(hooks?, config?) => ...` and forwards both through to the strategy.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — origin resolution moved out of `attach()` and now happens at the top of every `enable()`, so it's picked up fresh on each open/close cycle instead of once for the strategy's whole lifetime.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: You're right — confirmed `_origin` is a public field on this CDK version. Dropped the interface, the cast, and the dev-mode warning; the compiler guards it directly now, so a future rename becomes a real compile error.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Dropped the `hide` Observable, keeping `onHide` as the only notification path. Also fixed the factory's `@example` to actually show `onHide`/`config` being passed instead of the token-typed usage that couldn't reach either.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Renamed to `KbqAutoHideScrollStrategyHooks`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — ancestor containers are now computed once per `enable()` and cached, instead of being recomputed on every scroll tick.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Folded into the fix for the repeated-hide bug above — the bounds check now runs first, and `updatePosition()` is skipped entirely on the tick where it hides.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — takes `overlayRef` as a parameter now, same as `isOriginOutsideAncestors` takes `originElement`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — explicit `HTMLElement | null` return type, dropped the redundant optional chaining, and removed the underscore prefix from the private methods to match the rest of the codebase (underscore-prefixed privates here are only used for fields backing legacy public getter/setters).

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed both — `detach()` now also clears `originElement` and the cached ancestor containers, and `attach()` throws in dev mode if called a second time, matching `getMatScrollStrategyAlreadyAttachedError()` in CDK's own `RepositionScrollStrategy`/`CloseScrollStrategy`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: Fixed — added a doc comment directly on `KbqAutoHideScrollStrategyConfig`.

packages/components/core/overlay/auto-hide-scroll-strategy.ts
NikGurev: UPD - hide observable used for external hide handling. Added example for select closing

lskramarov (approved):

PR #1806, opened 2026-07-30
fix(dropdown,autocomplete,filter-bar): remove overlay hosts from layout (#DS-4886)

artembelik:


NikGurev (approved):

packages/components/dropdown/dropdown.component.ts
```diff
@@ -76,6 +76,10 @@ export class KbqDropdownFooter {}
     ],
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
+    host: {
+        // Remove the TemplatePortal host box from layout while keeping kbqDropdownStaticContent in the document flow.
+        '[style.display]': '"contents"'
```
lskramarov: Nit: this is a constant expressed as a binding, so it gets re-checked on every change-detection cycle of the component and buys nothing in return.

Two alternatives:

- **Static host attribute** — `host: { style: 'display: contents' }`. There's already a precedent in the repo (`packages/components/breadcrumbs/roving-focus-group.directive.ts:14`). Applied once at element creation, and the new unit tests keep passing unchanged. This is the smallest improvement.
- **Component SCSS** — `kbq-dropdown { display: contents; }` in `dropdown.scss`. This is what Angular Material does (`mat-menu { display: none }`), and it keeps the value overridable by author CSS. Downside: jest-preset-angular doesn't apply component styles, so both new unit tests would have to go.

One genuine argument for keeping it inline: it lands in the SSR-serialized HTML, so there's no layout flash before the `ViewEncapsulation.None` styles are injected. If that's the reason, worth saying so in the comment — otherwise the static attribute is strictly better than the binding.

packages/components/autocomplete/autocomplete.component.ts
```diff
@@ -86,7 +86,9 @@ export function KBQ_AUTOCOMPLETE_DEFAULT_OPTIONS_FACTORY(): KbqAutocompleteDefau
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     host: {
-        class: 'kbq-autocomplete'
+        class: 'kbq-autocomplete',
+        // Prevent the empty TemplatePortal host from participating in flex/grid layout.
+        '[style.display]': '"none"'
```
lskramarov: Same as on `dropdown.component.ts` — a constant in a host binding. `host: { style: 'display: none' }` (static attribute) applies once at element creation instead of being dirty-checked every CD cycle, and the new spec still passes unchanged.

Nit on the comment wording: it isn't only flex/grid. The host was an inline box in normal flow too, so it also contributed a line box wherever it was declared.

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -93,6 +93,16 @@ describe('KbqDropdown', () => {
         overlayContainer.ngOnDestroy();
     }));
 
+    it('should not create a layout box at the declaration point', () => {
```
lskramarov: Two things about this test:

- The name overstates what `display: contents` does. It removes the *host's own* box, but projected `kbqDropdownStaticContent` still creates boxes at the declaration point — that's the whole reason `contents` was chosen over `none` here. Something like `should remove the host box via display: contents` matches what's actually asserted.
- The assertion reads back the inline style the host binding just wrote, so it can only fail if the binding is deleted. That's a real (if narrow) guard, but it also pins the implementation — reimplementing the fix in `dropdown.scss` would fail this test even though the behavior is identical. jsdom can't do layout, so the actual bug can only be covered by the Playwright baseline; a one-line comment pointing at the navbar snapshot would make that explicit.

packages/components/autocomplete/autocomplete.spec.ts
```diff
@@ -91,6 +91,16 @@ describe('KbqAutocomplete', () => {
         overlayContainer.ngOnDestroy();
     }));
 
+    it('should not create a layout box at the declaration point', () => {
```
lskramarov: Same naming point as the dropdown spec: the body asserts an exact `display` value, so name it for that. Here the name is at least accurate — with `none` nothing is rendered at the declaration point at all.

(`getComputedStyle` as a bare global is fine in specs — `eslint.config.js:456` exempts `**/*.spec.ts` from the SSR global restriction.)

lskramarov (commented):
Reviewed this end-to-end — CI, jsdom behavior, the snapshot delta, and the other places in the repo that share this pattern. The root cause and the fix are right, and the `none` vs `contents` split is correct and well documented. Everything below is non-blocking; #1 and #2 are the ones I'd fold in.

### What I verified

- Pulled both navbar baselines: the diff is exactly a 20 CSS-px leftward shift of all four navbars (40 device-px at 2x DPR) — precisely one `gap: 20px` from the phantom flex item in `packages/components/navbar/e2e.ts:113`. Nothing else moved.
- `app-switcher` is the only consumer of `kbqDropdownStaticContent`, and its `kbq-dropdown` sits directly in `.kbq-app-switcher__app-container` (`display: flex; flex-direction: column`). All 6 of its snapshots are unchanged — good evidence the `contents` path doesn't regress.
- jsdom 26.1 round-trips `display: contents` through `getComputedStyle`, so the new unit tests are meaningful rather than silently passing on a dropped value.
- `display` doesn't affect selector matching, so `:nth-child` / sibling combinators around these hosts are untouched.

### 1. Stale workaround left behind in a docs example

`packages/docs-examples/components/dropdown/dropdown-x-position/dropdown-x-position-example.ts:21`

```html
<kbq-dropdown #appDropdown="kbqDropdown" style="display: none" [xPosition]="xPosition()">
```

The example's host is `layout-column layout-align-center-center layout-gap-s` — a flex column with a gap. That `style="display: none"` *is* this bug, hand-patched at the call site. This PR obsoletes it, and leaving it in ships a doc that teaches the exact pattern that would break anyone using `kbqDropdownStaticContent`. Removing the attribute is a one-line change and renders identically either way.

### 2. A third phantom host is still unfixed

`KbqFilterSavePopover` has the same shape — its whole template is three `<ng-template>`s pushed imperatively into a popover (`packages/components/filter-bar/filter-save-popover.ts:41-61`) — and it's declared as a direct child of `.kbq-filters`, which is `display: flex` (`filters.html:159`, `filters.scss:4`).

It's latent rather than visible today only because `.kbq-filters` has no `gap`, so the zero-width phantom item doesn't show. The same fix applies, and since the component is `@docs-private` the risk is nil. Either fold it in or leave a follow-up — as it stands, the title's "remove overlay hosts from layout" is 2/3 done.

### 3. Host-box styling is now definitively dead — 4 examples quietly rely on it

Four docs examples use `<kbq-dropdown style="max-width: 200px">` (`dropdown-overview`, `dropdown-disabled`, `dropdown-navigation-wrap`, `dropdown-open-by-arrow-down`). These were already no-ops — `max-width` doesn't apply to a non-replaced inline box, and the panel lives in a separate overlay subtree — but under `display: contents` it's provably impossible. Since `panelMaxWidth` exists now, those examples should use it instead of modelling an attribute that does nothing.

More generally: inline styles beat author CSS without `!important`, so consumers can no longer restore a host box from a stylesheet (a template `[style.display]` binding still wins, per Angular's styling precedence). Real-world risk is low — the `@Input('class')` setter already wipes the host `className`, so host-class styling was never supported — but it is a behavior change for every consumer.

### 4. Empty PR description

This changes the rendered layout at every `kbq-dropdown` / `kbq-autocomplete` declaration site in every consuming app. Given the conventional-commit changelog pipeline, a couple of sentences — what broke, what changed, and that host box properties no longer apply — would be worth adding.

### Checked, not an issue

- **`display: contents` a11y-tree bug** — fixed in Chrome 89 / Firefox 61 / Safari 15.4, all well below Angular 20's baseline. The only real host is `<kbq-dropdown role="none">` in app-switcher, which has nothing to lose from the a11y tree anyway.
- **Animations** — `kbqDropdownAnimations` run on `.kbq-dropdown__panel` inside the overlay, unaffected by the host.
- **Coverage** — swept `packages/components` and `packages/components-experimental` for `<ng-template>`-first templates; only these two components match (plus the inline-template case in #2).

lskramarov (approved):

PR #1814, opened 2026-07-31
chore: removed unused styles (#DS-4684)

NikGurev:
## Summary

- Removed _tokens.kbq.import.scss and theming/_components-theming.scss — dead code that computed a     
  $theme map nobody consumed (verified compiled CSS is byte-identical before/after).                     
- Marked everythingh but `kbq-core-theme();` as `@deprecated`.                                                                    
- Simplified light-theme.scss/dark-theme.scss to re-export theme.scss and marked them `@deprecated` (they
  compiled to identical output anyway).                                                                  
- Marked unused params from koobiq-theme()/koobiq-typography() as deprecated.                                       
- Removed dead $tokens/getToken() from common/_tokens.scss and the orphaned default-visual.css export  
  from package.json.                                                                                     
- Marked _tokens.scss `@deprecated` for future removal.                                                  
                                                                                                         
Not breaking: compiled CSS output is unchanged (diffed theme.css/light-theme.css/dark-theme.css and e2e/main.scss).

Use css variables approach for easy theming: https://koobiq.io/en/main/theming/overview


packages/components/core/styles/_koobiq-theme.scss
```diff
@@ -11,7 +11,7 @@
 @use '../forms/forms-theme' as *;
 @use '../highlight/highlight-theme' as *;
 
-@mixin koobiq-theme($theme: null) {
+@mixin koobiq-theme() {
```
lskramarov: **`koobiq-theme()` / `koobiq-typography()`: params removed outright rather than deprecated**

Both mixins are forwarded through the package's public Sass entry (`packages/components/_index.scss` → `core/styles/koobiq-theme`), so this is a public-API signature change, not an internal one. I checked — neither `$theme`, `$tokens`, `$config`, nor `$md-config` was referenced inside either mixin body before this change, so nothing in this repo's own compiled CSS is affected.

The risk is external consumers: anyone still calling `koobiq-theme($theme: $x)` or `koobiq-typography($tokens: $x, ...)` (a plausible carry-over from the old `$theme`-map API) will now hit a hard Sass compile error (`no argument named $theme`) instead of the value being silently ignored as before. Same question applies to `koobiq-typography()` below.

This PR deprecates `kbq-color`/`kbq-palette`/`kbq-contrast` by keeping their signatures and just annotating `@deprecated`. Could these two mixins follow the same pattern for one more major version (keep the now-unused params, just mark them `@deprecated`), or is this intentionally being treated as a breaking change? If the latter, it'd be good to reflect that in the title/changelog, since `yarn run check-api` only guards the TS public API and won't catch this.

packages/components/core/styles/theming/_theming.scss
```diff
@@ -43,224 +44,9 @@
...
-        background: _theme_states-background($tokens, $scheme),
-        foreground: _theme_states-foreground($tokens, $scheme),
-        icon: _theme_states-icon($tokens, $scheme),
-        line: _theme_states-line($tokens, $scheme),
-        // будет удалено, после удаления из компонентов ↓
-        disabled-opacity: map.get($tokens, '#{$scheme}-states-disabled-opacity')
-    );
-}
-
-@function kbq-palette($base-palette, $default) {
```
lskramarov, line 204: **`kbq-palette`: the deprecated definition being kept isn't the one that's actually live**

`_theming.scss` defines `kbq-palette` twice. The one kept in this PR (currently line 10, now marked `@deprecated`) adds `default` plus a `{hue}-contrast` key per hue via `kbq-contrast()`. The one this PR deletes (this line) just adds `default`.

Sass doesn't error on redefining a function under the same name in the same module — the *later* definition silently wins for every caller, including consumers using the module via `@use`. I verified this with `sass` 1.93.3 (this repo's own version): a two-definition repro across a module boundary resolves to the second definition's output, with no warning. So today, before this PR, `kbq-palette` calls actually run this "simple" version, not the "rich" one at line 10 that's being preserved.

`kbq-palette` is forwarded unrestricted through `packages/components/_index.scss`, so it's public API. Keeping line 10 and deleting this one means external callers of `kbq-palette(...)` will silently start getting extra `{hue}-contrast` keys they never got before — a behavior change the "byte-identical compiled CSS" verification in the PR description can't catch, since nothing in this repo's own component styles calls `kbq-palette` anymore either way.

Suggest keeping the definition that's actually been shipping (this one) and dropping the dead one (line 10) instead — that way the eventual removal in the next major version doesn't sneak in a behavior change first.

packages/components/core/styles/_koobiq-theme.scss
NikGurev: Оно 3-4 мажорные версии действует как no-op, есть смысл удалить

UPD: оставил, промаркировал как deprecated.

packages/components/core/styles/theming/_theming.scss
NikGurev: Эта функция использовалась только для построения внутренних структур , ни в одном проекте не вижу ее использования

UPD: промаркировал как deprecated, удалим все разом

packages/components/core/styles/theming/_theming.scss
```diff
@@ -1,12 +1,12 @@
 @use 'sass:map';
 @use 'sass:meta';
 
-@use 'components-theming' as *;
-
+// @deprecated Unused internally. Will be removed in the next major version.
 @function kbq-contrast($palette, $hue) {
     @return map.get(map.get($palette, contrast), $hue);
```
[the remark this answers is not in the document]
NikGurev: marked with `@deprecated`

lskramarov (approved):

PR #1823, opened 2026-08-03
chore(deps)!: resolve all dependabot advisories and add audit checks

lskramarov:
## Summary

Clears every open Dependabot advisory (126 → 0) and adds the checks that keep it that way.

Nothing here ever reached consumers: the published packages declare only `tslib` at runtime, everything else is a peer dependency or build tooling, and the root manifest is `private`. Only two advisories were in the runtime graph at all. The rest was development tooling that had drifted, mostly because `dependabot.yml` ignored **all** major updates, so packages whose only fix shipped in a newer major could never be proposed.

Split into one commit per step so each can be reviewed — and reverted — on its own.

## List of notable changes:

- **updated** `yarn.lock` via `yarn dedupe` (322 duplicated packages collapsed) plus targeted re-resolution of `lodash` and `@messageformat/runtime` within their existing ranges — 126 → 98 advisories, no manifest edits
- **added** a `resolutions` block for transitives their parents hold back — 98 → 8, all 3 criticals gone. The `tar` critical plus 11 more came from `node-gyp` stuck at 10.0.1 behind a stale `npm:latest` descriptor
- **added** `npmAuditExcludePackages` / `npmAuditIgnoreAdvisories` to `.yarnrc.yml` for the two advisories that cannot be fixed by upgrading, each with its reasoning and the condition for removing it
- **added** `audit.yml` — blocks a PR only on advisories that reach the runtime graph (`--environment production --severity high`), so a vulnerable linter cannot block unrelated work
- **added** `audit-report.yml` — weekly full audit that creates or updates a single issue and never fails
- **updated** `dependabot.yml`: the blanket major-version ignore is replaced with a named list of packages tied to the Angular release train; npm interval monthly → weekly; explicit `open-pull-requests-limit`; a group with `applies-to: security-updates`; and `/packages/cli`, whose published manifest the root-only config never updated
- **updated** `security.md`, which was GitHub's own template verbatim and told reporters to email `opensource-security@github.com` about "GitHub-owned repositories" — vulnerability reports were being routed to the wrong organisation
- **updated** the release CLI to current majors: `@octokit/rest` 18 → 22, `inquirer` 7 → 14, `commander` 11 → 15, `chalk` 4 → 6, `glob` 7 → 13, `conventional-changelog` 3 → 8, in both the root devDependencies and the published `packages/cli` manifest
- **updated** `.nvmrc` to 24.16.0 — it pinned 24.11.1 while both manifests require `node >=24.16`, so CI ran a version that did not satisfy the project's own `engines`
- **updated** the eslint global ignores to `**/dist` etc. — bare patterns are anchored to the config's directory, so `packages/cli/dist/cli.js` (a 4 MB minified bundle) was being linted

### No ESM migration was needed

All the new majors are ESM-only, but Node 22.12+ and TypeScript 5.8 both support `require()` of an ES module, so `packages/cli` stays CommonJS. `conventional-changelog` is the only one publishing no `require` condition; it is loaded with a dynamic `import()`, which TypeScript preserves in the CommonJS output under `nodenext` and rollup keeps as `import()` in the bundle. The tsconfigs move to `nodenext` / `bundler` resolution so TypeScript can read the new `exports` maps.

**`packages/cli/src/release/changelog.ts` — this is the code that publishes the library.** `conventional-changelog-writer` 9 dropped Handlebars, so `template.hbs` and `commit.hbs` are now render functions in that file, and the transform receives immutable commits and returns a diff instead of mutating.

It was verified by generating the changelog before and after against the same commit range and diffing the result: **byte for byte identical, 310618 bytes both ways**. Worth a second pair of eyes anyway, since there are no automated tests over the release CLI.

Three behaviour changes worth checking:

- **Breaking, release CLI:** `-n` was declared for `--without-references`, `--without-notification` and `--repo-name` simultaneously. Commander resolved it to the first, leaving the other two unreachable by short flag, and commander 15 rejects the duplicate outright. `-n` keeps its meaning; `--without-notification` is now `-w` and `--repo-name` is now `-r`.
- **glob 13 sort order.** It no longer sorts, treats a backslash as an escape rather than a separator, and returns native separators — which changed the order of the generated example module on Windows only (`empty-state-actions` moved past `empty-state-actions2`, because `\` sorts after `2` while `/` sorts before it). Fixed with an explicit sort plus `posix: true`, applied only to relative patterns: combined with an absolute pattern it returns the `//?/C:/…` extended-length form, which sass cannot open.
- **`brace-expansion` was deliberately not resolved to 5.x.** Its CommonJS build exports an object rather than a function, so `minimatch@3` throws on any `{a,b}` pattern. A plain match still succeeds, so tests would not necessarily have caught it. Documented as an audit exception instead.

### Verified locally

Builds (components, cli, styles, docs-content), 3851 + 164 + 7 tests, `check-api`, eslint, prettier, stylelint, cspell, `validate:license`, and the audit gate. **E2E was not run locally** — the Playwright baselines are Linux with `threshold: 0`, so CI owns that check.

### After merge

- Dismiss the remaining alerts on GitHub with the same reasoning recorded in `.yarnrc.yml`: `xlsx` ×2 as "Vulnerable code is not actually used" (SheetJS left npm after 0.18.5 and both advisories are in the parsing path, while this repo only writes workbooks), `brace-expansion` as "Risk is tolerable to this project".
- Confirm Dependabot security updates are enabled in the repository settings — that is a setting, not a file.
- Expect a wave of major-bump PRs on the first weekly run now that majors are no longer blanket-ignored; it is bounded by `open-pull-requests-limit: 10` and the existing groups.

Deliberately left out: the Playwright 1.55 → 1.62 upgrade, which swaps the bundled Chromium and invalidates every visual snapshot (pinned to `~1.55.1` here, which is all the advisory needed), and the stale `packages/docs-examples/example-module.ts`, which is missing four AG Grid examples that exist on disk.


.github/dependabot.yml
```diff
@@ -4,13 +4,42 @@ updates:
...
+    groups:
+      github-actions:
+        patterns:
+          - '*'
+
   - package-ecosystem: 'npm'
     directory: '/'
     schedule:
-      interval: 'monthly'
+      interval: 'weekly'
```
artembelik, line 15: @lskramarov  осознанное изменение с месяца на неделю?

.github/dependabot.yml
```diff
@@ -4,13 +4,42 @@ updates:
...
+    # Grouping collapses most updates into a handful of pull requests, but the limit keeps
+    # a bad week from filling the queue. Raise it if updates start starving each other.
+    open-pull-requests-limit: 10
```
artembelik: давай не будем ограничивать (независимо от interval), тк добавлен audit воркфлоу
```suggestion
```

packages/cli/src/release/changelog.ts
artembelik, line 1: кажется что это выходит за рамки задачи по добавлению аудита, или я ошибаюсь?

.yarnrc.yml
```diff
@@ -1,2 +1,26 @@
...
+# and CI audit — the matching alert must also be dismissed on GitHub (Security ->
+# Dependabot alerts) with the same reasoning, otherwise the two views disagree.
+npmAuditExcludePackages:
+  # SheetJS stopped publishing to npm after 0.18.5, so a registry-hosted fix will never
+  # exist for GHSA-4r6h-8v6p-xvw6 and GHSA-5pgg-2g8v-p4x9. Both are in the parsing path,
+  # and this repository only ever writes workbooks: see
+  # packages/docs-examples/components/ag-grid/ag-grid-export/ag-grid-export-example.ts,
+  # which imports only { utils, writeFile }.
+  # Drop this exclusion and re-evaluate if anything starts reading .xlsx input.
+  - xlsx
```
artembelik: давай уберем это исключение, пусть проверяется все

packages/cli/src/release/changelog.ts
lskramarov: мажорная версия ченджлога изменилась с 3 на 8 ну и здесь давно ничего не обновляли...

.github/dependabot.yml
lskramarov: да, OpenSSF Scorecard обновляется раз в неделю, нужно как можно быстрее закрыть все что он нашел и потом верну обратно

.yarnrc.yml
lskramarov: этот пакет в npm не обновляется https://www.npmjs.com/package/xlsx?activeTab=versions уже 4 года и судя по всему фикс мы уже не дождемся. ок, еще раз посмотрю..

package.json
```diff
@@ -280,7 +286,7 @@
         "unit:api-gen": "jest --roots tools/api-gen",
         "-----E2E_TESTS-----": "--------------------------------------------------------------------------------------",
         "dev:e2e": "ng serve dev-e2e",
-        "e2e:setup": "playwright install chromium --with-deps && playwright install webkit --with-deps",
```
NikGurev, line 283: В некоторых тестах используется webkit, перепроверь что все работает нормально

NikGurev (approved):

PR #1840, opened 2026-08-04
feat(inline-edit): only overlay for single select (#DS-4870)

NikGurev:
## Summary

Makes the "select-style editor" pattern (dropdown-only edit mode, no bordered field) a built-in `kbq-inline-edit` behavior instead of per-consumer CSS/markup boilerplate.

- **`KbqInlineEdit`**: auto-detects a nested single-value (non-multiple, non-multiline) select via `contentChild`, and when present, hides the bordered field and opens just the dropdown, anchored to the inline-edit host — all built in, no manual wiring or `::ng-deep` CSS required from consumers.
- **`KbqFormField`**: `getConnectedOverlayOrigin()` now consults an optional ancestor-provided `KBQ_CONNECTED_OVERLAY_ORIGIN` (`KbqConnectedOverlayOriginProvider`) before falling back to its own container. This lets an ancestor like `KbqInlineEdit` redirect where any nested control's overlay (select, autocomplete, datepicker, etc.) is anchored — reusing the existing form-field DI mechanism instead of adding new public API to `KbqSelect`.
- `KbqInlineEdit` implements `KbqConnectedOverlayOriginProvider`, returning its own box as the origin only for the select-style-editor case (`isSingleSelect()`); otherwise it returns `undefined` so unrelated controls (input, textarea, autocomplete, datepicker) keep their default form-field-anchored overlay.
- Added docs examples: `inline-edit-select-basic` (vertical + horizontal single-select editors), `inline-edit-select-multiline` (multiline dropdown options); moved the original `inline-edit-select` example under the "Reset"/"Сброс" section. Synced `inline-edit.en.md` with `inline-edit.ru.md`.
- Added unit tests for single-vs-multi-select detection, the overlay-origin override (and that it's skipped for non-select controls), plus a Playwright visual regression test for the multiline example.


packages/components/select/select.component.ts
```diff
@@ -724,6 +725,19 @@ export class KbqSelect
      */
     protected readonly panelMaxHeightToken = computed(() => kbqResolvePanelMaxHeightToken(this.panelMaxHeight()));
 
+    /**
+     * Custom element the panel is positioned and sized against, overriding the default
+     * (the parent `kbq-form-field`'s container, if any, or the trigger itself). Useful when the select
+     * is nested inside another component whose own box better represents the panel's visual anchor
+     * than the select's immediate wrapper. Defaults to the nearest ancestor providing `kbqSelectOrigin`, if any.
+     */
+    readonly connectedTo = input<KbqSelectOrigin | undefined>(
```
lskramarov: новый инпут, паблик интерфейс, все дела... так не нада

Нужно через имеющийся механизм `parentFormField` (возможно чуток подправить), но в целом родитель должен реализовать интефейс для доступа к `getConnectedOverlayOrigin()` и тебе нужно будет только запровайдить его в своем компоненте..

Завтра до синка можно обсудить.

NikGurev: Решил через новый DI токен, который позволяет переопределить origin без нового API в компоненте

lskramarov (approved):

PR #1842, opened 2026-08-04
fix(accordion): allow interactive elements in header and content (#DS-4614)

lskramarov:
Move keydown handling from the accordion host onto each trigger, so buttons, dropdowns and form controls placed beside the trigger or inside the expanded content keep their own keys instead of toggling or navigating the accordion.

Scope the item content query to the accordion that owns each item: a descendant query also matched a nested accordion's items, which sent arrow-key focus into the nested headers and reported nested values through valueChange.

Name the header heading after its trigger, so the labels of the actions beside it no longer leak into the section title announced when navigating by heading.


packages/components-dev/accordion/module.ts
```diff
@@ -1,10 +1,11 @@
 import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
 import { KbqAccordionModule } from '@koobiq/components/accordion';
 import { KbqIconModule } from '@koobiq/components/icon';
+import { AccordionInteractiveElementsExample } from '../../docs-examples/components/accordion';
 
 @Component({
     selector: 'dev-app',
-    imports: [KbqAccordionModule, KbqIconModule],
+    imports: [KbqAccordionModule, KbqIconModule, AccordionInteractiveElementsExample],
```
artembelik: давай сделаем по аналогии с другими dev компонентами, через `DevDocsExamples`

packages/components/accordion/accordion-item.ts
```diff
@@ -33,8 +33,11 @@ export type KbqAccordionItemState = 'open' | 'closed';
     }
 })
 export class KbqAccordionItem implements OnDestroy {
-    /** @docs-private */
-    protected readonly accordion = inject(KbqAccordion);
+    /**
+     * The accordion this item belongs to — always the nearest one, because it is injected.
+     * The accordion reads it back to tell its own items apart from a nested accordion's.
+     */
+    readonly accordion = inject(KbqAccordion);
```
NikGurev, line 40: Должен ли фокус заходить в контент item в закрытом состоянии?

Если нет, можно для `KbqAccordionContentDirective` добавить атрибут [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert)

artembelik (approved):

NikGurev (approved):

PR #1850, opened 2026-08-05
chore(deps-dev): bump @playwright/test from 1.55.0 to 1.62.1 (#DS-5215)

lskramarov:
Resolves GHSA-7mvr-c777-76hp / CVE-2025-59288 (high): browsers were downloaded without verifying the SSL certificate, fixed in 1.55.1. The advisory had been parked in `npmAuditIgnoreAdvisories` precisely because the upgrade could not ride along inside an unrelated change; that entry goes away with it.

1.62.1 rather than the minimal 1.55.1: the screenshot cost is identical either way, since any release past 1.55.0 bumps the bundled Chromium and the baselines are compared with `threshold: 0`. Taking the latest buys seven minor releases for the same single regeneration.

Bundled Chromium moves from 140.0.7339.16 to 151.0.7922.34, so all 215 committed screenshots are regenerated in this pull request via /approve-snapshots.

None of the APIs removed between 1.56 and 1.62 are used here: the suite only touches `test`, `expect`, `Page`, `Locator`, `page.evaluate` and `toHaveScreenshot`. Verified locally against Chromium 151 and WebKit 26.5 — every failure is a screenshot mismatch (expected: the baselines are Linux), with no behavioural regression. In particular the clipboard round-trip in textarea still works under the headless clipboard isolation introduced in 1.62.


NikGurev:
Посмотрел, багов нет

NikGurev (approved):

artembelik (approved):

PR #1854, opened 2026-08-06
fix(timepicker): prevent extra digits when editing incomplete time (#DS-3985)

KamilEmeleev:
## Summary

You could type endless digits into an incomplete value: `11:1` plus digits at the start grew to `911:1`, `9911:1`. Nothing limited how long a time part could be.

## List of notable changes:

- **added** `normalizeTimePart` because the old check only caught big numbers, not long ones like `001`
- **updated** `replaceNumbers` to cut every part down to two digits
- **updated** `onInput` to write the fixed value back while the time is still incomplete, and move the caret to the next part
- **added** unit tests and a Playwright test

- Typing digit by digit still works — `1`, `12:`, `12:3` are left alone
- The caret now jumps on the incomplete path too
- The bad value still hits the DOM for a moment: this fixes the value, it does not block the key

### DIFF
--- a/packages/components/timepicker/timepicker.directive.ts
+++ b/packages/components/timepicker/timepicker.directive.ts
@@ -508,10 +508,12 @@ export class KbqTimepicker<D>
         if (shouldUpdateView) {
             this.setViewValue(nextViewValue);
 
-            this.selectionStart = selectionStart;
-            this.selectionEnd = newTimeObj ? selectionEnd : selectionStart;
+            if (selectionStart !== null) {
+                this.selectionStart = selectionStart;
+                this.selectionEnd = newTimeObj ? selectionEnd : selectionStart;
 
-            this.createSelectionOfTimeComponentInInput((selectionStart as number) + 1);
+                this.createSelectionOfTimeComponentInInput(selectionStart + 1);
+            }
         }
 
         if (!newTimeObj) {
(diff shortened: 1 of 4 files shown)


packages/components/timepicker/timepicker.directive.ts
```diff
@@ -494,6 +497,23 @@ export class KbqTimepicker<D>
...
+        // the extra digits stay in the input and the value grows unbounded.
+        const shouldUpdateView = !!newTimeObj || nextViewValue !== this.viewValue;
+
+        if (shouldUpdateView) {
+            this.setViewValue(nextViewValue);
+
+            this.selectionStart = selectionStart;
+            this.selectionEnd = newTimeObj ? selectionEnd : selectionStart;
+
+            this.createSelectionOfTimeComponentInInput((selectionStart as number) + 1);
```
NikGurev: selectionStart может быть null, нигде выше это не проверяется, надо добавить проверку

artembelik:
редеплой сообщение с ссылкой на доку не показал, [вот адрес](https://github.com/koobiq/angular-components/actions/runs/31103445339/job/92622495298#step:8:47)

artembelik (approved):

NikGurev (approved):

lskramarov (approved):

PR #1907, opened 2026-08-17
chore(schematics): document the scrollbar-deprecated-path migration

lskramarov:
## Summary

`packages/schematics/src/migrations/scrollbar-deprecated-path/` was the only migration of 20 without a `README.md`, so it shipped to consumers undocumented while all its siblings ship a guide.

The missing file turned out to be more than a documentation gap. `packages/schematics/scripts/copy-meta-to-dist.js` copies each migration's README into the published package, its loop is sequential and rethrows, and the final rejection was swallowed — so the copy step aborted at the first missing file and the build still exited 0. Verified locally: with the README absent, `dist/components/schematics/migrations/v20-upgrade/index.js` and `dist/components/schematics/utils/messages.js` were never copied, yet `yarn run build:schematics` reported success. `rimraf packages/schematics/dist` does not clear `dist/components`, so a stale local build masks the gap while CI publishes the truncated package.

## List of notable changes:

- **added** `README.md` for the `scrollbar-deprecated-path` migration, following the structure of the existing ones — what the schematic rewrites, which specifiers the quote-anchored match deliberately skips, the `fix` default, and a before/after example.
- **documented** the silent part of the entry-point split: `@koobiq/components/scrollbar` and `@koobiq/components/scrollbar/deprecated` share only `KbqScrollbar` and `KbqScrollbarOptions`, so most unmigrated imports fail with `TS2305`, while `<kbq-scrollbar>` keeps compiling and rendering against the new component and the `[kbq-scrollbar]` attribute form of the selector quietly matches nothing.
- **fixed** `copy-meta-to-dist.js` so a copy failure fails the build: the catch now sets `process.exitCode = 1`, failures are collected instead of aborting the loop at the first one, and a missing `README.md` / `schema.json` is reported as `No README.md for the <migration> migration` instead of a raw ENOENT path.

Verified with `yarn run build:schematics` on a cleared `dist/components` (exit 0, all 20 migrations complete), a negative run with `alert-signals/README.md` and `v20-upgrade/schema.json` removed (exit 1, both reported by name, everything else still copied), and `npx jest packages/schematics` (21 suites, 275 tests).

- Whether the README's account of the behaviour change is what consumers should be told — in particular the claim that repointing the import preserves current rendering exactly, and that adopting the new scrollbar is a separate manual step.
- Whether `README.md` and `schema.json` should be hard requirements for every migration directory. All 20 provide both today, so the new check is green, but it does mean a future migration cannot land without its README.

### DIFF
--- a/packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
+++ b/packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
@@ -13,7 +13,7 @@ component together with `KbqScrollbarDirective` — moved to
 
 Nothing about that implementation changed: `initializationTarget`, `options`, `events`, `defer`,
 `scrollbarInstance`, the `onInitialize` / `onUpdate` / `onDestroy` / `onScroll` outputs and the
-`kbq-scrollbar` element selector all behave exactly as before at the new path. Repointing the
+`kbq-scrollbar` element selector all behave exactly as before at the new path. Rewriting the
 import is the whole migration — adopting the new scrollbar is a separate, manual step.
 
 ## Behaviour change
--- a/packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
+++ b/packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
@@ -42,12 +42,12 @@ the module specifier `@koobiq/components/scrollbar` to `@koobiq/components/scrol
 The match is quote-anchored (`(['"])@koobiq/components/scrollbar\1`), so only the exact, bare
 specifier is taken, in the quote style it was written in:
 
-| Specifier                                   | Result                                      |
-| ------------------------------------------- | ------------------------------------------- |
-| `'@koobiq/components/scrollbar'`            | rewritten, still single-quoted              |
-| `"@koobiq/components/scrollbar"`            | rewritten, still double-quoted              |
-| `'@koobiq/components/scrollbar/deprecated'` | untouched — the migration is idempotent     |
-| `'@koobiq/components/scrollbar-x'`          | untouched — a sibling package, not a prefix |
+| Specifier                                         | Result                                                                                 |
+| ------------------------------------------------- | -------------------------------------------------------------------------------------- |
+| `'@koobiq/components/scrollbar'`                  | rewritten, still single-quoted                                                         |
+| `"@koobiq/components/scrollbar"`                  | rewritten, still double-quoted                                                         |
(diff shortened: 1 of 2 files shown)


packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+The match is quote-anchored (`(['"])@koobiq/components/scrollbar\1`), so only the exact, bare
+specifier is taken, in the quote style it was written in:
+
+| Specifier                                   | Result                                      |
+| ------------------------------------------- | ------------------------------------------- |
+| `'@koobiq/components/scrollbar'`            | rewritten, still single-quoted              |
+| `"@koobiq/components/scrollbar"`            | rewritten, still double-quoted              |
+| `'@koobiq/components/scrollbar/deprecated'` | untouched — the migration is idempotent     |
+| `'@koobiq/components/scrollbar-x'`          | untouched — a sibling package, not a prefix |
```
artembelik: `@koobiq/components/scrollbar-x` такого пакета (пути) у нас нет

artembelik (approved):

packages/schematics/src/migrations/scrollbar-deprecated-path/README.md
lskramarov: Да, такого пути нет — пример вымышленный намеренно, это негативный кейс для якоря в регулярке.

Замена текстовая, а не по AST:

```ts
const FROM = `(['"])@koobiq/components/scrollbar\1`;
const TO = `$1@koobiq/components/scrollbar/deprecated$1`;
```

Обратная ссылка `\1` требует закрывающую кавычку сразу после `scrollbar`. Без неё совпадение стало бы префиксным, и любой более длинный спецификатор с тем же началом переписался бы в середину: `'@koobiq/components/scrollbar-x'` → `'@koobiq/components/scrollbar/deprecated-x'`. Строка таблицы фиксирует, что этого не происходит; ей соответствует тест `does not touch an unrelated sibling package whose name merely starts with "scrollbar"` в `index.spec.ts`.

Но формулировка «a sibling package» действительно читается так, будто путь у нас есть. Предлагаю убрать это слово и пометить пример как гипотетический:

| `'@koobiq/components/scrollbar-x'` (hypothetical) | untouched — the closing quote is required, so a longer specifier is not a prefix match |

PR #1910, opened 2026-08-17
fix(tree): prevent tooltip for non-truncated option text (#DS-4931)

KamilEmeleev:
## Summary

Tree options showed a tooltip even when the text was not truncated. `KbqTreeOption` exposed only
`parentTextElement`, so `KbqTitleDirective` measured the `.kbq-option-text` container against the
option host — which is always bigger, so every option looked overflown.

Reported on `kbq-username`, but it is unchanged and correct: plain-text options were affected too.

## List of notable changes:

- **fixed** tooltip without truncation, by exposing `textElement` so the clipping box is measured against itself
- **added** two unit tests: no tooltip when the text fits, tooltip when it is clipped

Content marking its own `#kbqTitleText` keeps priority, so tree-select and filter-bar should be unaffected.

Trade-off: with `parent === child` a sub-pixel clip is no longer reported for tree options that do not
mark their own `#kbqTitleText`. Previously that case reported truncation unconditionally, so it goes
from always shown to never shown. Content marking its own text keeps both the sub-pixel branch and the
`hasEllipsis` check intact.

### DIFF
--- a/packages/components/tree/tree-option.component.ts
+++ b/packages/components/tree/tree-option.component.ts
@@ -108,8 +108,12 @@ export class KbqTreeOption extends KbqTreeNode<KbqTreeOption> implements AfterCo
     preventBlur: boolean = false;
 
     @ViewChild('kbqTitleContainer') parentTextElement: ElementRef;
-    /** Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself. */
-    @ViewChild('kbqTitleContainer') textElement: ElementRef;
+
+    // Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself.
+    get textElement(): ElementRef {
+        return this.parentTextElement;
+    }
+
     readonly toggleElementDirective = contentChild(KbqTreeNodeToggleDirective);
     readonly toggleElementComponent = contentChild(KbqTreeNodeToggleComponent);
     readonly pseudoCheckbox = contentChild(KbqPseudoCheckbox);
--- a/packages/components/tree/tree-option.component.ts
+++ b/packages/components/tree/tree-option.component.ts
@@ -108,6 +108,8 @@ export class KbqTreeOption extends KbqTreeNode<KbqTreeOption> implements AfterCo
     preventBlur: boolean = false;
 
     @ViewChild('kbqTitleContainer') parentTextElement: ElementRef;
+    /** Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself. */
+    @ViewChild('kbqTitleContainer') textElement: ElementRef;
     readonly toggleElementDirective = contentChild(KbqTreeNodeToggleDirective);
     readonly toggleElementComponent = contentChild(KbqTreeNodeToggleComponent);
     readonly pseudoCheckbox = contentChild(KbqPseudoCheckbox);
--- a/packages/components/tree/tree-selection.component.spec.ts
+++ b/packages/components/tree/tree-selection.component.spec.ts
@@ -1582,14 +1582,14 @@ describe('KbqTreeSelection', () => {
     });
 
     describe('kbq-title', () => {
-        let fixture: ComponentFixture<KbqTreeAppWithTitle>;
+        let fixture: ComponentFixture<TreeAppWithTitle>;
         let titleDirective: KbqTitleDirective;
         let optionElement: HTMLElement;
         let containerElement: HTMLElement;
 
         beforeEach(() => {
             configureKbqTreeTestingModule();
-            fixture = TestBed.createComponent(KbqTreeAppWithTitle);
+            fixture = TestBed.createComponent(TreeAppWithTitle);
             fixture.detectChanges();
 
             const optionDebugElement = fixture.debugElement.query(By.directive(KbqTitleDirective));
@@ -1619,6 +1619,46 @@ describe('KbqTreeSelection', () => {
             jest.spyOn(containerElement, 'scrollHeight', 'get').mockReturnValue(20);
 
             expect(titleDirective.isOverflown).toBe(true);
+        });
+
+        it('should report overflow when the container is clipped vertically', () => {
+            // Two-line options (`.kbq-option-caption`) clamp vertically, so the height check must stay a
+            // real signal and not just mirror the width one.
+            jest.spyOn(containerElement, 'offsetWidth', 'get').mockReturnValue(232);
+            jest.spyOn(containerElement, 'scrollWidth', 'get').mockReturnValue(232);
+            jest.spyOn(containerElement, 'offsetHeight', 'get').mockReturnValue(20);
+            jest.spyOn(containerElement, 'scrollHeight', 'get').mockReturnValue(40);
+
+            expect(titleDirective.isOverflown).toBe(true);
+        });
+    });
+
+    describe('kbq-title with projected #kbqTitleText', () => {
+        let fixture: ComponentFixture<TreeAppWithTitleText>;
+        let titleDirective: KbqTitleDirective;
+        let containerElement: HTMLElement;
+        let textElement: HTMLElement;
+
+        beforeEach(() => {
+            configureKbqTreeTestingModule();
+            fixture = TestBed.createComponent(TreeAppWithTitleText);
+            fixture.detectChanges();
+
+            const optionDebugElement = fixture.debugElement.query(By.directive(KbqTitleDirective));
+
+            titleDirective = optionDebugElement.injector.get(KbqTitleDirective);
+            containerElement = optionDebugElement.nativeElement.querySelector('.kbq-option-text')!;
+            textElement = optionDebugElement.nativeElement.querySelector('.projected-text')!;
+        });
+
+        it('should measure the projected text element, not the option container', () => {
+            // Projected `#kbqTitleText` wins over the `KbqTitleTextRef` fallback, so the container stays the
+            // measured parent while the child is the projected span. Consumers relying on this must keep working.
+            jest.spyOn(containerElement, 'offsetWidth', 'get').mockReturnValue(200);
+            jest.spyOn(containerElement, 'scrollWidth', 'get').mockReturnValue(200);
+            jest.spyOn(textElement, 'scrollWidth', 'get').mockReturnValue(400);
+
+            expect(titleDirective.isHorizontalOverflown).toBe(true);
         });
     });
 });
@@ -1850,7 +1890,22 @@ abstract class TreeParams {
         </kbq-tree-selection>
     `
 })
-class KbqTreeAppWithTitle extends TreeParams {}
+class TreeAppWithTitle extends TreeParams {}
+
+@Component({
+    imports: [
+        KbqTreeModule,
+        KbqTitleModule
+    ],
+    template: `
+        <kbq-tree-selection [dataSource]="dataSource" [treeControl]="treeControl">
+            <kbq-tree-option *kbqTreeNodeDef="let node" kbq-title kbqTreeNodePadding>
+                <span #kbqTitleText class="projected-text">{{ node.name }}</span>
+            </kbq-tree-option>
+        </kbq-tree-selection>
+    `
+})
+class TreeAppWithTitleText extends TreeParams {}
 
 @Component({
     imports: [KbqTreeModule, FormsModule],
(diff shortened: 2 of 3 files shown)


NikGurev (approved):
Pointing both parentTextElement and textElement at the same #kbqTitleContainer isn't a mistake — it's parent.offsetWidth < child.scrollWidth degenerating into el.offsetWidth < el.scrollWidth on one element, which is exactly the check needed here.

artembelik:
https://koobiq-next--prs-1910-ab389bc7.web.app

packages/components/tree/tree-option.component.ts
```diff
@@ -108,6 +108,8 @@ export class KbqTreeOption extends KbqTreeNode<KbqTreeOption> implements AfterCo
     preventBlur: boolean = false;
 
     @ViewChild('kbqTitleContainer') parentTextElement: ElementRef;
+    /** Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself. */
+    @ViewChild('kbqTitleContainer') textElement: ElementRef;
```
lskramarov: **Sub-pixel truncation is no longer detected for tree options.**

With `parent === child`, the "fraction of a pixel" branch of `isOverflown` (`title.directive.ts:96-121`) can never return `true`:

- `.kbq-option-text` is a shrinkable flex item (`width: 100%`, `overflow: hidden`, `text-overflow: ellipsis`), so its used width is fractional.
- When the box is 232.4px and the text 232.45px, both `offsetWidth` and `scrollWidth` round to 232 → `parent.offsetWidth === child.scrollWidth` → the branch is taken.
- `hasOnlyText` is `false` for a tree option (measured: the host view has 4 child nodes — 2 comments, the span, 1 comment), so it falls through to `isWidthOverflown(this.parent.getBoundingClientRect().width, this.child.getBoundingClientRect().width)`.
- Both reads hit the same node, so the comparison is `w < w` — always `false`.

Confirmed in the suite: with a computed `text-overflow: ellipsis` and `offsetWidth === scrollWidth === 232`, `titleDirective.isOverflown` returns `false`. The browser paints `…`, no tooltip opens.

So `isWidthOverflown` — and the `hasEllipsis` check written specifically for `KbqTreeOption` — become dead code here. Still a large net improvement over "always truncated", but the sub-pixel case goes from *always shown* to *never shown*.

Two ways out: an explicit early return when `parent === child`, or a real `#kbqTitleText` span inside `.kbq-option-text` so the two measurements stay distinct — which is what every other `KbqTitleTextRef` host does (`option.html`, `dropdown-item.html`, `list-option.html`, `button.component.html`).

lskramarov: **The non-truncated case now takes the expensive measurement path.**

Because `parent` and `child` are the same node, `offsetWidth === scrollWidth` holds for *every* option whose text fits — so the common case always enters the sub-pixel branch instead of returning from the two integer reads above it.

Measured on a plain tree option with `offsetWidth === scrollWidth === 200`: `isOverflown` performs **2 `getBoundingClientRect()` and 2 `getComputedStyle()` calls** before returning `false`. Both force a layout/style flush, and `isOverflown` runs on every `mouseenter` (once per hovered option), on each throttled `ContentObserver` emission, and on every debounced resize.

Before this change the branch was almost never entered (`child` was the host, whose `scrollWidth` rarely equalled the container's `offsetWidth`), so the non-truncated path cost two integer reads.

```ts
if (this.parent === this.child) return this.isHorizontalOverflown || this.isVerticalOverflown;
```

as an early return in `isOverflown` would fix this and the sub-pixel point above in one go.

lskramarov: **Two view queries for the same element.**

Angular allocates and refreshes a separate view-query slot for each `@ViewChild('kbqTitleContainer')`, on every change-detection pass of every rendered tree option — and leaves two independently assignable fields that can drift apart if either is ever re-pointed. A getter keeps the `KbqTitleTextRef` contract and makes "same element" self-evident.

```suggestion
    @ViewChild('kbqTitleContainer') parentTextElement: ElementRef;

    /**
     * Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself.
     * @docs-private */
    get textElement(): ElementRef {
        return this.parentTextElement;
    }

```

Verified locally — `prettier`, `eslint` and both new tests pass with this form. It also folds in the `@docs-private` point from the next comment. Note it changes the API-guard entry to `get textElement(): ElementRef;`, so `yarn approve-api` needs a re-run.

packages/components/tree/tree-option.component.ts
```diff
@@ -108,6 +108,8 @@ export class KbqTreeOption extends KbqTreeNode<KbqTreeOption> implements AfterCo
     preventBlur: boolean = false;
 
     @ViewChild('kbqTitleContainer') parentTextElement: ElementRef;
+    /** Same element as `parentTextElement` — `.kbq-option-text` clips the text, so it is measured against itself. */
```
lskramarov: **An internal measurement hook is published as *documented* API.**

`tools/public_api_guard/components/tree.api.md` now lists `textElement: ElementRef;` **without** the `// (undocumented)` marker — unlike `parentTextElement` one line above, and unlike every other implementation of this contract: `KbqOption` (`core/option/option.ts:169`), `KbqDropdownItem` (`:58`), `KbqListOption` (`:768`), `KbqTag` (`:307`).

That advertises an internal overflow-measurement detail as consumer-facing API and invites people to depend on it. `title.directive.ts` tags exactly this class of member `@docs-private` — the suggestion above already includes it.

packages/components/tree/tree-selection.component.spec.ts
```diff
@@ -1579,6 +1580,47 @@ describe('KbqTreeSelection', () => {
             });
         });
     });
+
+    describe('kbq-title', () => {
```
lskramarov, line 1584: **The stated risk of this PR is the one thing no test pins down.**

The description says content marking its own `#kbqTitleText` keeps priority — and `packages/components-dev/tree/template.html:96` plus every filter-bar tree-select pipe depend on exactly that. Both new tests use a plain-text option (`{{ node.name }}`), so nothing covers it.

If the fallback order in `KbqTitleDirective.childElements` (`title.directive.ts:200-206`) were ever reversed so `componentInstance?.textElement` won, every consumer that marks its own text span would silently start measuring `.kbq-option-text` against itself instead of the inner span — and this suite would stay green.

A third case with `<span #kbqTitleText>` in the projected content, asserting the child resolves to that span, closes it.

packages/components/tree/tree-selection.component.spec.ts
```diff
@@ -1579,6 +1580,47 @@ describe('KbqTreeSelection', () => {
...
+            fixture.detectChanges();
+
+            const optionDebugElement = fixture.debugElement.query(By.directive(KbqTitleDirective));
+
+            titleDirective = optionDebugElement.injector.get(KbqTitleDirective);
+            optionElement = optionDebugElement.nativeElement;
+            containerElement = optionElement.querySelector('.kbq-option-text')!;
+        });
+
+        it('should measure the clipping container, not the option host', () => {
```
lskramarov, line 1602: **No true-positive for vertical clipping.**

Both tests set `offsetHeight === scrollHeight` (20/20), so `isVerticalOverflown` is only ever exercised as a negative. It now compares `.kbq-option-text` against itself too, and two-line options are a supported shape — `tree/e2e.ts:365,371,386,394` render `.kbq-option-caption` inside the option, and `tree-option.scss:52` carries dedicated `:has(.kbq-option-caption)` layout.

The comment right below says the host is "always wider **and taller**" — the *taller* half has no positive assertion. A case with `offsetHeight: 20` / `scrollHeight: 40` asserting `isOverflown === true` would cover it.

packages/components/tree/tree-selection.component.spec.ts
```diff
@@ -1797,6 +1839,19 @@ abstract class TreeParams {
...
+        KbqTreeModule,
+        KbqTitleModule
+    ],
+    template: `
+        <kbq-tree-selection [dataSource]="dataSource" [treeControl]="treeControl">
+            <kbq-tree-option *kbqTreeNodeDef="let node" kbq-title kbqTreeNodePadding>{{ node.name }}</kbq-tree-option>
+        </kbq-tree-selection>
+    `
+})
+class KbqTreeAppWithTitle extends TreeParams {}
```
lskramarov: **Test host naming.**

`KbqTreeAppWithTitle` re-introduces the `Kbq` prefix that the newer hosts in this file have dropped — `TreeSelectionFocusStates` (L1867), `TreeWithActionButtonApp` (L1999). `TreeAppWithTitle` matches the direction the file is already moving in.

No suggestion block since the rename also touches L1585 and L1592.

lskramarov (commented):
Reviewed the fix. The approach is sound and I verified both new tests are genuine regression tests — they fail when `textElement` is removed and pass with it. `prettier`, `eslint` and the API-guard entry are clean.

Eight notes below: **7 inline**, plus one that lands in a file outside this diff and so has to live here.

---

### `packages/components/title/title.directive.ts:181-184` — the `hasEllipsis` rationale is now stale

The JSDoc justifies checking both elements like this:

> Both elements are checked because `text-overflow` is not inherited and consumers place it differently: `KbqOption`/`KbqDropdownItem` style the measured `child`, whereas `KbqTreeOption` styles the `parent` container that wraps the child.

`KbqTreeOption` was the *only* cited case for the `elementHasEllipsis(this.child) || elementHasEllipsis(this.parent)` disjunction — and after this PR its `parent` **is** its `child`, so nothing wraps anything. The sentence describes an arrangement that no longer exists, and the `||` is redundant for the one component it was written for.

`AGENTS.md` ("Comments"): *"Update or remove comments that become outdated because of your changes."*

packages/components/tree/tree-selection.component.spec.ts
KamilEmeleev: done

packages/components/tree/tree-selection.component.spec.ts
KamilEmeleev: done

packages/components/tree/tree-option.component.ts
KamilEmeleev: Confirmed, and accepted deliberately — the description now states it.

Scope of the loss: `offsetWidth`/`scrollWidth` are integers, so an overflow of one pixel or more is still caught by the fast path. Only a clip that rounds away — under one CSS pixel — goes unreported, and only for options that do not mark their own `#kbqTitleText`. Before this PR that same case reported truncation unconditionally, so it moves from *always shown* to *never shown*.

On the second way out — a real `#kbqTitleText` span inside `.kbq-option-text`: it restores the sub-pixel branch, but it costs the other half of your review. The child has to stay inline for its intrinsic width to be visible, and an inline element reports `scrollWidth === 0` (measured on the tree dev app: the projected spans read `inner_sw: 0`), so `isOverflown` would enter the rect branch on *every* hover instead of never — permanently, not just in the sub-pixel case. It also puts the wrapper in front of `kbq-normalize-text`'s `& > *` and makes `.kbq-option-text:empty` unreachable, and I cannot run the screenshot suite locally to check the fallout.

So the two suggested ways out are mutually exclusive on the performance axis. Trading a sub-pixel clip for a layout flush per hover did not look worth it — happy to revisit if you disagree.

KamilEmeleev: The observation is right, and it is a direct consequence of this PR: with `parent === child`, `offsetWidth === scrollWidth` holds for every option whose text fits, so the branch that used to be skipped is now entered on the common path.

The proposed one-liner does not work as written, though. `if (this.parent === this.child) return ...` fails the existing `should return true via getBoundingClientRect for text-only content (fraction-pixel case)` spec, and more importantly it disables the branch for every bare `kbq-title` host — the most common usage — where `parent === child` too, but the branch does real work there: it appends a temporary wrapper and measures the actual text width.

The condition needs `&& !this.hasOnlyText`. I have that version working locally (full component suite green, plus a spec asserting no `getBoundingClientRect` / `getComputedStyle` call on the fitting path), but it lives in `title.directive.ts`, which every tooltip consumer shares — breadcrumbs, modal, toast, filter-bar, username. Sending it as its own PR rather than widening this one.

KamilEmeleev: done

packages/components/tree/tree-option.component.ts
KamilEmeleev: done

packages/components/tree/tree-selection.component.spec.ts
KamilEmeleev: done

lskramarov (approved):

PR #1918, opened 2026-08-21
feat(components): use KbqScrollbar (#DS-3584,#DS-4543)

artembelik:


NikGurev (approved):

lskramarov (approved):

packages/components/scrollbar/scrollbar.ts
```diff
@@ -230,51 +245,69 @@ export type KbqScrollbarScrollToElementOptions = {
...
+        alias: 'kbqScrollbarHideDelay',
+        transform: numberAttribute
+    });
 
     // Reference to the dynamically created track, used to update its mode and destroy it when custom
     // scrollbars are disabled.
     private trackRef: ComponentRef<KbqScrollbarTrack> | null = null;
 
+    // Fires each `flashScrollIndicators()` call; the track reveals itself on it just like on a scroll event.
+    private readonly flashSubject = new Subject<void>();
```
NikGurev, line 290: @rmnturov возможно, нужно будет доработать спеку

Здесь поведение, схожее с IOS - скролл мигает в начале и потом пропадает в режиме появления по ховеру

packages/components/popover/popover.component.html
```diff
@@ -32,7 +32,8 @@
             <div
                 #popoverContent="kbqOverflowShadowContainer"
                 kbqOverflowShadowContainer
-                class="kbq-popover__content kbq-scrollbar"
+                kbqScrollbarViewport
```
lskramarov, line 35: **1. A popover with `closeOnScroll` now closes when its own content is scrolled**

`kbqScrollbarViewport` brings `CdkScrollable`, so `.kbq-popover__content` registers with the global `ScrollDispatcher`. `KbqPopoverTrigger.closingActions()` merges that stream **unfiltered**:

```ts
// popover.component.ts:553
return merge(...this.closingActionsForClick(), this.closeOnScroll ? this.scrollDispatcher.scrolled() : NEVER);
```

**Repro:** `docs-examples/components/popover/popover-scroll/popover-scroll-example.ts:46` — `[closeOnScroll]="true"` with content taller than the popover. Wheel inside `.kbq-popover__content` → the panel's own `CdkScrollable` broadcasts → `subscribeOnClosingActions` calls `hide()`. The popover closes the instant you try to scroll it. Before this PR the content was not a registered scrollable, so inner scrolling was inert.

`KbqAutoHideScrollStrategy` already guards exactly this (`core/overlay/auto-hide-scroll-strategy.ts:92-95`): `!overlayRef.overlayElement.contains(scrollable.getElementRef().nativeElement)`. `closingActions()` needs the same filter.

---
_Generated with Claude Code_

packages/components/dropdown/dropdown.component.ts
```diff
@@ -481,6 +492,20 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
...
     }
 
+    /**
+     * Closes the dropdown on a panel click. The panel is the custom scrollbar's viewport, so its
+     * track and thumb are real DOM children whose clicks bubble here — interacting with the scrollbar
+     * must not close the dropdown, unlike a click on an item.
+     * @docs-private
+     */
+    protected onPanelClick(event: MouseEvent): void {
+        if ((event.target as HTMLElement).closest('.kbq-scrollbar-track')) {
```
lskramarov, line 502: **2. Dragging the scrollbar thumb closes the dropdown**

The guard reads `event.target`, but `click` is dispatched on the **nearest common inclusive ancestor** of the mousedown and mouseup targets — not on the mousedown target.

**Repro:** open a dropdown whose panel overflows → press the mouse on the vertical thumb → drag down while drifting a few px sideways → release over a `kbq-dropdown-item`. mousedown target = thumb, mouseup target = item, so the browser fires `click` on `.kbq-dropdown__panel`; `closest('.kbq-scrollbar-track')` returns `null`; `close()` runs. The dropdown closes mid-drag. `(mousedown)="$event.preventDefault()"` on the bar does not suppress `click`.

The new e2e only covers a stationary `.click()` on the thumb, never a drag.

Suggestion: key the decision off the **mousedown** origin, or suppress the close while a thumb drag is active. Worth noting this line is also the only cross-package reference to the scrollbar's internal `.kbq-scrollbar-track` class in the whole library — `(click)="$event.stopPropagation()"` on the track's own bar would fix it once for every consumer.

---
_Generated with Claude Code_

packages/components/modal/modal.scss
```diff
@@ -127,10 +127,6 @@
         var(--kbq-modal-size-content-padding-bottom) var(--kbq-modal-size-content-padding-horizontal);
 
     overflow-wrap: break-word;
```
lskramarov, line 129: **3. Deleting `.kbq-modal-body > * { position: relative; }` reverts the DS-2701 fix**

The rule removed right below this line was added by `0f964fd24` — *"fix(modal): fixed scroll if childs contains radio or toggle (#DS-2701)"*. Nothing in this PR re-establishes it, and no unit or e2e test covers DS-2701, so the revert is silent.

It made every direct child of the scrolling body a containing block, so `position: absolute` descendants of radio/toggle could not escape the scrollport. Neither `.kbq-radio-button` nor `.kbq-radio-label` carries `position` (`radio.scss:19,23`), and the radio's hidden input is `<input class="kbq-radio-input cdk-visually-hidden">` (`position:absolute; left:0; margin:-1px`), so without the rule its containing block escapes all the way to `.kbq-modal-content` (`modal.scss:75`).

The regression would now be **more** visible than before: a spurious overflow makes the new custom track render a bar, where previously it only nudged a native one.

I assume the motive was the specificity tie — `.kbq-modal-body > *` is (0,1,0), the same as the track's `:host { position: sticky }` compiled to `[_nghost-…]`, so injection order decides the winner. If so, `> *:not(kbq-scrollbar-track)` keeps both the fix and the sticky track.

This is the one finding I could not reproduce end-to-end — the DS-2701 repro would settle whether it still matters.

---
_Generated with Claude Code_

packages/components/autocomplete/autocomplete.html
```diff
@@ -1,6 +1,6 @@
 <ng-template>
     <div class="kbq-autocomplete-panel" [id]="id" [class]="classList" [style.max-width.px]="panelMaxWidth()">
-        <div #panel role="listbox" class="kbq-autocomplete-panel__content kbq-scrollbar">
+        <div #panel kbqScrollbarViewport role="listbox" class="kbq-autocomplete-panel__content">
```
lskramarov, line 3: **4. Injecting the scrollbar track into `role="listbox"` breaks AXE**

`kbqScrollbarViewport` was added to the same element that carries `role="listbox"`, and `KbqScrollbarViewport.createTrack()` inserts the track as that element's **first child**:

```html
<div role="listbox">
  <kbq-scrollbar-track>
    <div class="kbq-scrollbar-track__bar…">
      <div role="scrollbar" aria-controls aria-valuemin aria-valuemax></div>
```

`listbox` permits only `option`/`group` among its owned children. axe-core's `aria-required-children` recurses through the role-less `kbq-scrollbar-track`/`__bar` wrappers and reports `role="scrollbar"` as *"Element has children which are not allowed"* — for any autocomplete panel whose list overflows.

AGENTS.md: *"It MUST pass all AXE checks."* The previous `class="… kbq-scrollbar"` was pure CSS and injected no DOM, so the listbox owned exactly its options.

Nothing catches this — only `accordion`, `app-switcher` and `button-toggle` run `jest-axe`. Select/tree-select/timezone escape it because their `role` is not on `#optionsContainer`; autocomplete is the only one that co-locates the two. Moving `role="listbox"` to an inner wrapper would resolve it.

---
_Generated with Claude Code_

packages/components/select/select.html
```diff
@@ -119,8 +119,9 @@
 
         <div
             #optionsContainer
+            kbqScrollbarViewport
```
lskramarov, line 122: **5. Scrolling a panel now re-runs `updatePosition()` on every scroll event, unthrottled**

Panels that gained `kbqScrollbarViewport` register on the global `ScrollDispatcher`, and CDK's `RepositionScrollStrategy` — the default for select, dropdown and autocomplete — subscribes with **no containment filter and no throttle**:

```js
// overlay-module.mjs:268-270
const throttle = this._config ? this._config.scrollThrottle : 0;
this._scrollSubscription = this._scrollDispatcher.scrolled(throttle).subscribe(() => {
    this._overlayRef.updatePosition();
```

Both `kbqSelectScrollStrategyProviderFactory` (`core/select/constants.ts:31`) and `KBQ_DROPDOWN_SCROLL_STRATEGY_FACTORY` (`dropdown-trigger.directive.ts:63`) return `reposition()` with no `scrollThrottle`, i.e. `scrolled(0)`.

So wheel-scrolling an open select's option list fires a full `FlexibleConnectedPositionStrategy.apply()` — `getBoundingClientRect` on origin, overlay and viewport plus style writes — on *every* scroll event, for this overlay **and every other open reposition-strategy overlay in the app**. Before this PR the panels were not registered scrollables.

Same exposure at `dropdown.html:3`, `autocomplete.html:3`, `popover.component.html:35`, `modal.component.html:98`, `tree-select.html:119`, `timezone-select.component.html:74`.

---
_Generated with Claude Code_

packages/components/tabs/tab-body.html
```diff
@@ -1,7 +1,8 @@
 <div
     #content
     cdkScrollable
-    class="kbq-tab-body__content kbq-scrollbar kbq-hide-nested-popup"
+    kbqNativeScrollbar
```
lskramarov, line 4: **6. `kbqNativeScrollbar` drops descendant scrollbar theming that `.kbq-scrollbar` provided**

The class this replaces styled the host **and every descendant** — `_kbq-scrollbar()` emits both selectors (`core/styles/theming/_scrollbar-theme.scss:36-37`):

```scss
&::-webkit-scrollbar,
::-webkit-scrollbar { … }   // → .kbq-scrollbar ::-webkit-scrollbar
```

`KbqNativeScrollbar` defaults `descendants` to `false` and, on Firefox, actively resets them (`native-scrollbar.scss:6-8`):

```scss
:where(.kbq-native-scrollbar:not(.kbq-native-scrollbar_descendants) *) { scrollbar-color: auto; }
```

None of the new call sites pass `kbqNativeScrollbarDescendants`, so consumer-authored scrollers projected into a tab body, into `.kbq-code-block__main` (`code-block.html:100`), or into `<kbq-list kbqNativeScrollbar>` (`multiple-file-upload.component.html:26`) silently revert to the browser default.

First-party nested scrollers survive (their own `.kbq-native-scrollbar` rule outranks the zero-specificity `:where`), so this only bites projected content — which is exactly what a tab body holds.

---
_Generated with Claude Code_

packages/components/scrollbar/scrollbar.ts
```diff
@@ -459,11 +522,11 @@ class KbqScrollbarThumb {
     private getScrolled({ clientY, clientX }: MouseEvent, offsetY: number, offsetX: number): ScrollPosition {
         const { offsetHeight, offsetWidth } = this.nativeElement;
         const { top, left, right, width, height } = this.nativeElement.parentElement!.getBoundingClientRect();
-        const rtl = this.nativeElement.matches('[dir="rtl"] :scope');
+        const rtl = this.directionality.value === 'rtl';
```
lskramarov, line 525: **7. RTL detection narrowed from a live DOM check to a DI value that is usually `'ltr'`**

This was `this.nativeElement.matches('[dir="rtl"] :scope')` — a live query re-run on every `getScrolled()` call that matched *any* `dir="rtl"` ancestor. CDK's root `Directionality` reads `document.body.dir || document.documentElement.dir` **once in its constructor** and never updates; only the `Dir` directive (`[dir]`, from `BidiModule`, which a consumer must explicitly import) supplies a live per-subtree value.

**Two regressions:**
1. `<div dir="rtl"><kbq-scrollbar>…</kbq-scrollbar></div>` in a component that does not import `Dir` — CSS lays the bar out RTL, but `directionality.value === 'ltr'`, so `getScrolled` uses `inline = left` / `multiplier = 1`. Dragging or click-jumping the horizontal thumb scrolls to the mirrored offset.
2. An app that flips locale with `document.documentElement.setAttribute('dir','rtl')` after bootstrap never updates.

This repo's own fixtures show the gap: `select/e2e.ts:382,401` writes `<div dir="rtl">` **and** has to hand-provide `{ provide: Directionality, useValue: { value: 'rtl', … } }`, because the bare attribute does not feed DI.

The spec was changed to add `Dir` to `imports` so the new mechanism resolves — which masks the narrowing rather than covering it. `nativeElement.closest('[dir="rtl"]')` keeps the old semantics *and* works in jsdom, which was the original motivation for the change.

---
_Generated with Claude Code_

packages/components/scrollbar/scrollbar.ts
```diff
@@ -581,50 +639,125 @@ class KbqScrollbarThumb {
...
+    readonly hideDelay = input.required<number>();
+
+    constructor() {
+        // Reapply the geometry whenever the viewport's box changes. `SharedResizeObserver` fires only on an
+        // actual size/padding change (a padding change shifts the content box too), not every frame, so
+        // `getComputedStyle` runs only when something changed — and its `shareReplay` delivers the current
+        // size on subscribe, applying the initial geometry right away. Writes styles directly, no change
+        // detection. In SSR (no `ResizeObserver`) the stream simply never emits.
+        this.resizeObserver
+            .observe(this.viewportElement)
```
lskramarov, line 696: **8. Track geometry has no CSS fallback — nothing renders without `ResizeObserver`**

Every geometry declaration was deleted from `scrollbar-track.scss` (`inset-block-start`, `inset-inline-start`, `min/max-inline-size: calc(100% - 1px)`, `margin-inline-end: calc(-100% + 1px)`) along with the `[style.block-size.px]` host binding. All of it now comes only from `applyGeometry()`, driven only by this subscription.

`SingleBoxSharedResizeObserver` guards `typeof ResizeObserver !== 'undefined'` and otherwise returns a stream that never emits. So `applyGeometry` never runs: `:host` keeps `block-size: auto` → 0 (all children are `position: absolute`), `.kbq-scrollbar-track__bar_vertical { inset-block: 0 }` resolves against a 0px box, and **no scrollbar paints at all** — while `visibility()` (still rAF-polled) says the bars should be shown.

The same holds for the frames between track insertion and the browser's first RO callback, on every panel open.

The constructor comment is also inaccurate on one point:

> its `shareReplay` delivers the current size on subscribe, applying the initial geometry right away

`shareReplay({bufferSize: 1, refCount: true})` has an **empty buffer for the first subscriber** of an element — which is the normal case here, one track per viewport. The first value comes from the browser's initial RO callback, asynchronously.

The previous implementation degraded to "correct within 300 ms" everywhere, because the poll had no platform dependency.

---
_Generated with Claude Code_

packages/components/modal/modal.directive.ts
```diff
@@ -51,16 +53,21 @@ export class KbqModalTitle {
 @Directive({
     selector: `[kbq-modal-body], kbq-modal-body, [kbqModalBody]`,
     host: {
-        class: 'kbq-modal-body kbq-scrollbar'
+        class: 'kbq-modal-body'
     },
-    hostDirectives: [KbqOverflowShadowContainer]
+    hostDirectives: [KbqOverflowShadowContainer, KbqScrollbarViewport]
```
lskramarov, line 58: **9. The viewport's `[attr.id]` host binding overwrites a consumer's bound `[id]`**

`KbqScrollbarViewport` carries `'[attr.id]': 'id'` where

```ts
// scrollbar.ts:271
protected readonly id = this.getNativeElement().id || this.idGenerator.getId('kbq-scrollbar-viewport-');
```

That is a **field initializer** — it reads the DOM id during directive construction, when only *static* attributes exist. Adding the viewport as a host directive pushes this onto every public `<kbq-modal-body>` and `<kbq-sidepanel-body>` in every consumer app.

**Repro:** `<kbq-modal-body [id]="describedById">` used as an `aria-describedby` target. At construction the DOM `id` is `''`, so `id` resolves to `kbq-scrollbar-viewport-N`. Angular runs a view's host bindings *after* that view's template property bindings in the same `refreshView`, so `[attr.id]` stamps the generated id over the bound one on first CD and it sticks until `describedById` itself changes — breaking the ARIA reference with no error. Static `id="foo"` is preserved; a bound one is not.

Same at `sidepanel-directives.ts:124`.

Related gap: neither `hostDirectives` entry lists `inputs`, so `<kbq-modal-body kbqScrollbarMode="native">` is an inert DOM attribute.

---
_Generated with Claude Code_

packages/components/select/select.component.ts
```diff
@@ -382,6 +385,17 @@ export class KbqSelect
...
+    /** The options container's custom scrollbar viewport, flashed when the panel opens. */
+    private readonly scrollbarViewport = viewChild(KbqScrollbarViewport);
+
+    /**
+     * A custom scrollbar a consumer projects onto the panel's own scroller — under virtual scroll the
+     * options container is `native` (never overflows), so the real scroller is the projected
+     * `cdk-virtual-scroll-viewport` carrying `kbqScrollbarViewport`. Flashed on open alongside the
+     * container's; whichever is inert (`native`) no-ops.
+     */
+    private readonly projectedScrollbarViewport = contentChild(KbqScrollbarViewport, { descendants: true });
```
lskramarov, line 397: **10. The projected-viewport query is unfiltered and positional**

The JSDoc's *"whichever is inert (`native`) no-ops"* holds only **under virtual scroll**. I verified the no-op claim itself is correct — `mode: 'native'` makes the viewport's effect call `destroyTrack()`, so no track exists and `flashSubject.next()` has no subscribers.

But `scrollbarMode` (line 919) returns `'native'` only when `withVirtualScroll` is true, while this query matches the first projected `KbqScrollbarViewport` in document order with no filter.

So a consumer who projects any `kbqScrollbarViewport`/`<kbq-scrollbar>` into a **non-virtual** `<kbq-select>` — a scrollable custom option group, a nested scroller in a projected footer or search template — gets both viewports live in `hover` mode, and two tracks flash on open. Being positional, it can also pick a viewport that is not the panel's real scroller.

The select already holds `virtualScrollViewport = contentChild(CdkVirtualScrollViewport)` (line 412); filtering to viewports hosted on *that* would express the actual intent.

---
_Generated with Claude Code_

packages/components/scrollbar/scrollbar.ts
```diff
@@ -230,51 +245,69 @@ export type KbqScrollbarScrollToElementOptions = {
...
-        // viewport itself whenever ANY descendant is keyboard-focused — projected content doesn't have
-        // to opt in individually with its own `cdkMonitorElementFocus`. Drives the hover-mode track's
-        // visibility in scrollbar-viewport.scss.
-        '[attr.cdkMonitorSubtreeFocus]': 'true',
         // A stable id for the thumb's `aria-controls` to point at, preserving one a consumer already
         // set rather than clobbering it.
         '[attr.id]': 'id'
     },
-    hostDirectives: [CdkScrollable, CdkMonitorFocus]
+    hostDirectives: [CdkScrollable]
```
lskramarov, line 258: **11. A `cdk-virtual-scroll-viewport` ends up with two `CdkScrollable`s registered**

`CdkVirtualScrollViewport` declares

```js
// scrolling.mjs:1057
providers: [{ provide: CdkScrollable, useFactory: (virtualScrollable, viewport) => virtualScrollable || viewport, … }]
```

Node **providers are scanned before node directives**, so `inject(CdkScrollable)` at line 262 resolves to the `CdkVirtualScrollViewport` — while this host-directive instance is still constructed and still registers itself (`CdkScrollable.ngOnInit` → `scrollDispatcher.register(this)`, `scrolling.mjs:367`).

**Result:** `<cdk-virtual-scroll-viewport kbqScrollbarViewport>` puts two entries in `ScrollDispatcher.scrollContainers` for the same element, so every scroll emits **twice** on the global `scrolled()` bus and `getAncestorScrollContainers()` reports it twice — doubling the per-frame work of every reposition/close-on-scroll subscriber, compounding the unthrottled `updatePosition()` noted on `select.html`.

This is the pattern the PR actively promotes: `docs-examples/.../select-virtual-scroll`, `docs-examples/.../list-virtual-scroll`, the `E2eVirtualScrollSelectScrollbar` fixture, and the `scrollbar.en.md` guidance.

---
_Generated with Claude Code_

packages/components/scrollbar/scrollbar-viewport.scss
```diff
@@ -13,13 +13,14 @@
...
-// scoped to keyboard-originated focus (`KbqScrollbarViewport` monitors its subtree via
-// `cdkMonitorSubtreeFocus`), so the track stays hidden for mouse focus but visible while, e.g., arrow
-// keys scroll the content.
+// The hover track is revealed while:
+// - the pointer is over the viewport (`:hover`);
+// - the track is transiently revealed (`kbq-scrollbar-track_revealed`, toggled by KbqScrollbarTrack on a
+//   scroll or `flashScrollIndicators()` and auto-cleared shortly after) — matches native/overlayscrollbars,
+//   and covers reveal on wheel/keyboard scrolling, the programmatic scroll-into-view when a dropdown opens
+//   by mouse, and an explicit flash.
 .kbq-scrollbar-viewport:hover > .kbq-scrollbar-track_hover,
```
lskramarov, line 22: **12. Keyboard focus no longer reveals the track, with nothing put in its place**

`CdkMonitorFocus` + `'[attr.cdkMonitorSubtreeFocus]': 'true'` and the `.kbq-scrollbar-viewport.cdk-keyboard-focused > .kbq-scrollbar-track_hover` rule are gone, leaving `:hover` and the transient `_revealed` (scroll/flash) as the only reveal paths. The e2e that guarded the old behaviour (`hover mode: keyboard-focusing a plain descendant keeps the track visible`) was **deleted**, not replaced.

A keyboard-only user who tabs into an overflowing `hover`-mode region whose focused control is already in view gets no track at all: nothing scrolls, nothing hovers, so there is no indication the region is scrollable. The compensating `flashScrollIndicators()` calls only went to modal/sidepanel/popover/dropdown/select/tree-select/autocomplete/timezone — a plain `<kbq-scrollbar>` or bare `kbqScrollbarViewport` consumer (content-panel, app-switcher popup, any consumer scroller) gets nothing.

Two further consequences worth an explicit decision:
- `revealed` self-clears after `hideDelay`, so even during Tab navigation the track can no longer *stay* visible.
- `.cdk-focused`/`.cdk-keyboard-focused` silently stop appearing on every `.kbq-scrollbar-viewport` host — a DOM contract consumers could have styled against.

Also: the surviving test at `e2e.playwright-spec.ts:251` (`hover mode: focusing a descendant does not reveal the track`) can no longer fail — there is no focus-driven reveal left to regress.

---
_Generated with Claude Code_

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -481,13 +481,19 @@ describe('KbqDropdown', () => {
         dispatchMouseEvent(triggerEl, 'touchstart');
         triggerEl.click();
         fixture.detectChanges();
-        flush();
+        // Not `flush()`: the panel now hosts `kbqScrollbarViewport`, whose track drives a self-rescheduling
+        // `requestAnimationFrame` loop that `flush()` can never drain (it hits the 20-task limit). `tick`
+        // advances a fixed span instead — enough for the dropdown's own open timers — and
+        // `discardPeriodicTasks()` clears the still-pending scrollbar tasks so the test ends cleanly.
+        tick(500);
```
lskramarov, line 488: **13. The perpetual rAF loop is worked around in the tests instead of being fixed**

This workaround (and its siblings in `modal.spec.ts:419`/`:549` and `timezone-select.component.spec.ts:933`/`:950`) treats the symptom. The cause is `animationFrame()` in `scrollbar.ts:54` — a self-rescheduling `requestAnimationFrame` loop with **no stop condition**, subscribed twice per scrollbar (`KbqScrollbarTrack.visibility` throttled to 300 ms, `KbqScrollbarThumb` to 100 ms). This PR multiplies live instances roughly tenfold: every modal body, sidepanel body, select/tree-select/timezone/dropdown/autocomplete panel and popover now runs two permanent loops doing throttled forced-layout reads whether or not anything scrolled.

The same root cause already leaked into e2e — `sidepanel/e2e.playwright-spec.ts` needs `hover({ force: true })` because *"the track's own scroll-position `requestAnimationFrame` loop repaints every frame, which webkit intermittently reports as the body never settling."*

Only three call sites were converted, while **~600 `flush()` calls remain** in the same suites (select 339, tree-select 155, timezone 44, autocomplete 25, dropdown 19, sidepanel 13, modal 9). So the boundary between "works" and `flush failed after reaching the limit of 20 tasks` is undocumented and load-dependent — and downstream consumers hit it with no migration note, since `tick(N)` is a hand-tuned magic number where `flush()` was not.

Both inputs the loops poll are already available event-driven: `viewport.scrollChanges`, and the `SharedResizeObserver` subscription this PR just added to the same class. Fixing the loop removes all three test workarounds, the e2e `force: true`, and the idle cost.

---
_Generated with Claude Code_

packages/components/select/select.component.ts
```diff
@@ -1113,6 +1138,11 @@ export class KbqSelect
                         search.focus();
                     }
 
+                    // Briefly reveal the scrollbar to hint that the list is scrollable — the panel may open
+                    // on an already-visible option with no scroll, so the hover track would otherwise stay hidden.
+                    this.scrollbarViewport()?.flashScrollIndicators();
```
lskramarov, line 1143: **14. "Flash on open" is hand-wired at nine sites through seven different hooks**

One rule — *flash when the viewport first becomes visible* — is implemented nine times, each with its own `viewChild`/`inject` field and a near-verbatim two-line comment:

| site | hook |
|---|---|
| `select.component.ts:1143`, `:1144` | `panelDoneAnimatingStream` |
| `tree-select.component.ts:844` | `panelDoneAnimatingStream` |
| `autocomplete.component.ts:231` | `outputToObservable(opened)` |
| `dropdown.component.ts:468` | `onAnimationDone`, `toState === 'enter'` |
| `modal.component.ts:626` | a visibility promise `.then()` |
| `modal.directive.ts:70` | `afterOpen` |
| `popover.component.ts:145` | an `animationDone` override |
| `sidepanel-directives.ts:139` | `afterOpened()` |

`flashes` is a plain `Subject` (no replay) and `KbqScrollbarTrack` — its only subscriber — is created lazily in an `effect()` + `afterNextRender`. Every site is therefore silently timing-coupled to track creation: a hook that fires one tick early drops the flash with **no error and no failing unit test** (the specs assert host classes, not the flash).

Two of the shapes subscribe to **one-shot** sources: `afterOpened$` is `complete()`d immediately after emitting (`sidepanel-ref.ts:73-74`) and `afterOpen` is a non-replaying `EventEmitter`. A body behind `@if`/`@defer` that resolves after the open animation never flashes.

The tests carry the same duplication: nine copy-pasted `describe` blocks (~330 lines) across eight spec files, eight near-identical `E2e*Scrollbar` hosts, five copies of the `{scrolled, register, deregister}` `ScrollDispatcher` stub, and a fourth hand-rolled `SharedResizeObserver` double — three already exist in `code-block.spec.ts`, `overflow-shadow.spec.ts`, `clamped-text.spec.ts`, and `packages/components/core/testing/` is the established home for shared doubles.

Detecting "my viewport just became visible" inside `KbqScrollbarViewport` would cover all nine, plus every future panel, once.

---
_Generated with Claude Code_

lskramarov (commented):
## Code review — 15 findings

Reviewed `main...feat/DS-3584` at `aeaf1ba` (71 files, screenshots excluded). 14 findings are inline; the 15th is below, because its file is not part of this diff.

Ordered by severity. Every finding was verified against the source — mechanisms are quoted, not inferred.

**The two I would not merge without:** the `closeOnScroll` popover closing on its own scroll, and the scrollbar-thumb drag closing the dropdown. **The highest-leverage change** is the perpetual `requestAnimationFrame` loop — it is the single cause of the three `flush()` workarounds, the e2e `hover({ force: true })`, and the per-viewport idle cost, and both signals it polls are already wired up in the same class.

---

### 15. `scrollbar-deprecated-path` README states a now-wrong `KbqScrollbarOptions` shape

`packages/schematics/src/migrations/scrollbar-deprecated-path/README.md:33`

> `KbqScrollbarOptions` is now `{ mode: KbqScrollbarMode }` instead of the `overlayscrollbars` `PartialOptions`.

This PR adds a required `hideDelay: number` to that exported type. A consumer running the repo's own migration follows the README, keeps `const opts: KbqScrollbarOptions = { mode: 'always' }`, and gets `TS2741: Property 'hideDelay' is missing` — from the document that explicitly promised the new shape.

Compounding it, `grep -c "hideDelay\|HideDelay"` returns **0** for both `scrollbar.en.md` and `scrollbar.ru.md`, even though this PR substantially rewrote both (mode list → table, methods list → table, two brand-new sections). The new public `kbqScrollbarHideDelay` input, its millisecond unit, its `1000` default, and `kbqScrollbarOptionsProvider({ hideDelay })` are undiscoverable: the provider is still described as changing "the default **mode**" only, and the new "Scroll indicators" section documents `flashScrollIndicators()` without saying how long the reveal lasts.

---

### Checked and refuted

Recording these so nobody re-spends time on them:

- **Removing `KBQ_SCROLLBAR_VIEWPORT` is not a breaking change** — its introducing commit `3e49f593b` is in no release tag (latest is `20.2.0`), so no `ng update` migration is owed. Same for `KbqScrollbarOptions` gaining a required field.
- **`ComponentRef.destroy()` does not strand the track in the DOM** — `appRef.detachView` → `detachFromAppRef` → `detachViewFromDOM` removes it, which is why `native mode: no custom track is rendered` passes.
- **The `flush()` → `tick(500)` swaps do not make assertions vacuous** — the tooltip `enterDelay` is 400 ms.
- **The RTL spec is not vacuous** — flipping `dir` to `ltr` makes it fail with `Expected -100, Received 100`.
- Screenshot filenames do not collide; the `prefers-reduced-motion` `:where()` override wins on source order; the injected track adds no flex `gap` (the affected containers are `display: block`); no `:first-child`/`nth-child` logic keys off the containers that gained the track.

---
_Generated with Claude Code_

rmnturov:
В дропдауне дергается скролл. Мак, хром

https://github.com/user-attachments/assets/ede58eca-8625-442f-a904-14fb6d8df3bd

rmnturov:
В аг-гриде вижу disabled-scrollbar в левой стики-колонке
image
https://koobiq-next--prs-1918-23wtc4bj.web.app/ru/components/ag-grid/overview

rmnturov:
В блоке кода в доках всегда виден скроллбар, а я так понял нужно по ховеру
image

artembelik:
> В аг-гриде вижу disabled-scrollbar в левой стики-колонке

ag-grid не трогали в данной задаче, баг на это есть DS-5190

artembelik:
> В блоке кода в доках всегда виден скроллбар, а я так понял нужно по ховеру

скрол в код-блоке, файлпикере, текстарее и табах будет исправлен отдельно в рамках DS-5453

rmnturov:
Натыкался на флеш скроллбара в выпадашках, где в итоге не было прокрутки. Надеюсь, получится исправить.

rmnturov:
Сам флеш мне нра, как идеалогически, так и его анимация, тайминг

packages/components/scrollbar/scrollbar.ts
rmnturov: > @rmnturov возможно, нужно будет доработать спеку
> 
> Здесь поведение, схожее с IOS - скролл мигает в начале и потом пропадает в режиме появления по ховеру

Я доработал в Фигме

packages/components/popover/popover.component.html
artembelik: Поправил

packages/components/dropdown/dropdown.component.ts
artembelik: поправил

packages/components/autocomplete/autocomplete.html
artembelik: поправил

packages/components/select/select.html
artembelik: добавил `KbqRepositionScrollStrategy` который умеет `ignoreInnerScroll`

packages/components/tabs/tab-body.html
artembelik: поправил, задачу на миграцию завел

packages/components/scrollbar/scrollbar.ts
artembelik: оставил оба варианта

packages/components/scrollbar/scrollbar.ts
artembelik: проверил, это не актуально, ResizeObserver отрабатывает при инициализации в браузере

packages/components/modal/modal.directive.ts
artembelik: поправил

packages/components/scrollbar/scrollbar-viewport.scss
artembelik: больше не актуально, сошлись на том, что скрол отображается по hover и при скроле
если нужно отображать его всегда, есть always mode

packages/components/dropdown/dropdown.spec.ts
artembelik: поправил

packages/components/scrollbar/scrollbar.ts
artembelik: исправил

artembelik:
> В дропдауне дергается скролл. Мак, хром

> Натыкался на флеш скроллбара в выпадашках, где в итоге не было прокрутки. Надеюсь, получится исправить.

поправил

rmnturov (approved):

NikGurev:
В поповере скролбар отображается за крестиком. По идее должен появляться над всем контентом

image

NikGurev:
Контент поповера в [Примерах использования](https://koobiq-next--prs-1918-2k7fsry6.web.app/ru/components/popover/overview#%D0%BF%D1%80%D0%B8%D0%BC%D0%B5%D1%80%D1%8B-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D1%8F) прыгает. Ожидал, что как и в других компонентах контент будет выводиться "под скролбаром"

https://github.com/user-attachments/assets/0ee0c06d-74cd-452b-8ce4-a9462ca09974

artembelik:
> Контент поповера в [Примерах использования](https://koobiq-next--prs-1918-2k7fsry6.web.app/ru/components/popover/overview#%D0%BF%D1%80%D0%B8%D0%BC%D0%B5%D1%80%D1%8B-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D1%8F) прыгает. Ожидал, что как и в других компонентах контент будет выводиться "под скролбаром"
> 
>  Screen.Recording.2026-08-31.at.10.40.04.mov

это никак не связано со скролбаром и текущей реализацией, контент и до этого "прыгал", если обратишь внимание он прыгает и без скролбара (даже в старой реализации) - видимо особенность реализации

artembelik:
> В поповере скролбар отображается за крестиком. По идее должен появляться над всем контентом

посмотрю, что можно сделать

PR #1920, opened 2026-08-21
feat: update icons migration schematics (#DS-4328)

NikGurev:
## Summary

`deprecated-icons` and `new-icons-pack` are the two schematics that rename old icon usages (`pt-icons-*` / `mc-*`) to the current `kbq-*` names. They used to work by find-and-replace across the whole file, so they sometimes renamed things that weren't icons at all — a component name, an unrelated CSS class, random matching text. This PR rewrites both to actually read and understand the code first (parsing the HTML/TypeScript instead of scanning raw text), so they only touch real icon usages.

## What changed for users

- Only genuine icon code gets renamed — a component like `<mc-button>`, an unrelated class such as `.mc-panel-header`, or an unrelated string is left alone.
- Covers more places than before: `class` attributes, `[class.icon-name]` bindings, plain and bound icon attributes (with quotes preserved), bare string literals in `.ts` files, a component's inline `host` class and `styles` array, and icons inside `{count, plural, ...}` blocks.
- A value the migration can't safely resolve on its own (a dynamic expression, string concatenation, or `{{ }}` interpolation) is reported as a warning instead of being silently rewritten into broken code.
- `fix: false` (the default) never changes any file — it only reports what it found.
- An icon that was already partly migrated (e.g. already has the `kbq-` prefix but not the current name) still gets corrected, instead of being treated as already done.
- `updatePrefix` (`new-icons-pack`) now only controls the extra prefix rename in stylesheets — icon *name* renames always apply, matching how it worked before this schematic pass.
- Fixed a packaging bug where the schematics would crash on first use after being published.

## Known limitation

Stylesheets and Markdown files don't have a real parser behind them, so matching there is a careful regex instead — this can occasionally still match a comment that spells out a class selector. Documented in both READMEs.

## Test plan

- [x] `yarn run unit:schematics` — both migrations' specs pass, including new coverage for TS/Markdown paths, host/styles, ICU expansions, and dynamic-value warnings


packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+        forEachStringLiteral(sourceFile, (node) => {
+            const start = node.getStart(sourceFile) + 1;
+            const end = node.getEnd() - 1;
+            const isInsideTemplate = templateRanges.some((range) => start >= range.start && end <= range.end);
+
+            if (isInsideTemplate) {
+                return;
+            }
+
+            const resolved = resolveBareStringLiteral(node.text, tokenMap, data);
```
lskramarov: **[blocking] String-literal edits use raw source offsets but content from `node.text`, so escape sequences are destroyed and the file stops compiling.**

`node.text` is the *decoded* literal value, but the edit on line 385 is spliced over the *raw* span `[getStart()+1, getEnd()-1)`.

Verified end-to-end (`new-icons-pack`, `fix: true`):

```ts
// in
export const a = 'mc-add-to-list_16 don\'t';
export const b = 'mc-add-to-list_16 a\tb';
export const c = 'mc-add-to-list_16 C:\\dir';
export const d = "mc-add-to-list_16 say \"hi\"";

// out
export const a = 'kbq-file-plus-o_16 don't';      // syntax error
export const b = 'kbq-file-plus-o_16 a b';        // tab escape silently lost
export const c = 'kbq-file-plus-o_16 C:\dir';     // backslash escape destroyed, runtime value changed
export const d = "kbq-file-plus-o_16 say "hi"";   // syntax error
```

Re-parsing the output throws `Unexpected identifier 't'`. The migration writes a file the project can no longer build. Same in `deprecated-icons`.

The fix needs either a raw-text-preserving rewrite (operate on `content.slice(start, end)` rather than `node.text`) or re-escaping the result before splicing.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+                            }`
+                        );
+                        continue;
+                    }
+
+                    const cleanValue = getSimpleAttributeValue(attr.value).replace(/"/g, '');
+                    const resolved = resolveValueToken(cleanValue);
+
+                    if (resolved) {
+                        edits.push({
```
lskramarov, line 320: **[blocking] A quoted string literal in a bound icon attribute loses its quotes and becomes an identifier expression.**

`attr.valueSpan` covers the value *including* the inner quotes, but `cleanValue` strips them for the lookup and the unquoted `resolved` is spliced over the whole span.

Verified end-to-end:

```html
<!-- in -->
<i [kbq-icon]="'mc-add-to-list_16'"></i>
<i [kbq-icon]="'pt-icons-word-wrap_16'"></i>

<!-- out -->
<i [kbq-icon]="kbq-file-plus-o_16"></i>
<i [kbq-icon]="kbq-wrap-text_16"></i>
```

Angular now parses the binding as subtraction of undeclared identifiers (`kbq - file - plus - o_16`).

This branch is reached deliberately — `isDynamicExpression` on line 230 exists specifically to let quoted literals through — but nothing re-adds the quotes. The pre-PR fragment `{ replace: "'mc-", replaceWith: "'kbq-" }` produced the correct `"'kbq-file-plus-o_16'"`, so this is a regression.

The only dynamic-binding spec uses `[kbq-icon]="iconVar"`, which takes the early-out and never reaches this line.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+
+            return undefined;
+        };
+
+        for (const el of visitor.elementsToMigrate) {
+            for (const attr of el.attrs) {
+                const simpleName = getSimpleAttributeName(attr.name);
+
+                if (data.valueAttrs.includes(simpleName)) {
+                    const isDynamicExpression = attr.name.startsWith('[') && !/^\s*['"].*['"]\s*$/.test(attr.value);
```
lskramarov: **[blocking] The "is it a quoted literal" test uses a greedy `.*`, so string concatenations are rewritten instead of warned about.**

`/^\s*['"].*['"]\s*$/` matches any expression that merely *starts and ends* with a quote, because `.*` spans the interior.

Verified end-to-end:

```html
<!-- in  --> <i [kbq-icon]="'mc-' + name + '_16'"></i>
<!-- out --> <i [kbq-icon]="kbq- + name + _16"></i>
```

A template compile error, emitted silently with no `console.warn` — while the far safer `[kbq-icon]="iconVar"` control case *does* warn. The `'scope-' + iconName` shape is exactly the pattern the module doc-comment on line 327 cites as motivating the string-literal path, so it is a realistic input.

A correct check needs an actual expression parse, or at minimum an anchor that rejects interior quotes.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+                    if (isDynamicExpression) {
+                        console.warn(
+                            `Element is using a dynamic value. Check the code and change value on your own.${
+                                (fileName && ' File: ' + fileName) || ''
+                            }`
+                        );
+                        continue;
+                    }
+
+                    const cleanValue = getSimpleAttributeValue(attr.value).replace(/"/g, '');
```
lskramarov: **[blocking] Interpolation inside a plain icon attribute is destroyed.**

`getSimpleAttributeValue` strips `{` and `}` from anywhere in the value, so an interpolated value looks like a plain scope-prefixed token. `isDynamicExpression` (line 230) only guards `[`-prefixed names, so a plain attribute is unguarded.

Verified end-to-end:

```html
<!-- in  --> <i kbq-icon="mc-{{ name }}"></i>
<!-- out --> <i kbq-icon="kbq- name "></i>

<!-- in  --> <i kbq-icon="pt-icons-{{ iconName }}"></i>
<!-- out --> <i kbq-icon="kbq- iconName "></i>
```

The interpolation is gone and the attribute becomes a static string with embedded spaces — the icon silently stops rendering, no warning. The old blind regex produced the correct `kbq-{{ iconName }}`.

Related, same path: `<i kbq-icon="{{ 'mc-add-to-list_16' }}"></i>` is left untouched **and** omitted from the `fix: false` report, so a dry run reports the file as clean.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+ * `[class.X]` bindings, bare TS string literals, style-file regex): the bare scope word (mapped
+ * per `scopeClassInList`) plus `scope.from-<icon.from>` -> `scope.to-<icon.to>` for every icon.
+ */
+export function buildIconTokenMap(data: IconMigrationData): Map<string, string | null> {
+    const map = new Map<string, string | null>();
+
+    map.set(data.scope.from, data.scopeClassInList === 'remove' ? null : data.scope.to);
+
+    for (const icon of data.icons) {
+        map.set(`${data.scope.from}-${icon.from}`, `${data.scope.to}-${icon.to}`);
```
lskramarov: **[high] The token map only keys renames under the *old* scope, so icons already on `kbq-` are never renamed — which is how the entire data table is authored.**

`buildIconTokenMap` emits only `${scope.from}-${icon.from}` keys. But `newIconsPackData` entries are written in the new scope on both sides (`{ replace: 'kbq-add-to-list_16', replaceWith: 'kbq-file-plus-o_16' }`), and `stripScopePrefix` re-keys them to `mc-add-to-list_16`.

Verified end-to-end (`new-icons-pack`, `fix: true, updatePrefix: true`) — all of these come back byte-identical, and `fix: false` reports nothing:

```html
<i kbq-icon="kbq-add-to-list_16"></i>
<i class="kbq kbq-icon kbq-add-to-list_16"></i>
```
```scss
.kbq-add-to-list_16 { width: 16px; }
```

The old engine ran two passes — rescope, then rename in kbq-space — so these matched directly. `deprecated-icons` had the same shape via `` new RegExp(`kbq-${replace}`) ``.

Scope of the loss: all 391 `iconsMapping` entries and ~419 real renames in `newIconsPackData`. Every project that already ran the prefix half of the migration — precisely the population `updatePrefix: false` exists for — now gets a total no-op.

Both specs derive fixtures with `bareSuffix()`, which strips `kbq-` to build `mc-`/`pt-icons-` inputs, so nothing covers the case the data was written for.

lskramarov (commented):
## Review of the AST-based icon migration rewrite

The direction here is right — replacing blind whole-file text replacement with real parsing is the correct fix for the collateral-damage problem described in the PR body. The findings below are about the implementation, not the approach.

Everything reported was reproduced by actually running the schematics through `SchematicTestRunner` against a generated app, and by doing a clean `yarn run build:schematics` — none of it is inferred from reading. All 293 tests under `packages/schematics` pass on this branch, so none of these cases are currently covered.

---

### Blocking: the built package does not load

This one can't be attached inline because the file isn't in the diff — it's `packages/schematics/scripts/copy-meta-to-dist.js`.

`src/utils/icon-migration.ts` is registered in neither `rollup.config.js`'s `input` map nor the hand-maintained copy list in `copy-meta-to-dist.js` (which ends at `angular-parsing.js` and knows only five util files). Rollup therefore emits it as an anonymous shared chunk that is never shipped:

```
$ yarn run build:schematics          # exits 0, copy script reports no failures
$ ls packages/schematics/dist/*.js
packages/schematics/dist/icon-migration-5UXIgQLT.js

$ grep -n icon-migration dist/components/schematics/migrations/*/index.js
dist/components/schematics/migrations/deprecated-icons/index.js:8:  require('../../icon-migration-5UXIgQLT.js')
dist/components/schematics/migrations/new-icons-pack/index.js:9:  require('../../icon-migration-5UXIgQLT.js')

$ find dist/components -name 'icon-migration*'
                                     # nothing

$ node -e "require('./dist/components/schematics/migrations/new-icons-pack/index.js')"
Error: Cannot find module '../../icon-migration-5UXIgQLT.js'
```

`packages/components/package.json` publishes `./schematics/collection.json`, so every consumer running `ng g @koobiq/components:new-icons-pack` (or `deprecated-icons`) gets `MODULE_NOT_FOUND` on the first call. The build exits 0, so CI stays green.

Note that the chunk name is content-hashed, so hardcoding a sixth copy line would rot on the next edit to the util. Adding `'utils/icon-migration'` to the rollup `input` map gives it a stable path, and the copy list then needs one matching entry — or, better, derive the shipped util set from the emitted tree the way `getMigrations()` already derives the migration set.

---

### Inline comments

21 comments are attached to specific lines, tagged by severity:

- **blocking (4)** — cases where the migration writes code that no longer compiles: destroyed escape sequences in TS string literals, and quoted / concatenated / interpolated expressions in icon attributes losing their syntax.
- **high (8)** — cases where the migration silently doesn't do its job, or damages unrelated code: icons already on the `kbq-` prefix never being renamed (that's how the whole data table is authored), `updatePrefix` defaulting to `false` and skipping the entire styles pass, unusable custom-scope handling, ICU expansions never visited, inline `styles`/`host`/tagged templates dropped, and the bare scope word clobbering SCSS variables and Markdown prose.
- **medium (5)** — narrower false positives and silent gaps.
- **test (4)** — coverage gaps and assertions that cannot fail.

---

### Secondary findings

Not attached inline, and none of them blocking — listing them so they aren't lost.

**Reuse.** `escapeRegExp` in `icon-migration.ts:41` is the eighth copy of that function inside `packages/schematics/src` (`alert-signals`, `app-switcher-signals`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `filter-bar-signals`, `form-field-signals`, `v20-upgrade`), plus a byte-identical one exported from `packages/components/core/highlight/highlight-base.ts`. Since this PR is already creating a shared util module, that's a natural home for it. The `onFound` callback body is byte-identical between the two `index.ts` files and to `logReplacement` in `migrations/icons-replacement/index.ts`; the read → replace → report → overwrite block appears three times; and `IconAttributeCollector`'s traversal and no-op visitor methods are byte-identical to `ElementCollector` in `utils/typescript.ts`.

**Efficiency.** `replaceKnownIconTokens` compiles a fresh `RegExp` for every one of the ~572 token-map keys on every style and Markdown file, then compiles the identical regex a second time on line 442 for the `.replace`. Measured against a 7.5 KB SCSS file with the real map: ~453 ms vs ~1.1 ms per 200 files for a single alternation regex. Separately, both `index.ts` files eagerly `ts.createSourceFile(..., /* setParentNodes */ true)` over every `.ts` path into one array before processing, though `migrateIconsInTsFiles` consumes them strictly one at a time — measured at +200 MB retained on this repo's 1,743 files, which projects to roughly 1 GB on a 10k-file app. `migrateIconsInTsFiles` also re-reads content that `sourceFile.text` already holds, and walks the AST twice (`collectInlineTemplateRanges` then `forEachStringLiteral`) where one pass would do.

**Logging.** `console.warn` in `icon-migration.ts:233` bypasses the `context.logger` channel every other message in these migrations uses, so the one case that *requires* manual intervention never reaches the CLI's logger or a `SchematicTestRunner` subscriber. It also forces both specs to `jest.spyOn(console, 'warn')` because of the repo-wide `jest-fail-on-console` setup. `createIconTemplateTransform` already takes an `opts` bag, so passing the logger there is a small change.

**`bind-` forms.** `getSimpleAttributeName` doesn't strip the canonical `bind-` prefix, so `<i bind-kbq-icon="'mc-add-to-list_16'"></i>` is left untouched and unreported (verified). Three sibling migrations in this same package — `button-state-and-styles:33`, `button-supported-colors:39`, `dropdown-demote-overlay:40` — all do `getSimpleAttributeName(name).replace(/^bind-/, '')`, and two have dedicated specs for it.

**Dead extraction.** `migrateTsWithTransform` in `utils/angular-parsing.ts:77` was split out of `migrateTs` but has no caller other than the `migrateTs` wrapper itself — the icon path uses `migrateIconsInTsFiles` instead. That leaves two different AST walks for finding inline `@Component({template})` ranges living in the same utils folder (`analyzeDecorators` + `AnalyzedFile` vs `collectInlineTemplateRanges`), which already differ in sort order.

**Docs.** The README's own "with custom data" examples still say `--custom-icon-replacement-path replacement.json`, but the shipped `replacement.json` is in the legacy fragment shape that `resolveScope` now rejects with a warning — following the documented command silently ignores the file. And the comment on `new-icons-pack/index.ts:177` reads "styles in node_modules are still updated" directly above code that returns early on `node_modules`.

**Comment style.** A few of the new JSDoc blocks narrate the refactor rather than the current contract — `data.ts:7` in both migrations ("Unused since the migration moved to AST-based matching … these *were* whole-file regex fragments") and `resolveScope`'s opening lines ("historically accepted", "the old blind-text-replace engine", "The new engine"). `AGENTS.md` asks for no comments describing change history; the forward-looking halves of those blocks are the useful part.

---

_Generated with Claude Code_

packages/schematics/src/migrations/new-icons-pack/index.ts
```diff
@@ -34,78 +87,118 @@ export default function newIconsPack(options: Schema): Rule {
...
-            if (path.endsWith(stylesExt)) {
-                const initialContent = entry?.content.toString();
-                let newContent = initialContent;
+            if (changed) {
+                tree.overwrite(filePath, content);
+            }
+        }
+
+        // update styles
+        if (updatePrefix) {
```
lskramarov: **[high] `updatePrefix` now gates the *entire* styles pass, and its schema default is `false` — so the default invocation migrates and reports nothing in any stylesheet.**

The old code ran `handleDeprecatedIcons` on style files *outside* the `if (updatePrefix)` block, so icon-name renames and reporting happened regardless of the flag.

`schema.json` declares `"updatePrefix": { "default": false }` and `collection.json` wires that schema, so the `= true` destructuring default on line 74 never fires for a CLI run.

Verified end-to-end with `updatePrefix` omitted and `fix: true` — `styles.scss` containing both of these comes back unchanged, and the dry-run report is empty:

```scss
.kbq-add-to-list_16 { a: 1; }
.mc-add-to-list_16 { b: 2; }
```

So `ng g @koobiq/components:new-icons-pack --fix=true --project x` — the exact command in the README — silently skips stylesheets. Every spec passes `updatePrefix` explicitly, so nothing catches it.

The README's "`updatePrefix` now only gates the styles pass (as it always effectively did)" is not accurate — it previously gated only the prefix replace.

packages/schematics/src/migrations/new-icons-pack/index.ts
```diff
@@ -34,78 +87,118 @@ export default function newIconsPack(options: Schema): Rule {
...
-                    ]);
-                }
-            }
+        const migrationData: IconMigrationData = {
+            valueAttrs: ['kbq-icon', 'kbq-icon-item', 'kbq-icon-button'],
+            scope: resolvedScope,
+            scopeClassInList: 'rename',
+            icons: resolvedData.map(({ replace, replaceWith }) => ({
+                from: stripScopePrefix(replace, [resolvedScope.from, resolvedScope.to]),
+                to: stripScopePrefix(replaceWith, [resolvedScope.to])
```
lskramarov: **[high] A custom scope whose target is not `kbq` builds unusable token keys and produces half-migrated output.**

`stripScopePrefix(replaceWith, [resolvedScope.to])` strips only the *target* scope, so the hardcoded `kbq-` prefix baked into `newIconsPackData` is never stripped when the user names a different target.

Verified end-to-end with `customIconReplacementPath` = `[{"replace":"mc","replaceWith":"newpack"}]`:

```html
<!-- in -->
<i kbq-icon="mc-add-to-list_16"></i>
<i class="mc mc-add-to-list_16"></i>

<!-- out -->
<i kbq-icon="newpack-add-to-list_16"></i>   <!-- prefix swapped, deprecated NAME never renamed -->
<i class="newpack mc-add-to-list_16"></i>   <!-- icon class not migrated at all -->
```

The map key becomes `mc-kbq-add-to-list_16`, which matches nothing. Half-migrated markup is written to disk with no warning.

The option is documented as accepting any bare word pair, but works only when `replaceWith` is literally `kbq` — the single value the spec exercises.

packages/schematics/src/migrations/new-icons-pack/index.ts
```diff
@@ -15,16 +25,59 @@ function readJsonFile<T>(filePath: string): T {
     return JSON.parse(fileContents) as T;
 }
 
+const BARE_SCOPE_WORD = /^[\w-]+$/;
```
lskramarov: **[high] `BARE_SCOPE_WORD` is too permissive in two ways, and both failures are silent.**

**1. Trailing hyphen accepted.** `/^[\w-]+$/` accepts `mc-`, which is the most natural way to write a "prefix" pair and the shape the old engine used. Verified with `[{"replace":"mc-","replaceWith":"kbq-"}]`: `<i kbq-icon="mc-add-to-list_16"></i>` is returned unchanged and no legacy-format warning is emitted at all. `buildIconTokenMap` produces `mc--<suffix>` keys and `stripScopePrefix` tests `startsWith('mc--')`, which never fires. The user gets a completely silent no-op run.

**2. `undefined` passes the test.** `RegExp.prototype.test` coerces its argument, so `BARE_SCOPE_WORD.test(undefined) === true` (it tests the string `"undefined"`). An entry with a typo'd or missing key is accepted as a valid scope pair and skips the warning on line 58. Verified:

- `[{"replaceWith":"kbq"}]` → `TypeError: Cannot read properties of undefined (reading 'replace')`, thrown from `escapeRegExp` once `undefined` reaches the token map as a key
- `[{}]` → same crash
- a JSON object instead of an array → `TypeError: entries.find is not a function` (`readJsonFile` has no try/catch and `entries.find` on line 52 is unguarded)

So the schematic aborts with an opaque TypeError deep inside a regex helper instead of "your replacement file is malformed". Worth validating the *shape* of each entry (both keys present and `string`) and rejecting a leading/trailing hyphen.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+ * silently dropped/renamed on no real evidence of an icon reference.
+ */
+function resolveBareStringLiteral(
+    text: string,
+    tokenMap: Map<string, string | null>,
+    data: IconMigrationData
+): { result: string; found: TokenReplacement[] } | undefined {
+    const tokens = text.split(/\s+/).filter(Boolean);
+    const hasIconReference = tokens.some(
+        (token) => (tokenMap.has(token) && token !== data.scope.from) || data.icons.some(({ from }) => from === token)
```
lskramarov: **[high] A bare, unprefixed icon suffix is rewritten in any TS string literal, with no scope word and no icon context.**

The `data.icons.some(({ from }) => from === token)` arm matches a token that carries no scope prefix at all.

Verified end-to-end:

```ts
// new-icons-pack
export const analyticsEvent = 'add-to-list_16';   // -> 'file-plus-o_16'
// deprecated-icons
export const d = 'word-wrap_16';                  // -> 'wrap-text_16'
```

Neither literal contains a scope word or sits in any icon context — these are exactly the collateral rewrites this PR set out to eliminate. The old engine required the `kbq-` prefix, so a bare suffix was never touched; this is a new false positive, applied to every `.ts` file in the project including tests and config.

This arm exists only to support `'pt-icons word-wrap_16'`, so it could require the scope word to be present in the same literal.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+
+    visitBlock(block: Block, _1: any) {
+        this.visitChildren(block);
+    }
+
+    // part of interface
+    visitAttribute(_: Attribute, _1: any) {}
+    visitText(_: any, _1: any) {}
+    visitComment(_: any, _1: any) {}
+    visitExpansion(_: any, _1: any) {}
```
lskramarov: **[high] Elements inside ICU expansions are never visited.**

`visitExpansion` and `visitExpansionCase` are empty no-ops, and ICU children live on `expansion.cases[].expression` rather than `children`, so `visitChildren` never reaches them.

Verified end-to-end — this comes back byte-identical (`changed === false`) and is not reported in dry-run mode either:

```html
<span>{count, plural, =1 {<i kbq-icon="mc-add-to-list_16"></i>} other {<i class="mc mc-alert-enabled_16"></i>}}</span>
```

`parseTemplate` passes `tokenizeExpansionForms: true`, so the nodes are in the tree and simply not descended into. I confirmed by contrast that `<ng-template>`, `<ng-container>`, `@if`, `@for`/`@empty` and `@defer`/`@placeholder` all migrate correctly — ICU is the only structural hole. The old whole-file regex reached this text.

```ts
visitExpansion(expansion: any) { visitAll(this, expansion.cases); }
visitExpansionCase(c: any) { visitAll(this, c.expression); }
```

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+    sourceFiles: ts.SourceFile[],
+    context: SchematicContext,
+    data: IconMigrationData,
+    tokenMap: Map<string, string | null>,
+    opts: { fix: boolean; onFound: (fileName: string, found: TokenReplacement[]) => void }
+): Promise<void> {
+    for (const sourceFile of sourceFiles) {
+        const relativePath = sourceFile.fileName;
+        const content = tree.readText(relativePath);
+        const templateRanges = collectInlineTemplateRanges(sourceFile);
```
lskramarov, line 603: **[high] Inline `styles: [...]`, `host` class bindings, and `${}`-substituted template literals are no longer migrated.**

`collectInlineTemplateRanges` collects only the `template` property and only non-substituted literals; everything else falls through to `resolveBareStringLiteral`, whose whitespace-token split never matches a CSS selector.

Verified end-to-end on one component (`new-icons-pack`, `fix: true, updatePrefix: true`) — all three left untouched:

```ts
@Component({
  host: { class: 'mc kbq-icon mc-add-to-list_16', '[class.mc]': 'true' },  // '[class.mc]' untouched
  styles: ['.mc-add-to-list_16 { a: 1; }'],                                // untouched
  template: `<i kbq-icon="mc-add-to-list_16"></i>${1}`                     // untouched
})
```

`.ts` files are excluded from the styles pass by the `else if` chain in both index.ts files, and the `class.X` handling on line 272 lives only in the HTML-AST path, which never sees a `host` object. `ts.isStringLiteralLike` excludes `TemplateHead`/`TemplateSpan`.

The old fragments `pt-icons-`→`kbq-` and `[class.mc`→`[class.kbq` handled all three. Net effect: a component's template gets new icon names while its own inline styles and host classes keep the old ones.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+export function replaceKnownIconTokens(
+    content: string,
+    tokenMap: Map<string, string | null>,
+    fix: boolean
+): { content: string; changed: boolean; found: TokenReplacement[] } {
+    const found: TokenReplacement[] = [];
+    let result = content;
+
+    for (const [token, replacement] of tokenMap) {
+        const pattern = new RegExp(`(?<![\\w-])${escapeRegExp(token)}(?![\\w-])`, 'g');
```
lskramarov: **[high] The bare scope word is matched with only a word/hyphen boundary, so it rewrites far more than class selectors in SCSS and Markdown.**

Verified end-to-end (`new-icons-pack`, `fix: true, updatePrefix: true`):

```scss
/* in */                      /* out */
$mc: red;                     $kbq: red;
.a { color: $mc; }            .a { color: $kbq; }
@import "mc";                 @import "kbq";
.b { font-family: mc; }       .b { font-family: kbq; }
```

A broken stylesheet: `$kbq` is undefined and `kbq.scss` does not exist. `deprecated-icons` behaves the same for `@import "pt-icons"` / `$pt-icons` / `font-family: pt-icons`.

The `.md` pass hits every Markdown file in the tree: prose reading `The mc compiler ... use mc as an abbreviation` becomes `The kbq compiler ... use kbq as an abbreviation`.

The old style pass anchored on `mc-` or `.pt-icons`, so none of these matched. The README documents this exception as covering only comments, which understates it considerably. A cheap fix that keeps the regex approach: require a leading `.` for the bare scope token.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+                            end: attr.valueSpan.end.offset,
+                            replacement: resolved
+                        });
+                        allFound.push({ from: cleanValue, to: resolved });
+                    }
+
+                    continue;
+                }
+
+                if (simpleName === 'class') {
```
lskramarov: **[medium] `[class]="expr"` is run through the static class-list path, rewriting the binding expression.**

`getSimpleAttributeName('[class]')` returns `'class'`, and unlike the `valueAttrs` branch 26 lines above, this one has no `isDynamicExpression` guard — so the Angular expression source is split on whitespace and spliced over `attr.valueSpan`.

Verified end-to-end (`new-icons-pack`, where `tokenMap` holds `mc -> kbq`):

```html
<!-- in  --> <div [class]="mc"></div>
<!-- out --> <div [class]="kbq"></div>
```

That renames a component property reference that does not exist, silently breaking the binding.

The guard is effectively backwards here: the quoted forms are the ones that get *missed*, and both were migrated before this PR:

```html
<div [class]="'mc-add-to-list_16'"></div>             <!-- unchanged, not reported -->
<div [ngClass]="{ 'mc-add-to-list_16': true }"></div> <!-- unchanged, not reported -->
```

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+
+                    continue;
+                }
+
+                const bindingToken = getClassBindingToken(simpleName);
+
+                if (bindingToken !== undefined && tokenMap.has(bindingToken)) {
+                    const replacement = tokenMap.get(bindingToken);
+
+                    if (replacement) {
```
lskramarov, line 352: **[medium] A `[class.<scope>]` binding is skipped *and* never reported, because `if (replacement)` conflates "maps to removal" with "no mapping".**

For `deprecated-icons`, `scopeClassInList: 'remove'` makes `buildIconTokenMap` store `pt-icons -> null`, so the truthiness check drops out of the branch — and `allFound.push` sits inside that same `if`.

Verified end-to-end: `<div [class.pt-icons]="a"></div>` is returned unchanged by `--fix`, while the sibling `class="pt-icons"` in the same file is correctly emptied to `class=""`. In `fix: false` mode nothing is pushed into `allFound`, so the audit run reports the file as clean and gives zero signal that a stale scope class survived.

If a binding genuinely cannot be dropped automatically, it still needs to appear in the report.

Adjacent hazard on the same path: `ast.ts:51` declares `valueSpan` as non-optional, but the real Angular AST leaves it `undefined` for valueless attributes. The unguarded `attr.valueSpan.start.offset` derefs on lines 246/262 survive only because `resolveValueToken('')` and `replaceTokenList('')` happen to bail first.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+            if (tokenMap.has(token)) {
+                const replacement = tokenMap.get(token)!;
+
+                changed = true;
+                found.push({ from: token, to: replacement ?? '' });
+
+                return replacement;
+            }
+
+            const icon = data.icons.find(({ from }) => from === token);
```
lskramarov, line 219: **[medium] Identity mappings are treated as real renames, producing false "deprecated icon" reports and pointless file rewrites.**

There is no `token !== icon.to` check before `changed = true`. 152 of the 571 `newIconsPackData` entries map an icon to itself (`kbq-align-left_16` → `kbq-align-left_16`, `kbq-bell_16`, `kbq-api_16`, …), which `stripScopePrefix` turns into `align-left_16 -> align-left_16` in `data.icons`.

Verified end-to-end with `export const a = 'align-left_16   trailing';`:

- `fix: false` logs `align-left_16 -> align-left_16` — a deprecation notice for a no-op
- `fix: true` sets `changed = true`, so the literal is spliced anyway and comes back as `'align-left_16 trailing'` — the file is mutated (with the whitespace/escape damage noted on line 381) for zero semantic change

Every such literal produces a spurious diff hunk in the user's `ng update` review.

packages/schematics/src/migrations/new-icons-pack/index.ts
```diff
@@ -34,78 +87,118 @@ export default function newIconsPack(options: Schema): Rule {
...
 
-                    if (initialContent !== newContent) {
-                        tree.overwrite(path, newContent || '');
-                    }
-                }
+        await migrateIconsInTsFiles(tree, sourceFiles, context, migrationData, tokenMap, { fix, onFound });
+
+        // Markdown docs have no AST here, so — like styles — they're matched with the same
+        // word/selector-boundary-safe, known-token-only regex rather than an unscoped replace.
+        for (const filePath of markdownPaths) {
```
lskramarov, line 201: **[medium] The `.md` pass is not gated by `updatePrefix` while the styles pass is, so the default option set rewrites changelogs and leaves the stylesheets they describe alone.**

Verified end-to-end with `updatePrefix` omitted (its `schema.json` default is `false`) and `fix: true`:

```
CHANGELOG.md  "## 1.0.0\n- renamed `mc-add-to-list_16`\n"
           -> "## 1.0.0\n- renamed `kbq-file-plus-o_16`\n"

styles.scss   ".mc-add-to-list_16 { a: 1; }"
           -> unchanged
```

So the migration rewrites historical release notes — falsifying what a past version actually shipped — while producing docs that contradict the un-migrated styles.

`.md` should be gated the same way as styles, and changelog-like files are arguably worth excluding outright.

packages/schematics/src/migrations/new-icons-pack/index.ts
```diff
@@ -34,78 +87,118 @@ export default function newIconsPack(options: Schema): Rule {
...
+        const templatePaths: string[] = [];
+        const tsPaths: string[] = [];
+        const markdownPaths: string[] = [];
+
         // Update templates & components
-        targetDir.visit((path: Path, entry) => {
+        targetDir.visit((filePath: Path) => {
             // if project property not provided, skip files in node_modules & dist
-            if (path.includes('node_modules') || path.includes('dist')) {
+            if (filePath.includes('node_modules') || filePath.includes('dist')) {
```
lskramarov: **[medium] `includes('dist')` is a substring match, not a path-segment match, so real source directories are silently dropped.**

Verified end-to-end — both come back unmigrated and unreported:

```
src/app/distribution/list.html   <i kbq-icon="mc-add-to-list_16"></i>
src/app/edit-distance.ts         export const c = 'mc-add-to-list_16';
```

Same for `distinct/`, `district/`, `distributor.component.html`, or any project whose root path happens to contain `dist`.

The guard predates this PR (and `deprecated-icons/index.ts:67` has the identical line), but it is now the only thing standing between the AST engine and the file, and unlike a parse failure it produces no log line at all — the user has no way to notice the gap.

```ts
if (/(^|\/)(node_modules|dist)(\/|$)/.test(filePath)) { return; }
```

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
```diff
@@ -2,36 +2,38 @@ import { workspaces } from '@angular-devkit/core';
...
+    .filter(({ replace, replaceWith }) => replace !== replaceWith)
+    .map(({ replace, replaceWith }) => ({ replace: bareSuffix(replace), replaceWith: bareSuffix(replaceWith) }));
+
 /**
  * `@schematics/angular:application` changed file names across major versions
  * (`app.component.html` ↔ `app.html`). `createTestApp` pins to v17, so the
  * default is `app.component.html` — but pick whichever the generator produced.
  */
-const appHtml = (project: workspaces.ProjectDefinition, tree: Tree | UnitTestTree) => {
+const projectPaths = (project: workspaces.ProjectDefinition, tree: Tree | UnitTestTree) => {
```
lskramarov, line 24: **[test] This spec has no coverage of the TS pass or the `.md` pass at all.**

`projectPaths` returns only `html` and `styles` — there are zero `p.ts` references in the file (`grep -c "p\.ts"` → 0), and no test writes a `.md` file. So `migrateIconsInTsFiles`, `collectInlineTemplateRanges`, `forEachStringLiteral` and the entire markdown loop are unasserted for this schematic.

Both paths are live and both have confirmed bugs (see the comments on `icon-migration.ts:381`, `:144`, `:165` and `index.ts:156`).

The sibling `deprecated-icons` spec does expose a `ts` path — worth mirroring that here.

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
```diff
@@ -40,106 +42,302 @@ describe(SCHEMATIC_NAME, () => {
...
+            const tree = await runner.runSchematic(
+                SCHEMATIC_NAME,
+                { fix: true, project: currentProjectKey, updatePrefix: true },
+                appTree
+            );
+
+            expect(tree.read(p.html)!.toString()).toBe(original);
+        });
+
+        it('leaves an unrelated string literal untouched', async () => {
```
lskramarov: **[test] This test doesn't exercise what its name says.**

The fixture is an HTML comment written into `p.html`, so it goes through `IconAttributeCollector.visitComment` — a no-op — not through `resolveBareStringLiteral`. The assertion cannot fail no matter what the string-literal path does.

To actually cover "unrelated string literal", the fixture needs to be a TS string literal in a `.ts` file, e.g. `export const mode = 'mc-legacy-mode';`. Written that way it would catch the false positive reported on `icon-migration.ts:144`.

packages/schematics/src/migrations/deprecated-icons/index.spec.ts
```diff
@@ -134,124 +154,231 @@ class TestApp {
...
+
+            const tree = await runner.runSchematic(
+                SCHEMATIC_NAME,
+                { project: currentProjectKey, fix: true, stylesExt: '.less' },
+                appTree
             );
+
+            expect(tree.read(lessPath)?.toString()).toBe('.kbq { }');
+            // The default .scss file is untouched since it wasn't targeted by this run.
+            expect(tree.read(p.styles)?.toString()).not.toContain(DEPRECATED_SCOPE + ':');
```
lskramarov: **[test] Vacuous assertion — the needle appears in no fixture in this file.**

This `describe`'s `beforeEach` never writes `p.styles`, so it still holds the generator's `/* You can add global styles to this file... */`. The needle is `'pt-icons:'` (with a colon), which no fixture in this spec ever contains.

The assertion passes for a file that never held the token, and would pass equally if the schematic *had* rewritten it. Capturing the original content and asserting `toBe(original)` would test the intent stated in the comment above it.

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
```diff
@@ -40,106 +42,302 @@ describe(SCHEMATIC_NAME, () => {
...
+                appTree
+            );
+
+            expect(tree.read(p.html)!.toString()).toBe(html);
+            expect(warnSpy).toHaveBeenCalled();
 
-            expect(
-                elementsWithDeprecatedIconPrefixes.filter((item) => templateContent.indexOf(item) !== -1).length
-            ).toBeFalsy();
+            warnSpy.mockRestore();
```
lskramarov: **[test] The `console.warn` spy is restored only on the success path, and the jest config doesn't cover the gap.**

`warnSpy.mockRestore()` sits after `expect(...)` calls that can throw, so a failure leaves the silent implementation installed on `console.warn`. `jest.config.js:13` sets `clearMocks: true` but **not** `restoreMocks`/`resetMocks`, so `mockClear()` between tests only wipes `mock.calls` — the stub survives for the remainder of the file and mutes any later test that would have surfaced a warning.

Same pattern at `deprecated-icons/index.spec.ts:381`.

`afterEach(() => warnSpy.mockRestore())` (or `restoreMocks: true` in the config) fixes it.

Separately: `expect(warnSpy).toHaveBeenCalled()` asserts nothing about *which* warning fired — the migration's message text is never checked in either spec.

packages/schematics/src/utils/icon-migration.ts
```diff
@@ -0,0 +1,447 @@
...
+                        console.warn(
+                            `Element is using a dynamic value. Check the code and change value on your own.${
+                                (fileName && ' File: ' + fileName) || ''
+                            }`
+                        );
+                        continue;
+                    }
+
+                    const cleanValue = getSimpleAttributeValue(attr.value);
+                    const resolved = resolveValueToken(cleanValue);
```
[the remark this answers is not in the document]
NikGurev: Fixed as part of a bigger rewrite: quoted values (both single and double quotes) are now parsed properly, and the replacement keeps the original quotes instead of the value being stripped by hand.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. The migration now edits the raw text of the string, not the decoded value, so escaped characters like \t or \' are left alone.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. When the value is a quoted literal, the replacement is wrapped back in the same quote character, so the binding stays valid instead of turning into a bare identifier expression.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. The check now rejects any value that has a second matching quote in the middle (like a `+` concatenation), so it's treated as dynamic and skipped with a warning instead of being silently rewritten.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. Any value containing `{{ }}` is now skipped with a warning instead of being rewritten and broken.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. The lookup table now also has an entry for the icon under the *new* scope prefix, so an icon that's already on `kbq-` still gets its name corrected instead of being skipped as a no-op.

packages/schematics/src/migrations/new-icons-pack/index.ts
NikGurev: Fixed. Icon-name renames in styles/markdown now always run regardless of the option. `updatePrefix` only gates the extra "rename the bare mc/kbq word" part, matching what it did historically.

packages/schematics/src/migrations/new-icons-pack/index.ts
NikGurev: Fixed. The old icon names in the data file are always written with a `kbq-` prefix, so that's what gets stripped first now (with a fallback to the target scope), regardless of what scope you're migrating to.

packages/schematics/src/migrations/new-icons-pack/index.ts
NikGurev: Fixed. The check now rejects a trailing hyphen and validates both fields are real strings before accepting an entry, so a malformed file gives a clear warning instead of a crash or a silent no-op.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. A bare icon suffix in a TS string is only renamed if the scope word (mc/pt-icons) also shows up somewhere in that same string, so a plain unrelated string isn't touched anymore.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. The visitor now walks into ICU expansion cases, so icons inside `{count, plural, ...}` blocks are found and migrated too.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Partly fixed: `host` class bindings and the inline `styles` array on a component decorator are now migrated. A template literal with `${...}` inside it (as opposed to a plain string) is still not supported — calling that out as a known gap rather than claiming it's covered.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. In styles/markdown, the bare scope word now only matches right after a `.`, like a real class selector, so `$mc`, `@import 'mc'`, `font-family: mc`, etc. are left alone.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. A bound `[class]` expression is now left alone entirely — only a plain, unbound `class` attribute goes through the class-list rewrite.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. "Maps to removal" is now handled separately from "no match found" — it's both reported and actually removed, instead of being silently left in place either way.

packages/schematics/src/utils/icon-migration.ts
NikGurev: Fixed. A pair that maps to itself is no longer added to the lookup table at all, so it can't show up as a fake "change" in a report or get rewritten for nothing.

packages/schematics/src/migrations/new-icons-pack/index.ts
NikGurev: Fixed by making styles and markdown behave the same: icon-name renames always apply, and only the bare-word rename is gated by `updatePrefix`, in both places.

packages/schematics/src/migrations/new-icons-pack/index.ts
NikGurev: Fixed. It now checks for `/dist/` with slashes on both sides (matching how the other migrations in this repo already do it), so `distribution/`, `district/`, etc. aren't caught by accident.

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
NikGurev: Fixed. Added tests that write real `.ts` and `.md` fixtures and check the migration actually touches them, for both the positive and negative cases.

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
NikGurev: Fixed. The test now uses a real `.ts` string literal instead of an HTML comment, so it actually exercises the bare-string-literal code path it's named after.

packages/schematics/src/migrations/deprecated-icons/index.spec.ts
NikGurev: Fixed. The test now captures and compares against the actual original file content instead of checking for a string that was never in the fixture to begin with.

packages/schematics/src/migrations/new-icons-pack/index.spec.ts
NikGurev: Fixed two ways: the spy is now set up/torn down in beforeEach/afterEach so a failing assertion can't leak it into later tests, and the warning itself no longer goes through `console.warn` at all — it goes through the schematic's own logger now, checked the same way the rest of these tests already check logger output.

NikGurev:
Went through this review in full and fixed what was really broken. Quick summary:

**Fixed (blocking, from the review body):** `icon-migration.ts` wasn't wired into the build (`rollup.config.js`) or the copy step (`copy-meta-to-dist.js`), so the published package would crash with `MODULE_NOT_FOUND` on first use. Both are fixed now, and the copy step pulls every compiled util file automatically instead of a hand-written list, so this can't silently happen again for the next new util file.

**Not actually a bug:** the suppressed note about edits being applied in ascending order corrupting offsets in `migrateIconsInTsFiles`. That path uses `Tree`'s built-in `UpdateRecorder` (backed by `magic-string`), which tracks every edit against the *original* file offsets no matter what order you call it in — so ascending order there is safe. (The other migration that sorts descending is a different, hand-rolled string-splicing path where order really does matter.)

**Decided not to do:**
- `bind-` prefix support (`bind-class`, etc.) — not worth the added complexity for how rarely it's actually used.
- The "shared code should be deduplicated" and memory/performance notes (escapeRegExp copies, eagerly building every TS source file up front, etc.) — real observations, but a separate cleanup pass, not correctness fixes. Left as-is for now.

Everything else is fixed — replied on each comment with what changed.

NikGurev:
> bind- forms. getSimpleAttributeName doesn't strip the canonical bind- prefix, so <i bind-kbq-icon="'mc-add-to-list_16'"></i> is left untouched and unreported (verified). Three sibling migrations in this same package — button-state-and-styles:33, button-supported-colors:39, dropdown-demote-overlay:40 — all do getSimpleAttributeName(name).replace(/^bind-/, ''), and two have dedicated specs for it.


@lskramarov впервые вижу  такой байндинг. Обычно [kbq-icon]="" и kbq-icon=""

NikGurev:
Follow-up on the "Efficiency" part of the review that I'd previously left as-is: implemented now.

- `replaceKnownIconTokens` no longer compiles a regex per token per file. Patterns are built once per run (`buildIconTokenPatterns`) and reused across every style/markdown file and every inline `styles` array.
- Reusing the same regex objects surfaced a real bug worth calling out: `RegExp.prototype.test` keeps `lastIndex` between calls (unlike `String.replace`, which resets it), so testing the same pattern against different files' content could silently miss matches depending on file order. Fixed by resetting `lastIndex` before each use.
- `migrateIconsInTsFiles` now parses each `.ts` file's `SourceFile` lazily, one at a time, instead of the caller mapping the whole file list into an array of parsed ASTs up front.

Verified: `tsc`, `eslint`, `prettier` clean, and all 50 tests in both migration spec files plus the 3 sibling migrations sharing `angular-parsing.ts` still pass. Pushed in 4355f0a88.

Still left as a separate cleanup, not done here: the `escapeRegExp`/`onFound`/traversal duplication noted under "Reuse", and merging the two AST walks (`collectInlineTemplateRanges` + `forEachStringLiteral`) into one pass.

lskramarov (approved):

PR #1932, opened 2026-08-24
fix(inline-edit): show tooltip when out of viewport correctly (#DS-5420)

NikGurev:
## Summary

Scroll into view when form field is invalid.

Also fixed regression: on invalid input and tooltip is shown, hide tooltip when value changed.


packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +394,53 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+        const showTooltip = (): void => {
+            if (shown) return;
+
+            shown = true;
+            this.window.removeEventListener('scrollend', showTooltip);
+            this.tooltipTrigger()?.updatePosition();
+            this.tooltipTrigger()?.show();
+        };
+
+        this.window.addEventListener('scrollend', showTooltip, { once: true });
```
artembelik: листенер нужно вызывать вне ангуляр зоны - `runOutsideAngular`

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +394,53 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+        setTimeout(showTooltip, VALIDATION_TOOLTIP_SCROLL_TIMEOUT);
```
artembelik: нужно сохранять ссылку на таймаут и очищать при последующих вызовах `showValidationTooltip`, тк возможны сценарии при которых в макрозадачах будет несколько таймаутов, например если вызвать несколько раз подряд `showValidationTooltip`

packages/components/inline-edit/inline-edit.ts
```diff
@@ -56,8 +57,26 @@ const KBQ_INLINE_EDIT_ACTION_BUTTONS_ANIMATION = trigger('panelAnimation', [
...
+/**
+ * The validation tooltip is anchored to an element inside the edit-mode overlay — an overlay inside another
+ * overlay. Its default `close()` scroll strategy would close it the moment either overlay's origin scrolls
+ * (including the programmatic scroll `save()` uses to bring an invalid control into view), so it's overridden
+ * to `reposition()` here, scoped to just this tooltip instance.
+ */
+const KBQ_INLINE_EDIT_VALIDATION_TOOLTIP_SCROLL_STRATEGY = {
+    provide: KBQ_TOOLTIP_SCROLL_STRATEGY,
+    deps: [Overlay],
+    useFactory: (overlay: Overlay) => () => overlay.scrollStrategies.reposition()
```
artembelik, line 70: такое ощущение что из тултипа получился поповер, возможно стоит его использовать?

artembelik (approved):

rmnturov:
Трабл со скроллом и шапкой доков
image

packages/components/inline-edit/inline-edit.ts
rmnturov: Если что, я не настаиваю, чтобы это было именно тултипом. Это я в спеке, возможно, небрежно выразился. Суть в том, что сообщение об ошибке всплывающее рядом с полем, а не инлайновое.

rmnturov (changes requested):

NikGurev:
> Трабл со скроллом и шапкой доков 

@rmnturov сейчас такое же поведение для всех выпадашек и всплывашек, надо подумать

packages/components/inline-edit/inline-edit.ts
NikGurev: На мой взгляд, в данном случае проще использовать error тултип вместо покраски поповера 

можем обсудить на созвоне

packages/components/inline-edit/inline-edit.ts
NikGurev: Добавил

packages/components/inline-edit/inline-edit.ts
NikGurev: Добавил

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            if (timeoutId !== null) {
+                clearTimeout(timeoutId);
+                timeoutId = null;
+            }
+
+            detachSubscription?.unsubscribe();
+            detachSubscription = null;
+        };
+
+        const showTooltip = (): void => {
```
lskramarov: **Bug — the deferred tooltip still fires after the control became valid.**

`showTooltip` never re-checks `isInvalid()`. The `stateChanges` subscription added in `onAttach` only hides an *already open* tooltip (`if (tooltipTrigger?.isOpen)`), and between the Save click and `scrollend` the tooltip is not open yet, so nothing cancels the pending show.

Repro: invalid control that is off screen -> click Save -> while the smooth scroll is running, type a valid value -> `scrollend` fires -> "Required field" is shown over a now-valid control. I verified this with a probe spec on this branch: `show()` call count after the control became valid is `1` (expected `0`).

This is the same regression the PR description says it fixes, just moved into the new async path.

Suggested fix: re-check `this.isInvalid() && this.showTooltipOnError() && this.validationTooltip()` inside `showTooltip`, or have the `stateChanges` handler run the pending `cleanup()`.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            this.tooltipTrigger()?.updatePosition();
+            this.tooltipTrigger()?.show();
+        };
+
+        detachSubscription = this.overlayDir()
+            .overlayRef.detachments()
+            .subscribe(() => cleanup());
+
+        this.ngZone.runOutsideAngular(() => {
+            this.window.addEventListener('scrollend', showTooltip, { once: true });
```
lskramarov: **Bug — `scrollend` on `window` never fires for a nested scroll container.**

`scrollend` only bubbles when it is fired at the document. When `scrollIntoView` scrolls an `overflow: auto` ancestor (scrollable table body, modal body, `kbq-scrollbar` viewport), the event is dispatched at that element and never reaches this bubble-phase listener on `window`.

Verified with a probe spec: after an element-targeted `scrollend`, `show()` calls = 0; the tooltip only appeared after `tick(800)`. With the tooltip's own 400 ms `kbqEnterDelay` on top, that is ~1.2 s in what is probably the most common layout for inline edit.

Suggested fix: register in the capture phase — capture listeners on `window` do receive element-targeted non-bubbling events — and remove with the same flag:

```ts
this.window.addEventListener('scrollend', showTooltip, { once: true, capture: true });
// ...
this.window.removeEventListener('scrollend', showTooltip, { capture: true });
```

Alternatively subscribe through the CDK `ScrollDispatcher` that `KbqPopUpTrigger` already injects.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -56,8 +58,26 @@ const KBQ_INLINE_EDIT_ACTION_BUTTONS_ANIMATION = trigger('panelAnimation', [
     ])
 ]);
 
+/**
+ * The validation tooltip is anchored to an element inside the edit-mode overlay — an overlay inside another
+ * overlay. Its default `close()` scroll strategy would close it the moment either overlay's origin scrolls
+ * (including the programmatic scroll `save()` uses to bring an invalid control into view), so it's overridden
+ * to `reposition()` here, scoped to just this tooltip instance.
+ */
+const KBQ_INLINE_EDIT_VALIDATION_TOOLTIP_SCROLL_STRATEGY = {
```
lskramarov: **Bug — this override is not scoped to the validation tooltip.**

The doc comment says "scoped to just this tooltip instance", but the provider sits in the component's `providers`, and element injectors resolve through the declaration tree — so **every `kbqTooltip` in content projected into `kbq-inline-edit`** inherits `reposition()` too.

Verified with a probe spec: a tooltip projected into `[kbqInlineEditViewMode]` reports `RepositionScrollStrategy`, while an identical tooltip declared outside the inline edit reports `CloseScrollStrategy`. The same applies to tooltips on `kbq-option`s of a select projected into edit mode. Those consumer tooltips silently stop closing on scroll — a behaviour change the PR does not mention.

Suggested fix: scope the provider to the tooltip's own host (a small `@Directive` with `providers` applied to `.kbq-inline-edit__control-container`), or add a scroll-strategy input to `KbqTooltipTrigger`. Either way the comment needs correcting.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+     * Shows the validation tooltip, scrolling the invalid control into view first if it isn't fully visible.
+     *
+     * `scrollIntoView({ behavior: 'smooth' })` finishes asynchronously — for a long scroll distance it can take
+     * a while — so showing the tooltip right away would anchor its overlay to a stale, pre-scroll position.
+     * `scrollend` reports when the scroll actually completes;
+     * the timeout is a fallback for browsers without `scrollend` support and cases where nothing ends up scrolling.
+     */
+    private showValidationTooltip(): void {
+        const rect = this.overlayOrigin.getBoundingClientRect();
+        const isFullyVisible =
```
lskramarov: **Bug — the visibility test ignores clipping by scrollable ancestors.**

`getBoundingClientRect()` does not account for an ancestor's `overflow` clipping, so a control that is fully hidden inside a scrolled container is still reported as visible whenever its rect happens to land inside the window box.

Example: container occupies viewport `y = 100..500`, the inline-edit row is scrolled past its bottom edge and renders at `y = 600` in an 800 px window. `rect.top >= 0` and `rect.bottom (700) <= 800`, so `isFullyVisible` is `true`, `scrollIntoView` is skipped and `show()` fires immediately — the tooltip is anchored to a control the user cannot see. That is exactly the case DS-5420 is about.

Suggested fix: intersect the rect with the rects of `ScrollDispatcher.getAncestorScrollContainers(this.overlayOrigin)` instead of only the viewport (or use `Element.checkVisibility` / `IntersectionObserver`).

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            cleanup();
+            this.tooltipTrigger()?.updatePosition();
+            this.tooltipTrigger()?.show();
+        };
+
+        detachSubscription = this.overlayDir()
+            .overlayRef.detachments()
+            .subscribe(() => cleanup());
+
+        this.ngZone.runOutsideAngular(() => {
```
lskramarov, line 504: **`show()` ends up running outside the Angular zone.**

The registration is wrapped in `runOutsideAngular`, but the *handler* does full Angular work: `show()` attaches the overlay, emits the `visibleChange` / `placementChange` outputs and schedules the `enterDelay` timer — all outside the zone. Probe spec on this branch: `NgZone.isInAngularZone()` inside `show()` on this path is `false`.

Note that `KbqPopUpTrigger.hide()` explicitly wraps its work in `this.ngZone.run(...)` for precisely this reason, while `show()` has no such guard. Consumer handlers bound to the tooltip's outputs will run outside the zone and schedule no change detection. Angular 20's hybrid scheduler still reacts to `markForCheck`, which is why the unit tests pass — but the asymmetry with `hide()` looks unintended.

Suggested fix: keep `runOutsideAngular` for the registration, wrap the handler body in `this.ngZone.run(...)`.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            this.window.addEventListener('scrollend', showTooltip, { once: true });
+        });
+        timeoutId = setTimeout(showTooltip, VALIDATION_TOOLTIP_SCROLL_TIMEOUT);
```
lskramarov: **The fallback can fire before a long smooth scroll finishes.**

The timer is armed before the scroll starts and the constant is not derived from the scroll distance. On a long page `scrollIntoView({ behavior: 'smooth' })` easily exceeds 800 ms; the fallback then runs mid-flight, sets `shown` and `cleanup()` removes the listener — so the real `scrollend` is discarded and the position is never corrected. That is the exact staleness this method exists to avoid.

`RepositionScrollStrategy` masks it only while the scroll is broadcast through `ScrollDispatcher` (window scroll, or a registered `CdkScrollable`); an unregistered `overflow: auto` ancestor leaves the tooltip stuck at the mid-scroll position.

Suggested fix: on the fallback path keep listening and run `updatePosition(true)` when `scrollend` eventually arrives, instead of dropping it.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            rect.bottom <= this.window.innerHeight &&
+            rect.right <= this.window.innerWidth;
+
+        if (isFullyVisible) {
+            this.tooltipTrigger()?.show();
+
+            return;
+        }
+
+        let shown = false;
```
lskramarov, line 459: **Repeated `save()` stacks independent listeners, timers and scroll animations.**

All state lives in call-local closures, so each invocation arms its own `scrollend` listener, its own 800 ms timer and its own detach subscription, and re-issues `scrollIntoView`. While the control is invalid `save()` can fire repeatedly — from the Save button and from `onOverlayOutsideClick` on *every* outside click.

Two outside clicks 200 ms apart therefore restart the smooth scroll from the current mid-scroll offset (visible stutter), and the first pair's `cleanup()` only tears down its own resources.

Suggested fix: hold `shown` / `timeoutId` / `detachSubscription` as instance fields and no-op (or reset) when a request is already in flight.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            detachSubscription?.unsubscribe();
+            detachSubscription = null;
+        };
+
+        const showTooltip = (): void => {
+            if (shown) return;
+
+            shown = true;
+            cleanup();
+            this.tooltipTrigger()?.updatePosition();
```
lskramarov: **Redundant — `show()` calls `updatePosition()` itself.**

`updatePosition(false)` only swaps the preferred-positions list; it never re-applies the position. And `KbqPopUpTrigger.show()` calls `this.updatePosition()` internally right before `instance.show(delay)`, so this line adds nothing but an extra `createOverlay()` plus a teardown/rebuild of the closing-actions subscription. A probe counted 3 `updatePosition` calls per show on this path.

It also reads as if the position is being refreshed after the scroll, which it is not — if that refresh is actually wanted it has to be `updatePosition(true)`. Otherwise just drop the line.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -371,14 +396,73 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
 
+    /**
+     * Shows the validation tooltip, scrolling the invalid control into view first if it isn't fully visible.
+     *
+     * `scrollIntoView({ behavior: 'smooth' })` finishes asynchronously — for a long scroll distance it can take
+     * a while — so showing the tooltip right away would anchor its overlay to a stale, pre-scroll position.
+     * `scrollend` reports when the scroll actually completes;
+     * the timeout is a fallback for browsers without `scrollend` support and cases where nothing ends up scrolling.
+     */
+    private showValidationTooltip(): void {
```
lskramarov, line 448: **Altitude — this is generic pop-up infrastructure living inside one component.**

"Scroll an origin into view, wait for the scroll to settle, then open an overlay anchored to it" is not inline-edit-specific. As written, any other component that needs it (popover, dropdown, other validation surfaces) has to copy the whole `scrollend` / timeout / cleanup dance *and* re-override `KBQ_TOOLTIP_SCROLL_STRATEGY` — and every bug fixed here would have to be fixed again in each copy.

The repo already centralises this class of concern: `packages/components/core/overlay/auto-hide-scroll-strategy.ts`, `KbqScrollbar.scrollIntoView`, CDK's `ScrollDispatcher` / `ViewportRuler`.

Worth considering: a `scrollAnchorIntoView(element): Promise<void>` helper in `components/core`, plus a scroll-strategy input on `KbqTooltipTrigger` instead of the DI override.

packages/components/inline-edit/inline-edit.spec.ts
```diff
@@ -839,20 +839,128 @@ describe('KbqInlineEdit', () => {
...
+            const hideSpy = jest.spyOn(tooltipTrigger, 'hide');
+
+            componentInstance.control.setValue('Some text');
+            componentInstance.control.updateValueAndValidity();
+            fixture.detectChanges();
+
+            expect(hideSpy).toHaveBeenCalled();
+        }));
+
+        it('should show the tooltip immediately when the control is already fully visible', () => {
```
lskramarov, line 869: This test never sets up an on-screen rect — it passes only because jsdom's un-mocked `getBoundingClientRect()` returns all zeros, which happens to satisfy `isFullyVisible`.

The sibling `setupOffscreen` helper mocks the rect explicitly, so the asymmetry is easy to miss. If the visibility predicate is later tightened (e.g. to require non-zero dimensions, or to intersect with ancestor scroll containers), this test would keep passing while silently exercising the opposite branch.

Suggested fix: mock an explicitly on-screen `DOMRect` the way `setupOffscreen` mocks the off-screen one.

packages/components/inline-edit/inline-edit.spec.ts
```diff
@@ -839,20 +839,128 @@ describe('KbqInlineEdit', () => {
...
+            tick(tooltipTrigger.enterDelay + 10);
+            expect(tooltipTrigger.isOpen).toBe(true);
+
+            const hideSpy = jest.spyOn(tooltipTrigger, 'hide');
+
+            componentInstance.control.setValue('Some text');
+            componentInstance.control.updateValueAndValidity();
+            fixture.detectChanges();
+
+            expect(hideSpy).toHaveBeenCalled();
```
lskramarov, line 865: The test is named "should hide the tooltip" but only asserts that `hide()` was *called*.

`KbqPopUpTrigger.hide()` early-returns on `preventClose` and is a no-op when `instance` is null, so if either guard starts swallowing the call the tooltip stays visible and this test still passes.

The machinery for a real assertion is already here — `tooltipTrigger.isOpen` is checked a few lines above. Suggested fix: `tick(...)` and `expect(tooltipTrigger.isOpen).toBe(false)`.

lskramarov (commented):
## Code review — `fix(inline-edit): show tooltip when out of viewport correctly (#DS-5420)`

14 comments below: 4 correctness bugs I reproduced against this branch, 3 more that are mechanically real but need a browser to confirm, and 7 cleanup / test-quality notes.

**Reproduced on this branch** (probe specs, run and discarded):

1. `showValidationTooltip` never re-checks validity, so the deferred tooltip still pops up with a stale error if the user fixes the value while the scroll is running — the same regression the PR description says it fixes. `show()` calls after the control became valid: **1**, expected 0.
2. `scrollend` is listened for on `window` in the bubble phase, but it does not bubble from element scroll containers — an inline edit inside a scrollable table / modal / `kbq-scrollbar` always waits the full 800 ms fallback. Element-targeted `scrollend` -> `show()` calls: **0**.
3. The `KBQ_TOOLTIP_SCROLL_STRATEGY` provider is on the component, so it leaks to **every tooltip in projected content**, not "just this tooltip instance" as the comment claims. Projected tooltip: `RepositionScrollStrategy`; identical tooltip outside: `CloseScrollStrategy`.
4. `isFullyVisible` compares against the viewport only, and `getBoundingClientRect()` ignores clipping by scrollable ancestors — a control hidden inside a scrolled container is treated as visible and no scroll happens.

Nice catch on `merge(...formFieldRefList.map(...))` — the old `merge(array)` form emitted the observables as values and completed immediately, so the hide-on-valid subscription had never worked at all. The existing 46-test suite passes on this branch.

_Generated with Claude Code_

packages/components/inline-edit/inline-edit.spec.ts
```diff
@@ -839,20 +839,128 @@ describe('KbqInlineEdit', () => {
...
+                    behavior: 'smooth'
+                });
+                expect(showSpy).not.toHaveBeenCalled();
+
+                window.dispatchEvent(new Event('scrollend'));
+
+                expect(updatePositionSpy).toHaveBeenCalled();
+                expect(showSpy).toHaveBeenCalled();
+
+                // Flush the now-irrelevant fallback timer (a no-op, guarded by the `shown` flag).
```
lskramarov: This comment is inaccurate. `showTooltip` already ran `cleanup()`, which does `clearTimeout(timeoutId)` — there is no fallback timer left to flush. What `tick(800)` actually flushes is the tooltip's 400 ms `kbqEnterDelay` timer scheduled by `KbqPopUp.show()`.

As written it tells the reader the fallback timer survives cleanup, which is the opposite of the code, and invites someone to delete the `tick` as unnecessary.

packages/components/inline-edit/inline-edit.spec.ts
```diff
@@ -839,20 +839,128 @@ describe('KbqInlineEdit', () => {
...
+        it('should use a reposition scroll strategy instead of the tooltip default close strategy', () => {
+            // The validation tooltip is anchored inside the edit-mode overlay (an overlay inside another
+            // overlay) and is shown right after programmatically scrolling its origin into view — closing it
+            // on that very scroll (the tooltip's default behavior) would immediately undo `show()`.
+            const fixture = setup(TestWithDynamicValidationTooltip);
+            const inlineEditDebugElement = openEditAndInvalidate(fixture);
+            const tooltipTrigger = (inlineEditDebugElement.componentInstance as any).tooltipTrigger();
+
+            expect(tooltipTrigger.scrollStrategy()).toBeInstanceOf(RepositionScrollStrategy);
+            expect(tooltipTrigger.scrollStrategy()).not.toBeInstanceOf(CloseScrollStrategy);
```
lskramarov: This assertion cannot add information: `RepositionScrollStrategy` and `CloseScrollStrategy` are unrelated CDK classes, so the `toBeInstanceOf(RepositionScrollStrategy)` on the line above already excludes it. It also forces the extra `CloseScrollStrategy` import at the top of the file.

Suggested fix: drop the assertion and the import.

tools/jest/setup.ts
```diff
@@ -65,3 +65,7 @@ global.CSS = {
 if (!globalThis.structuredClone) {
     globalThis.structuredClone = <T>(value: T): T => JSON.parse(JSON.stringify(value));
 }
+
+if (typeof Element !== 'undefined' && !Element.prototype.scrollIntoView) {
```
lskramarov: The `typeof Element !== 'undefined'` half of the guard is dead — everything above it in this file (`global.URL.createObjectURL`, `global.CSS`, `class extends MouseEvent`) already assumes a DOM environment and would have thrown first. The surrounding polyfills guard on the *feature* (`if (!globalThis.structuredClone)`), so `if (!Element.prototype.scrollIntoView)` would be consistent.

Minor, separate note: this installs one shared `jest.fn()` on the prototype for every element in every suite, so a future test asserting on `element.scrollIntoView` calls would be reading a repo-global call log. A plain no-op function would make that misuse impossible.

packages/components/inline-edit/inline-edit.ts
NikGurev: Спасибо, поймали баг! Добавили повторную проверку: если поле успело стать правильным, подсказка больше не покажется.

packages/components/inline-edit/inline-edit.ts
NikGurev: Поправили: теперь подсказка появится, даже если прокручивается не вся страница, а блок внутри неё.

packages/components/inline-edit/inline-edit.ts
NikGurev: Да, вы правы, комментарий был неточным. Убрали общую настройку и сделали её только для этой одной подсказки — на другие подсказки внутри она больше не влияет.

packages/components/inline-edit/inline-edit.ts
NikGurev: Добавили проверку и на вложенные прокручиваемые блоки, а не только на экран целиком.

packages/components/inline-edit/inline-edit.ts
NikGurev: Исправили — теперь показ подсказки происходит правильно, и страница обновляется как надо.

packages/components/inline-edit/inline-edit.ts
NikGurev: Теперь если прокрутка закончилась чуть позже, подсказка просто сдвинется на нужное место, а не останется в старом.

packages/components/inline-edit/inline-edit.ts
NikGurev: Добавили защиту: пока одна попытка показать подсказку не завершилась, повторные нажатия её не дублируют.

packages/components/inline-edit/inline-edit.ts
NikGurev: Убрали лишний вызов, он ничего не делал.

packages/components/inline-edit/inline-edit.ts
NikGurev: Согласны, хорошая идея на будущее. Пока оставили код здесь, но пометили в коде, что стоит вынести это отдельно позже.

packages/components/inline-edit/inline-edit.spec.ts
NikGurev: Поправили тест, теперь он честно проверяет видимый элемент, а не проходит случайно.

packages/components/inline-edit/inline-edit.spec.ts
NikGurev: Добавили проверку, что подсказка правда закрылась, а не только что была такая попытка.

packages/components/inline-edit/inline-edit.spec.ts
NikGurev: Поправили комментарий, он был неточным.

packages/components/inline-edit/inline-edit.spec.ts
NikGurev: Убрали лишнюю проверку и импорт, спасибо!

tools/jest/setup.ts
NikGurev: Вернули как было, эта проверка была не нужна.

lskramarov (approved):

rmnturov (approved):

PR #1933, opened 2026-08-25
feat: add smart search (#DS-3732)

NikGurev:
## Summary

Adds a small, opt-in "smart search" helper that matches multi-word, case- and accent-insensitive queries
E.g. typing `10.125 all` now finds `10.125.123.0/24 - all`. 
Updates default search in select, filter-bar, timezone.
Added as a separate example to tree-select.

## List of notable changes:

- **added** `createSearchPredicate`, `tokenizeSearchQuery`, `normalizeSearchValue` to `@koobiq/components/core` for smart search matching
- **added** a "Smart search" guide with the reasoning behind it
- **updated** search examples for `select`, `tree-select`, `filter-bar`, and `timezone`
-  **added** examples to autocomplete, tag-autocomplete
- **updated** the highlight pipe to support highlighting several words at once, so multi-word matches show up correctly
- **updated** `timezone`'s `highlightText` input and city filtering to accept multiple search terms

- The new public API in `@koobiq/components/core` — does the shape make sense?
- Whether the guide is clear and the examples are easy to follow


rmnturov:
Возможно, не надо в каждом компоненте делать раздел на основной странице с умным поиском — копипаста какая-то. Надо сделать в селекте обязательно, в остальных компонентах или убрать совсем, или скрыть в раздел «Примеры», или ссылаться на пример в Select (если однотипен), ограничиться текстом, ограничиться ссылокой на гайд по умному поиску

NikGurev:
> Возможно, не надо в каждом компоненте делать раздел на основной странице с умным поиском — копипаста какая-то. Надо сделать в селекте обязательно, в остальных компонентах или убрать совсем, или скрыть в раздел «Примеры», или ссылаться на пример в Select (если однотипен), ограничиться текстом, ограничиться ссылокой на гайд по умному поиску

Давай добавлю в примеры без описания. Суть в том что детали реализации разные, хотя внешне выглядит одинаково

lskramarov:
Гайд из доки https://koobiq-next--prs-1933-0o9bpfap.web.app/ru/main/search-smart не открывается:
image

NikGurev:
> Гайд из доки https://koobiq-next--prs-1933-0o9bpfap.web.app/ru/main/search-smart не открывается:

https://koobiq-next--prs-1933-0o9bpfap.web.app/ru/other/search-smart/overview

В другой секции лежит
```diff
- main/search-smart
+ other/search-smart
```

NikGurev:
Добавил ссылки в examples. Поправил то о чем говорил @lskramarov  - заменил main -> other в markdown

packages/components/timezone/timezone.utils.ts
```diff
@@ -95,17 +96,23 @@ export function timezonesSortComparator(first: KbqTimezoneZone, second: KbqTimez
...
+        .filter((pattern): pattern is string => typeof pattern === 'string' && pattern.length > 0)
+        .filter((pattern) => !UTC_OFFSET_PATTERN.test(pattern));
 
-    if (!searchPattern || onlyUTC) {
+    if (!patterns.length) {
         return cities;
     }
 
-    const regex: RegExp = RegExp(`(${searchPattern})`, 'gi');
+    const regex: RegExp = RegExp(`(${patterns.map(escapeRegExp).join('|')})`, 'gi');
```
lskramarov: **Shared `/g` regex keeps `lastIndex` across `test()` calls** (the `.filter(...)` on line 119)

`regex` is built with the `g` flag and then reused across every `.test()` call inside `.filter()`. `RegExp.prototype.test` on a global regex is stateful: after a successful match it leaves `lastIndex` past the match, and the next `test()` resumes from there instead of index 0. Matching cities are silently dropped.

Running the implementation as written:

```js
filterCitiesBySearchString('Berlin, Bern', 'ber');
// => "Berlin"   ("Bern" dropped: lastIndex is 3, so the scan of " Bern" starts past the "Ber")

filterCitiesBySearchString('Amsterdam, Andorra, Belgrade, Berlin, Bratislava, Brussels', ['ber', 'bra']);
// => " Berlin"  (" Bratislava" dropped)
```

The bug predates this PR, but the change makes it materially worse: with a single pattern only one city could match per pass, whereas the new `|` alternation means many cities match and every match poisons the scan of the next one. This is exactly the multi-token path the PR is built for.

The new spec passes only by coincidence — in `'Kaliningrad, Kazan, Kirov, Moscow'` with `['kazan', 'moscow']` the city skipped by the stale `lastIndex` (` Kirov`) happens to be a true negative.

Fix: drop the `g` flag (it buys nothing for `test`), or reset `regex.lastIndex = 0` before each call.

packages/docs-examples/components/timezone/timezone-search-smart/timezone-search-smart-example.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    protected get searchTokens(): string[] {
+        return tokenizeSearchQuery(this.searchControl.value ?? '');
+    }
+
+    private getFilteredData(): KbqTimezoneGroup[] {
+        const predicate = createSearchPredicate(this.searchControl.value ?? '');
+
+        const options = this.data.filter((timezone: KbqTimezoneZone) =>
+            predicate([offsetFormatter(timezone.offset), timezone.city, timezone.cities])
```
lskramarov: **Timezone offsets cannot be searched: U+2212 vs ASCII hyphen**

The predicate matches raw substrings against `offsetFormatter(timezone.offset)`, and `offsetFormatter` renders a negative sign with `String.fromCharCode(0x2212)` — MINUS SIGN, not a hyphen. `normalizeSearchValue` folds case and combining marks, but U+2212 has no compatibility decomposition, so it survives NFKD untouched:

```js
normalizeSearchValue('UTC−03:00').includes('-03'); // false
```

So in a field whose placeholder is «Город или часовой пояс», typing `-03` on a normal keyboard returns «Ничего не найдено» even though UTC−03:00 zones exist. The half of the placeholder that promises timezone search does not work.

The example this one sits next to, `timezone-search-overview-example`, handles it explicitly:

```ts
return escapedString.replace(/[\-—−]/g, '(-|—|−)');
```

Suggest normalizing dashes in the field values before handing them to the predicate (e.g. map `−`/`—` to `-` in both the query and `offsetFormatter(...)`), so the smart example is not a functional step back from the plain one.

packages/components/core/search/search-base.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (typeof value !== 'string' || !value) return '';
+
+    return value.trim().normalize('NFKD').replace(COMBINING_MARKS, '').toLocaleLowerCase();
+}
+
+const SEARCH_QUERY_TOKEN = /"([^"]*)"|(\S+)/g;
+
+/**
+ * Splits a raw query into tokens, treating whitespace as a separator and `"quoted phrases"` as
+ * a single token (spaces inside the quotes are preserved). An unterminated quote degrades
```
lskramarov: **An unterminated quote does not degrade gracefully**

The JSDoc says an unterminated quote "degrades gracefully into ordinary whitespace-separated tokens", but the `"` stays inside the token — the new spec asserts exactly this:

```ts
expect(tokenizeSearchQuery('foo "bar baz')).toEqual(['foo', '"bar', 'baz']);
```

`createSearchPredicate` then requires the literal substring `"bar`, which no realistic haystack contains, so the AND across tokens can never be satisfied:

```js
createSearchPredicate('"gue')('Café Wi-Fi guest network'); // false
createSearchPredicate('gue"')('Café Wi-Fi guest network'); // false
```

In practice: the moment a user types the opening quote of a phrase the result list empties and stays empty until they type the closing quote. That is the opposite of graceful — a plain substring search would keep narrowing as they type.

Either strip an unmatched leading/trailing `"` from bare tokens, or reword the JSDoc to state that a half-open quote yields no results.

packages/docs-examples/components/select/select-search-smart/select-search-smart-example.ts
```diff
@@ -0,0 +1,81 @@
...
+    ];
+
+    readonly searchControl = new FormControl();
+    readonly filteredOptions = this.searchControl.valueChanges.pipe(
+        startWith(''),
+        map((query) => this.search(query))
+    );
+
+    protected get searchTokens(): string[] {
+        return tokenizeSearchQuery(this.searchControl.value ?? '');
```
lskramarov: **Accent-folded matches render with no highlight**

`searchTokens` hands the raw tokens to `kbqHighlightBackground`, which matches them literally (case-insensitively). `createSearchPredicate`, on the other side, matches after NFKD diacritic folding. The two disagree exactly on the case the guide leads with — and this example's own option list demonstrates it:

```js
createSearchPredicate('cafe')('Café Wi-Fi guest network');    // true  → row is shown
highlight('Café Wi-Fi guest network', ['cafe'], mark);        // unchanged → nothing marked
```

So typing `cafe` surfaces **Café Wi-Fi guest network** with no mark anywhere on it, and the user gets no signal about why it matched. `wi-fi cafe` is worse: `Wi-Fi` lights up and `Café` does not, which reads like a rendering bug.

This affects all four new examples. Making it right needs a fold-aware highlight (match on the normalized string, then map offsets back onto the original) — or, at minimum, a note in the guide that highlighting is literal while matching is folded.

packages/components/core/search/search-base.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Normalizes a string for case- and diacritic-insensitive search comparison: trims the ends,
+ * decomposes accented characters (NFKD), strips the resulting combining marks, and lowercases
+ * using the current locale. Internal whitespace and punctuation are left untouched.
+ */
+export function normalizeSearchValue(value: string): string {
+    if (typeof value !== 'string' || !value) return '';
+
+    return value.trim().normalize('NFKD').replace(COMBINING_MARKS, '').toLocaleLowerCase();
```
lskramarov: **NFKD folding collapses Cyrillic `й` into `и`**

`U+0300–U+036F` is the generic combining-diacritic block, not a Latin-only one. NFKD decomposes `й` into `и` + U+0306 (combining breve) and `ё` into `е` + U+0308, and both marks fall inside the stripped range:

```js
normalizeSearchValue('мой');        // => 'мои'
normalizeSearchValue('Йошкар-Ола'); // => 'иошкар-ола'
```

`ё` → `е` is almost certainly wanted. `й` → `и` is not: they are distinct letters, so a query `мои` matches rows containing `мой` and vice versa, and `Йошкар` is found by typing `Иошкар`. For a library whose docs are Russian-first this will read as a matching bug rather than a feature.

Suggest restoring the breve pair after the strip (re-compose `и`+U+0306 → `й`, `И`+U+0306 → `Й`), or limiting the fold to Latin base letters.

packages/docs-examples/components/timezone/timezone-search-smart/timezone-search-smart-example.ts
```diff
@@ -0,0 +1,110 @@
...
+            of(getZonesGroupedByCountry(this.data, 'Другие страны')),
+            this.searchControl.valueChanges.pipe(
+                distinctUntilChanged(),
+                debounceTime(500),
+                map(() => this.getFilteredData())
+            )
+        );
+    }
+
+    protected get searchTokens(): string[] {
```
lskramarov: **`searchTokens` getter defeats pure-pipe memoization**

`tokenizeSearchQuery` allocates a fresh array on every read, and this getter is read from the template. Pure pipes compare their arguments by reference, so `kbqHighlightBackground` and `citiesByFilter` see a "changed" argument on every change-detection pass and re-run for every rendered option.

In this example the token array is bound four times per `kbq-timezone-option` (two offset spans, city, cities) across the whole timezone dataset. Each CD tick therefore re-tokenizes the query and rebuilds a `RegExp` plus a `split` per binding. The pattern being replaced (`searchPattern` / `treeControl.filterValue.value` / `searchControl.value`) was a stable reference the pipes could skip.

The same getter is copy-pasted into the select, tree-select and dropdown examples, so this is the shape consumers will lift into their own code. Caching it — a `signal`/`computed`, or recomputing on `valueChanges` alongside the filtered list — restores memoization and keeps the example honest as a reference.

docs/guides/search-smart.en.md
```diff
@@ -0,0 +1,68 @@
...
+Typo tolerance isn't part of these principles — for example, edit-distance matching where `developr` finds `developer`. It's a separate and much larger problem. If a product genuinely needs it, reach for a dedicated fuzzy-search library rather than extending this matching logic.
+
+### Examples by component
+
+Live examples and code live on each component's own page:
+
+- [Select](/en/components/select/overview#list-with-search)
+- [Tree-select](/en/components/tree-select/overview#with-search)
+- [Filter-bar](/en/components/filter-bar/overview#search)
+- [Timezone](/en/components/timezone/overview#search)
```
lskramarov: **Three of the four links land on the Overview tab, but those examples live on the Examples tab**

`tree-select-search-smart`, `filter-bar-smart-search` and `timezone-search-smart` were embedded in `examples.tree-select.*.md`, `examples.filter-bar.*.md` and `examples.timezone.*.md`, which render on the **Examples** tab. The links here point at `/overview#…` anchors (and the ru guide at the matching `#поиск` anchors), which resolve to the pre-existing plain-substring search sections. A reader following them sees the old example and never reaches the smart-search one.

Only the Select link is correct, because that example was added to `select.en.md` under `##### List with search`.

Either point the other three at `/examples`, or move those examples into the overview markdown next to the search sections they extend.

lskramarov (commented):
## Code review — smart search (#DS-3732)

Reviewed the full diff (37 files) against `main`. CI is green, so everything below is behavioral rather than a build break. The matching, highlighting and city-filtering logic was extracted and executed under Node to confirm each scenario.

**7 findings**, posted inline:

| # | Severity | Where | What |
| - | -------- | ----- | ---- |
| 1 | high | `timezone.utils.ts` | Shared `/g` regex keeps `lastIndex` across `test()` calls, dropping cities |
| 2 | medium | `timezone-search-smart-example.ts` | UTC offsets unsearchable: `offsetFormatter` emits U+2212, the query carries an ASCII hyphen |
| 3 | medium | `search-base.ts` | An unterminated quote does not degrade gracefully — it yields a token that can never match |
| 4 | medium | `select-search-smart-example.ts` | Accent-folded matches render with no highlight |
| 5 | medium | `search-base.ts` | NFKD folding collapses Cyrillic `й` into `и` |
| 6 | low | `timezone-search-smart-example.ts` | `searchTokens` getter defeats pure-pipe memoization (repeated in all four examples) |
| 7 | low | `search-smart.en.md` | Three of four guide links point at Overview, but those examples live on the Examples tab |

Nothing here blocks the shape of the public API — `createSearchPredicate` / `tokenizeSearchQuery` / `normalizeSearchValue` read well and the opt-in framing is right. Findings 1 and 2 are the ones worth fixing before merge; 3–5 are contract/JSDoc mismatches that will surface as puzzling behavior in the field.

One non-issue worth noting: the regenerated `example-module.ts` also picks up `file-upload-multiple-add-strategy`, which already exists on `main` but was missing from the generated map. That is a correct catch-up, not a defect.

_Generated with Claude Code_

artembelik (approved):

artembelik:
@rmnturov для примеров autocomplete, tag-autocomplete, dropdown нужно добавлять?

rmnturov:
> @rmnturov для примеров autocomplete, tag-autocomplete, dropdown нужно добавлять?

У этих компонентах умный поиск тоже должен работать. Не откажусь от примеров в подразделе «Примеры»

NikGurev:
Поправил всё, что отметили в ревью: убрал баг с поиском городов (раньше при поиске по нескольким словам часть городов пропадала), обновил тексты в гайде и поправил в нём ссылки на примеры, сделал так, чтобы буквы «й» и «и» не путались при поиске, починил поиск часовых поясов со знаком минус, а слова с ударениями теперь подсвечиваются в найденных результатах, а не просто находятся молча.

packages/components/timezone/timezone.utils.ts
NikGurev: Убрал g-флаг у регулярки — из-за него поиск мог случайно пропускать города. Добавил тест на этот случай.

packages/docs-examples/components/timezone/timezone-search-smart/timezone-search-smart-example.ts
NikGurev: Теперь перед поиском минус заменяется на обычный дефис, поэтому «-03» находит нужные часовые пояса.

packages/components/core/search/search-base.ts
NikGurev: Поправил описание — теперь честно написано, что незакрытая кавычка ничего не найдёт, пока её не закрыть.

packages/docs-examples/components/select/select-search-smart/select-search-smart-example.ts
NikGurev: Сделал подсветку умнее: теперь она тоже понимает буквы с ударением, поэтому «cafe» подсветит «Café».

packages/components/core/search/search-base.ts
NikGurev: Буквы «й» и «и» теперь не путаются при поиске, а «ё» по-прежнему приравнивается к «е».

packages/docs-examples/components/timezone/timezone-search-smart/timezone-search-smart-example.ts
NikGurev: Список токенов теперь пересчитывается только при смене запроса, а не при каждой отрисовке.

docs/guides/search-smart.en.md
NikGurev: Поправил ссылки — теперь они ведут на вкладку с примерами, где эти примеры и находятся.

lskramarov (approved):

rmnturov:
Я как понял, что на созвоне мы дейлике решили применять умный поиск в собственных компонентах по умолчанию. Если я это понял верно, то нужно изменить формулировку

image

rmnturov:
Это же относится и к разделам про поиск в компонентах. Если я правильно понял, то надо сразу показывать умный поиск. Глупый поиск не показывать или убрать его в «Примеры» — типа если вам не нужен умный поиск, то используйте… Хотя я готов остаться вообще без упоминания глупого поиска.
image

rmnturov:
Если я правильно понял дефолтный статус умного поиска, то его подключение и применение должно быть легким с технической точки зрения. Есть вариант это упростить или и так уже все как надо? Умный поиск ранее был опциональным, вдруг это повлияло на то, что он подключался чуть сложнее, чем если бы он был дефолтным поведением.

NikGurev:
> Если я правильно понял дефолтный статус умного поиска, то его подключение и применение должно быть легким с технической точки зрения. Есть вариант это упростить или и так уже все как надо? Умный поиск ранее был опциональным, вдруг это повлияло на то, что он подключался чуть сложнее, чем если бы он был дефолтным поведением.

@rmnturov у нас практически во всех компонентах поиск настраивается извне. 
А вот в tree-select изнутри, но там надо оценивать. Для дерева с большим количеством данных производительность может упасть, поэтому лучше включать такой поиск осознанно , а не по дефолту
. Для tree-select создал отдельную задачу

NikGurev:
> Это же относится и к разделам про поиск в компонентах. Если я правильно понял, то надо сразу показывать умный поиск. Глупый поиск не показывать или убрать его в «Примеры» — типа если вам не нужен умный поиск, то используйте… Хотя я готов остаться вообще без упоминания глупого поиска.
> 
> image
> 
> 

Да, поправлю

rmnturov:
Изменения внесены в [коммите faf5b71](https://github.com/koobiq/angular-components/commit/faf5b7140896a9dec509fe709e929dba125671eb).

Прогнал русскоязычный текст по редполитике, а английские страницы синхронизировал с ним.

- добавил краткое описание алгоритма умного поиска на страницах компонентов;
- добавил описание перед примером Dropdown;
- оставил рекомендацию и ссылку на руководство в одном абзаце на странице Select;
- заменил кавычки на жирное выделение примеров запросов в лейбле Autocomplete;
- убрал пустую строку из примеров ранжирования в обеих версиях руководства.

rmnturov (approved):

PR #1934, opened 2026-08-25
feat: added docs section in changelog (#DS-3365)

NikGurev:


lskramarov:
@NikGurev а тесты ? как проверить/увидеть что он теперь генерит ?

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+                            commit.scope = null;
+                            commit.package = matchingPackage;
+                        }
+                    }
+
+                    const type = getTypeOfCommitGroupDescription(group.title || '');
+
+                    // A commit is doc-related if it's grouped under the Documentation type, or its
+                    // scope is exactly "docs" (a multi-value scope like "docs,select" doesn't count).
+                    const isDocsCommit = type === 'docs' || commit.scope === 'docs';
```
lskramarov, line 151: **`type === 'docs'` is unreachable, so ordinary `docs:` commits still never reach this section.**

`conventional-changelog-angular@9.4.0` discards them before `finalizeContext` ever runs. Its `writer.js` sets `let discard = true`, clears it only inside `commit.notes.map(...)`, and then hits `else if (discard) return undefined` **before** reaching `else if (commit.type === 'docs')`. `transformCommit` turns that into `null` (`conventional-changelog-writer/dist/commit.js:33`) and the writer skips the commit (`writers.js:29`), so a commit group titled `Documentation` is never built and `getTypeOfCommitGroupDescription` can never return `'docs'`.

Verified by running the installed preset directly:

```
docs: x          (no notes)      => DISCARDED
docs(select): x  (no notes)      => DISCARDED
docs: x          (BREAKING note) => group="Documentation"
```

So the branch fires only for a `docs:` commit carrying a `BREAKING CHANGE` / `DEPRECATED` footer — of which this repo has none. Counting lines that start with ` * docs ` in `CHANGELOG.md` gives 0 across the whole file.

Net effect: the 6 `docs:` / `docs(scope):` commits of the last 1500 (e.g. `docs(select): document custom tag content via kbqSelectTagContent`) are still dropped entirely. Only the `commit.scope === 'docs'` half — `feat(docs)` / `fix(docs)`, 80 of the last 1500 — actually lands in the new section.

If routing real `docs:` commits is in scope for DS-3365, the preset's discard has to be handled in the `transform` wrapper at L108 rather than delegated to blindly. If it is out of scope, drop the `type === 'docs'` half together with the test that depends on it, so the code stops implying coverage it does not have.

packages/cli/src/release/changelog-writer-options.spec.ts
```diff
@@ -0,0 +1,311 @@
+import fs from 'fs';
+import { IReleaseTaskConfig } from './base-release-task';
+import { createChangelogWriterOptions } from './changelog-writer-options';
+
+/** Minimal stand-in for the conventional-changelog-angular preset's writer options. */
+const presetWriterOptions = {
+    headerPartial: (context: any) => `# ${context.title} (2026-01-01)`,
+    transform: (commit: any) => commit
```
lskramarov: **This stub does not model the preset's discard, and no test ever calls it.**

The real `conventional-changelog-angular` transform returns `undefined` for a `docs:` commit with no notes; this one returns the commit unchanged. More to the point, `finalizeChangelog` hands hand-built `commitGroups` straight to `finalizeContext`, so nothing in the suite passes through the transform stage at all — `headerPartial` is the only member of this stub that is ever invoked.

That is precisely why a `{ title: 'Documentation', commits: [...] }` group can exist in the fixture: the fixture is constructed one layer below where the discarding happens. The resulting snapshot line ` * docs  third commit abc1234` is output the tool has never emitted and, as things stand, cannot emit (see the comment on `changelog-writer-options.ts` L143).

So the test green-lights the half of the feature that does not work, and would keep passing if the `type === 'docs'` branch were deleted outright. Driving at least one case through `writerOptions.transform` with a stub that mirrors the preset's discard would turn this into a real regression guard.

packages/cli/src/release/changelog-writer-options.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+const config: IReleaseTaskConfig = {
+    projectDir: '/repo',
+    repoToken: '',
+    distDir: '',
+    tagName: '',
+    repoOwner: '',
+    repoName: '',
+    repoUrl: '',
+    changelogScope: 'koobiq',
+    withoutReferences: true,
```
lskramarov: **Both snapshots render with links disabled — the opposite of the production default.**

`cli.ts:36` declares `.option('-n, --without-references', 'exclude changelog links', false)`, so a normal release runs with `linkReferences = true` and emits `([097b0aa](https://github.com/koobiq/angular-components/commit/097b0aa…))` — exactly as every line of the committed `CHANGELOG.md` shows. The golden files instead freeze bare hashes: ` * feature  **docs:** test commit. abc1234`.

Related gap: the single link test (L195) supplies no `context.repository`, so it exercises `renderRepositoryUrl`'s `return context.repoUrl` fallback. The `${host}/${owner}/${repository}` branch — the only one reachable after `.readRepository()` in `changelog.ts:41` — has no coverage at all, so a regression in that join would ship green.

Flipping the shared `config` to `withoutReferences: false` and supplying `host` / `owner` / `repository` on the fake context would make the snapshots represent what the tool actually writes.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+                    const packageName = isDocsCommit ? DOCS_SECTION_KEY : commit.package || config.changelogScope;
```
lskramarov, line 152: **Docs entries keep a redundant `**docs:**` prefix.**

Ten lines up (L134) the package-remap path clears `commit.scope = null` precisely so the heading is not repeated on every line, and the spec locks that in with `expect(output).not.toContain('**cli:**')`. The docs path skips that step, so the new snapshot reads:

```
### Documentation

 * feature  **docs:** test commit. abc1234
```

Every real `feat(docs)` / `fix(docs)` entry — 80 in the last 1500 commits — will carry that prefix under a heading that already says Documentation.

```suggestion
                    const packageName = isDocsCommit ? DOCS_SECTION_KEY : commit.package || config.changelogScope;

                    // The "Documentation" heading already names the scope, so drop the redundant
                    // prefix, mirroring the package remap above.
                    if (commit.scope === 'docs') {
                        commit.scope = null;
                    }
```

lskramarov (commented):
## Automated review — 10 confirmed findings

Reviewed `feat/DS-3365` at 774f2a0 against `main`. The baseline is green: `npx jest --roots packages/cli` passes 30/30 (3 suites, 2 snapshots) and eslint + prettier are clean on every changed file — so everything below is something the current suite does not catch.

The headline finding is the first one: **ordinary `docs:` commits still never reach the new Documentation section.** The preset discards them before `finalizeContext` runs, and the new snapshot test does not notice because it builds its fixture below the layer where that discarding happens. The `commit.scope === 'docs'` half of the feature (`feat(docs)` / `fix(docs)`, 80 of the last 1500 commits) does work.

Two further observations were rated *plausible* rather than confirmed and are left out of this review: a namespace collision between `DOCS_SECTION_KEY` and package names, and overlapping coverage between the two Documentation tests. Happy to expand on either if useful.

_Generated with Claude Code_

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
+import fs from 'fs';
+import type { IReleaseTaskConfig } from './base-release-task';
+
+/**
+ * Maps a commit note to a string that will be used to match notes of the
+ * given type in commit messages.
+ */
+export const enum CommitNote {
```
lskramarov: **Nine symbols gained `export` with no consumer.**

`CommitNote`, `DOCS_SECTION_KEY`, `IChangelogPackage`, `compareCommitsBy`, `renderRepositoryUrl`, `renderCommit`, `preferredOrderComparator`, `getTypeOfCommitGroupDescription` and `preparePackageName` were all file-private in `changelog.ts` before this PR. Grepping every `.ts` under `packages/` for each name returns zero hits outside this file — the new spec imports only `createChangelogWriterOptions`.

`export const enum CommitNote` is the one worth changing regardless of the rest: a `const enum` crossing a module boundary is erased to `undefined` under any isolated-transpile toolchain (ts-jest `isolatedModules`, esbuild, swc), so the first file that does import it would silently compare against `undefined` instead of `'DEPRECATED'`. Dropping `export` on all nine restores the pre-PR encapsulation at no cost.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+        : commit.shortHash;
+
+    return `${scope} ${subject} ${link}`;
+}
+
+export function createChangelogWriterOptions(
+    changelogPath: string,
+    presetWriterOptions: any,
+    config: IReleaseTaskConfig,
+    logSkippedDuplicate: (commit: any) => void = () => {}
```
lskramarov, line 75: **The one genuinely new API in this extraction has no test.**

`finalizeChangelog` in the spec always calls the 3-argument form, so `logSkippedDuplicate(commit)` at L125 is only ever the `() => {}` default; the duplicate test asserts on rendered output alone. If the callback were dropped, handed the wrong argument, or moved after the early `return false`, the suite stays green while `koobiq-cli` silently stops printing `↺ Skipping duplicate: "..."` during a release — the only signal a maintainer gets that a cherry-picked commit was filtered out.

Threading a `jest.fn()` through `finalizeChangelog` and asserting `toHaveBeenCalledWith(expect.objectContaining({ header: 'fix(button): fix duplicate' }))` closes it.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+) {
+    const existingChangelogContent = fs.readFileSync(changelogPath, 'utf8');
```
lskramarov: **Reading the changelog in here is what forces the spec to monkey-patch `fs` globally.**

`changelog.ts` on `main` imported `{ createReadStream, createWriteStream, readFileSync } from 'fs'` and called `readFileSync(changelogPath, 'utf8')`. A named import is captured at module load, so `jest.spyOn(fs, 'readFileSync')` could not intercept it — which is why this file switched to `import fs from 'fs'` plus `fs.readFileSync`. That is production code reshaped to fit the test mechanics, and it costs twice over: a process-wide `fs.readFileSync` mock returning `''` for the duration of every test (any unrelated read jest or ts-jest performs mid-test gets `''`), and a sibling `changelog.ts` that still uses named `fs` imports, so the two files now disagree on style for the same module.

Threading the content in as a parameter — exactly as `logSkippedDuplicate` already is — removes both.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+                    const packageName = isDocsCommit ? DOCS_SECTION_KEY : commit.package || config.changelogScope;
+
+                    if (!packageGroups[packageName]) {
+                        packageGroups[packageName] = { commits: [], breakingChanges: [], deprecations: [] };
+                    }
+
+                    const packageGroup = packageGroups[packageName];
+
+                    // Collect all notes of the commit. Either breaking change or deprecation notes.
+                    commit.notes.forEach((n: any) => {
```
lskramarov: **These notes are collected and then thrown away.**

`template` (L86-97) emits only the header plus `group.title` and `group.commits`. `breakingChanges` / `deprecations` are written at L195-196 and read nowhere. History agrees: the only section headings that have ever appeared in `CHANGELOG.md` are `### Koobiq`, `### Components` and `### Cdk` — no BREAKING CHANGES or Deprecations heading — and the file contains zero occurrences of `DEPRECATED`.

The new spec now asserts on this dead data, pinning output nothing produces. Either render the sections (the `throw` at L159 suggests someone meant to) or drop the collection loop, the two `IChangelogPackage` fields and the test. This one predates the PR, but the whole file is new lines and the PR adds a test for it, so it seems worth settling now.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+
+        // Overwrites the conventional-changelog-angular preset transform function. This is necessary
+        // because the Angular preset changes every commit note to a breaking change note. Since we
+        // have a custom note type for deprecations, we need to keep track of the original type.
+        //
+        // conventional-changelog-writer 9 hands the transform an immutable commit and merges the
+        // returned diff, so the note type is carried on a copy rather than assigned in place. The
+        // preset spreads each incoming note before overwriting its title, so `type` survives, and
+        // it returns `notes` as part of its diff.
+        transform: (commit: any, context: any) => {
```
lskramarov, line 112: **No test ever invokes `transform`, so the mapping this block exists for is uncovered.**

The spec calls only `finalizeContext` and `template`. The notes test hand-builds `{ type: 'BREAKING CHANGE' }` / `{ type: 'DEPRECATED' }` — which is the *output* of L109 — so it cannot detect the mapping breaking. If the preset stopped preserving `type` through its note spread, exactly the fragility the comment at L104-107 documents, every note would fall through to `throw Error('Found commit note that is not known')` at L159 and abort the release, with a green suite. That throw has no test either.

packages/cli/src/release/changelog-writer-options.ts
```diff
@@ -0,0 +1,251 @@
...
+                .map(
+                    (group: any) =>
+                        `### ${group.title}\n\n` +
+                        group.commits
+                            .map((commit: any) => ` * ${commit.type} ${renderCommit(context, commit)}\n`)
+                            .join('') +
+                        '\n'
+                )
+                .join(''),
+        commitPartial: renderCommit,
```
lskramarov: **`commitPartial` is inert.**

`conventional-changelog-writer` 9 only copies it onto the template context (`dist/context.js:82`); rendering goes through `options.template(templateContext)` (`dist/template.js:12`), and the custom `template` at L86-97 inlines `renderCommit(context, commit)` itself. Nothing ever calls this. It reads as though it drives the commit line format, so someone editing it later to change that format would see no effect.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Поправил: docs-коммиты без заметок теперь тоже попадают в раздел Documentation.

packages/cli/src/release/changelog-writer-options.spec.ts
NikGurev: Переписал стаб и тесты, теперь они честно проверяют это поведение.

packages/cli/src/release/changelog-writer-options.spec.ts
NikGurev: Поправил конфиг и снэпшоты — по умолчанию теперь со ссылками, как в проде.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Убрал лишний префикс docs у таких коммитов.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Убрал export у всего лишнего, что не используется снаружи.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Добавил тест на этот колбэк.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Теперь содержимое файла передаётся параметром, без подмены fs в тестах.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Убрал этот неиспользуемый сбор данных и тест на него.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Добавил тесты, которые гоняют коммиты через transform.

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Убрал эту неиспользуемую опцию.

lskramarov (approved):

packages/cli/src/release/changelog-writer-options.ts
NikGurev: Уточнение: решили оставить этот код для будущего использования, только покрыли его тестами.

PR #1937, opened 2026-08-26
feat: prerender documentation examples in CI (#DS-5064)

artembelik:


artembelik:
https://github.com/koobiq/angular-components/pull/1950

packages/components-dev/ssr/routes.ts
```diff
@@ -0,0 +1,48 @@
+import { Type } from '@angular/core';
+import { Routes } from '@angular/router';
+import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
+
+const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
+    // TODO: NG0500: During hydration Angular expected a comment node but found <p>. (#DS-5467)
+    'checkbox-indeterminate',
+    // TODO: KbqThemeService throws "this.window.matchMedia is not a function" during SSR. (#DS-5064)
+    'empty-state-content',
+    'notification-center-empty',
```
NikGurev, line 10: notification center тоже стоит пометить TODO, должен по идее работать с SSR

NikGurev (approved):

packages/components-dev/ssr/routes.ts
artembelik: todo есть, он общий для всех примеров ниже:
>  // TODO: KbqThemeService throws "this.window.matchMedia is not a function" during SSR. (#DS-5064)

tools/example-module/generate-example-module.ts
```diff
@@ -58,9 +58,7 @@ function inlineExampleModuleTemplate(parsedData: AnalyzedExamples): string {
     const loadText = [
         `export async function loadExample(id: string): Promise<any> {`,
         `  switch (id) {`,
-        ...exampleMetadata.map(
-            (data) => `  case '${data.id}':\nreturn import('@koobiq/docs-examples/${data.importPath}');`
-        ),
+        ...exampleMetadata.map((data) => `  case '${data.id}':\nreturn import('./${data.importPath}');`),
```
lskramarov, line 61: **Relative dynamic imports break the `@koobiq/docs-examples` entry-point boundary.**

`./components/accordion` is no longer a package specifier, so ng-packagr's rollup pass resolves it *inside* the primary entry point instead of leaving it external (`external` matches bare specifiers only; `treeshake: false`, `inlineDynamicImports: false`). I reproduced it with rollup using ng-packagr's exact output options:

```
emitted: [ 'koobiq-docs-examples.mjs', 'koobiq-docs-examples-index-C1MbhUh5.mjs' ]

// koobiq-docs-examples.mjs
case 'a': return import('./koobiq-docs-examples-index-C1MbhUh5.mjs');
```

Consequences in `dist/docs-examples`:
- every example is emitted **twice** — once in its published secondary entry point (`fesm2022/koobiq-docs-examples-components-accordion.mjs`, which stays for external consumers) and once as a duplicate chunk that `loadExample` now points at;
- all 580 chunks resolve to `components/<x>/index.js`, so `[name]` is always `index` and the emitted names carry no information;
- `apps/docs` loads the duplicate, not the published entry point.

On `main` the built primary bundle keeps `import('@koobiq/docs-examples/components/accordion')` external — I checked `dist/docs-examples/fesm2022/koobiq-docs-examples.mjs`.

This changes the shape of a published package to satisfy the dev SSR app. Consider keeping the package specifier and adding a `"@koobiq/docs-examples/*": ["packages/docs-examples/*"]` path mapping in `packages/components-dev/ssr/tsconfig.app.json` instead.

_Generated with Claude Code_

packages/components-dev/ssr/module.ts
```diff
@@ -1,280 +1,98 @@
...
+    protected readonly selectedExample = model(this.getExampleFromUrl() ?? this.options[0]);
+    protected readonly searchQuery = signal<string | null>('');
+    protected readonly filteredOptions = computed(() => {
+        const query = this.searchQuery()?.trim().toLowerCase() ?? '';
+
+        return query ? this.options.filter((option) => option.toLowerCase().includes(query)) : this.options;
+    });
+
+    constructor() {
+        effect(() => {
```
lskramarov, line 72: **The picker desyncs from the URL, and re-selecting the option it displays is a no-op.**

Nothing subscribes to router events, so browser Back/Forward changes the URL without touching `selectedExample`. Concretely:

1. pick `button-overview` → URL `/button-overview`, picker shows `button-overview`;
2. press Back → URL is `/alert-status`, the outlet renders `alert-status`, but the picker still reads `button-overview`;
3. pick `button-overview` again → the model already holds that value, the signal doesn't change, the effect never runs, and nothing navigates.

You are now stuck on `alert-status` with the picker claiming `button-overview`, and the only way out is to select some third example first. Syncing the model from `router.events` (`NavigationEnd`) would fix both the stale label and the dead re-selection.

_Generated with Claude Code_

packages/components-dev/ssr/routes.ts
```diff
@@ -0,0 +1,48 @@
...
+    'components/accordion',
+    // AG Grid does not support server-side rendering.
+    'components/ag-grid',
+    // TODO: Restore after breaking the circular dependency in filter-bar. (#DS-5467)
+    'components/filter-bar',
+    // TODO: Restore with filter-bar; this barrel exports UsernameFilterBarOptionExample. (#DS-5467)
+    'components/username'
+]);
+
+export const devSsrRoutes: Routes = Object.entries(EXAMPLE_COMPONENTS)
```
lskramarov, line 34: **No `''` route and no `'**'` fallback — the URL the README tells developers to open renders nothing.**

`ssr:start` builds with the `production` configuration, so `isDevMode` is false and `template.html` skips `<dev-header />`; the redirect effect that covers `/` in dev never runs. And `/` is not prerendered (the CI job reports "Prerendered 516 static routes", exactly the number of example routes), so `express.static(..., { index: 'index.html' })` in `server.ts` serves the plain CSR shell. The client then boots, the router cannot match `''`, and you get NG04002 with an empty outlet — no content and no way to navigate, at the `http://localhost:4000` the README documents.

A `{ path: '', redirectTo: <first id>, pathMatch: 'full' }` plus a wildcard fallback would fix this. The wildcard would likely also cover the `popover-scrolling-and-layering` exclusion above, whose NG04002 on `examples/popover` comes from the same gap rather than from anything specific to that example.

_Generated with Claude Code_

packages/components-dev/ssr/template.html
```diff
@@ -1,10 +1,6 @@
...
+}
 
-<hr />
-
-<dev-examples />
-
-<hr />
-
-<dev-breadcrumbs-hydration />
+<router-outlet />
```
lskramarov, line 6: **Dropping `<dev-breadcrumbs-hydration />` orphans the file and removes the only hydration case.**

`packages/components-dev/ssr/components/breadcrumbs.ts` is still in the tree but is now referenced by nothing — it should go with this change, or be routed to.

More importantly, it existed to reproduce https://github.com/angular/angular/issues/50543 (delayed hydration cleanup leaving duplicated content) — note the deliberate `timer(2000)` holding `ApplicationRef.isStable`. Nothing replaces it, and the new `ssr` CI job only builds and prerenders, so it never hydrates. That is exactly the class of bug the `checkbox-indeterminate` TODO in `routes.ts` records ("NG0500: During hydration Angular expected a comment node but found `<p>`") — a browser-only error that this PR's job cannot catch.

_Generated with Claude Code_

packages/components-dev/ssr/routes.ts
```diff
@@ -0,0 +1,48 @@
+import { Type } from '@angular/core';
+import { Routes } from '@angular/router';
+import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
+
+const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
+    // TODO: NG0500: During hydration Angular expected a comment node but found <p>. (#DS-5467)
+    'checkbox-indeterminate',
+    // TODO: KbqThemeService throws "this.window.matchMedia is not a function" during SSR. (#DS-5064)
```
lskramarov, line 8: **These TODOs point at the ticket this PR closes, so they will never be picked up.**

Eight of the twelve excluded ids are here because `KbqThemeService` (`providedIn: 'root'`) calls `this.window.matchMedia('(prefers-color-scheme: dark)')` in a field initializer — `packages/components/core/services/theme.service.ts:274` — which throws under SSR. That is a real library bug, and it is the same reason `DevThemeToggle` had to be dropped from the dev app entirely.

Filing it as a genuine SSR bug (its own ticket, like the `#DS-5467` entries below) rather than tagging it `#DS-5064` would keep it visible. As written, closing DS-5064 closes the only pointer to the bug, and the exclusion outlives it silently.

_Generated with Claude Code_

packages/components-dev/ssr/config.ts
```diff
@@ -4,13 +4,14 @@ import { provideClientHydration, withEventReplay } from '@angular/platform-brows
...
 import { kbqLocaleServiceProvider } from '@koobiq/components/core';
+import { devSsrRoutes } from './routes';
 import { devTimezoneBrowserProviders } from './timezone';
 
 export default {
     providers: [
         kbqLocaleServiceProvider(),
         devTimezoneBrowserProviders(),
-        provideRouter([]),
+        provideRouter(devSsrRoutes),
```
lskramarov, line 14: **This pulls the whole example catalogue into the initial browser bundle of all 516 prerendered pages.**

`devSsrRoutes` reads `EXAMPLE_COMPONENTS` at module scope via `Object.entries`, so the entire object literal is retained. Measured on this branch's `example-module.ts`:

| | bytes |
|---|---|
| `EXAMPLE_COMPONENTS` as JSON | 228,293 |
| the three fields actually used (`id`, `componentName`, `importPath`) | 61,914 |

The rest — `title`, `files`, `localImportFiles`, `packagePath`, `selector`, `additionalComponents` — is never read here. The CI build reports `main-SNCF5KPY.js` at 322.73 kB raw, most of it this metadata.

Emitting a small `{ id, componentName, importPath }` list from the generator (or deriving the routes at build time) would cut the initial bundle by roughly two thirds.

_Generated with Claude Code_

packages/components-dev/ssr/module.ts
```diff
@@ -1,280 +1,98 @@
...
+    }
+}
 
 @Component({
     selector: 'dev-app',
-    imports: [DevDocsExamples, DevThemeToggle, DevLocaleSelector, DevBreadcrumbsHydration],
+    imports: [RouterOutlet, DevHeader],
     templateUrl: './template.html',
     styleUrl: './styles.scss',
+    providers: [KbqSidepanelService],
```
lskramarov, line 95: **Unexplained `KbqSidepanelService` provider.**

`KbqSidepanelModule` already lists `KbqSidepanelService` in its `providers` (`packages/components/sidepanel/sidepanel.module.ts:33`), and every sidepanel example imports that module, so the standalone-import mechanism should already supply it — as it did when these examples were rendered inline from `DevDocsExamples` on `main`.

Declaring it here creates a second instance in `DevApp`'s node injector that shadows the environment one, so the two would keep separate `openedSidepanels` stacks. If a lazily loaded example actually failed without this, the comment explaining why is the valuable part; if it was added while debugging, it should come out.

_Generated with Claude Code_

packages/components-dev/ssr/routes.ts
```diff
@@ -0,0 +1,48 @@
+import { Type } from '@angular/core';
+import { Routes } from '@angular/router';
+import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
+
+const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
```
lskramarov, line 5: **Nothing keeps these two lists honest.**

Both are matched with `Set.has` against generated ids and import paths. Rename an example, or fix the underlying SSR bug, and the stale entry keeps silently excluding — or stops excluding — with no signal either way. All sixteen entries do match `EXAMPLE_COMPONENTS` today; the concern is drift.

A cheap guard: assert at module load (or in a spec) that every entry in both sets still resolves to at least one example, and fail loudly when it does not.

_Generated with Claude Code_

packages/components-dev/ssr/module.ts
```diff
@@ -1,280 +1,98 @@
...
+            if (example && this.getExampleFromUrl() !== example) {
+                this.router.navigateByUrl(example);
+            }
+        });
+    }
+
+    // `Router.url` still points at the default `/` here because the initial navigation
+    // hasn't completed yet; `Location.path()` reflects the real (server/browser) URL immediately.
+    private getExampleFromUrl(): string | undefined {
+        const path = this.location.path().replace(/^\//, '').split(/[?#]/)[0];
```
lskramarov, line 84: **Hand-rolled URL parsing, and the `#` half is dead.**

`Location.path()` defaults to `includeHash = false`, so the fragment is never in the string and `split(/[?#]/)` only ever splits on `?`.

`this.router.parseUrl(this.location.path())` is pure — it does not depend on the navigation having completed, which is what the comment above is guarding against — and gives the primary segment directly, handling query params, matrix params and encoding for free.

_Generated with Claude Code_

packages/components-dev/ssr/module.ts
```diff
@@ -1,280 +1,98 @@
...
     changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class DevDocsExamples {}
+export class DevHeader {
+    private readonly router = inject(Router);
+    private readonly location = inject(Location);
+    protected readonly options = devSsrRoutes
+        .map(({ path }) => path)
+        .filter((path): path is string => typeof path === 'string' && path.length > 0);
+    protected readonly selectedExample = model(this.getExampleFromUrl() ?? this.options[0]);
```
lskramarov, line 63: `selectedExample` is local component state that nothing outside `dev-header` binds to, but `model()` publishes it as an input **and** an output on the component's public surface. AGENTS.md reserves `model()` for "two-way bound properties with `[(prop)]` syntax" — consuming a child's `[(value)]` does not need it. `signal()` covers this and keeps the surface closed.

Note also that `this.options[0]` types as `string` while it is `undefined` when every route is excluded, which is why the effect below needs its `if (example && …)` guard.

_Generated with Claude Code_

lskramarov (commented):
Automated review (`/code-review xhigh`) — 14 inline findings, most severe first.

packages/components-dev/ssr/module.ts
```diff
@@ -1,280 +1,98 @@
...
+
+                @for (option of filteredOptions(); track option) {
+                    <kbq-option [value]="option">
+                        <span [innerHTML]="option | mcHighlight: searchQuery()"></span>
+                    </kbq-option>
+                }
+            </kbq-select>
+        </kbq-form-field>
     `,
+    styles: '',
```
lskramarov, line 54: `styles: ''` is a no-op — drop it.

_Generated with Claude Code_

angular.json
```diff
@@ -2191,23 +2191,43 @@
...
                         "polyfills": ["zone.js"],
-                        "extractLicenses": false,
-                        "sourceMap": true,
-                        "optimization": false,
-                        "namedChunks": true,
                         "browser": "packages/components-dev/ssr/main.ts",
                         "server": "packages/components-dev/ssr/main.server.ts",
-                        "prerender": true,
+                        "prerender": {
+                            "discoverRoutes": true
```
lskramarov, line 2197: `discoverRoutes` defaults to `true`, so `{ "discoverRoutes": true }` is exactly the `"prerender": true` it replaces. (Contrast `koobiq-docs`, which sets it to `false` and therefore needs the object form.) Reverting this line keeps the diff to the parts that change behaviour.

_Generated with Claude Code_

.github/workflows/build.yml
```diff
@@ -20,3 +20,12 @@ jobs:
       # have to be resolved with npm to catch an ERESOLVE before it reaches consumers.
       - run: yarn run check-npm-resolution
       - uses: ./.github/workflows/actions/build-docs
+
+  ssr:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
lskramarov, line 26: The 60-minute bound on the `build` job above is documented — `check-npm-resolution` talks to the npm registry and can stall on an outage. Nothing in this job has that property; it finishes in about two minutes. Copying the number without the reason means a hung prerender holds a runner for an hour. Something like `timeout-minutes: 20` would still be generous.

_Generated with Claude Code_

packages/components-dev/index.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8" />
+        <base href="/" />
```
lskramarov, line 5: This `index.html` is shared by 79 dev projects in `angular.json`, so `<base href="/">` lands in all of them to serve one. It is harmless today — every dev app is served from the root — but it does change relative URL resolution app-wide. Worth a comment saying it is here for the SSR prerender, since that is not discoverable from this file.

_Generated with Claude Code_

tools/example-module/generate-example-module.ts
artembelik: откатил

packages/components-dev/ssr/routes.ts
artembelik: поправил

packages/components-dev/ssr/template.html
artembelik: вернул

packages/components-dev/ssr/routes.ts
artembelik: на это есть отдельные задачи

packages/components-dev/ssr/module.ts
artembelik: поправил

packages/components-dev/ssr/config.ts
artembelik: осознанное решение

packages/components-dev/ssr/module.ts
artembelik: добавил комментарий

packages/components-dev/ssr/module.ts
artembelik: поправил

angular.json
artembelik: откатил

.github/workflows/build.yml
artembelik: уменьшил до 20

packages/components-dev/index.html
artembelik: добавил коммент, на остальные дев приложения не повлияло

lskramarov (approved):

PR #1940, opened 2026-08-26
fix(search-expandable): errors following a full review of the component

lskramarov:
## Summary

`kbq-search-expandable` replaced the removed `KbqFilterBarSearch`, and the review found the swap left
two holes: the control's error state never reflected the consumer's own form control, and the
`v20-upgrade` migration silently dropped part of the old component's surface.

The second commit closes the migration gaps the first pass left open and stops a handled Enter from
also submitting an enclosing form.

## List of notable changes:

- **fixed** the error state: the inner `<input>` self-injects `NgControl`, so `errorState` was
  computed from the component's own private, validator-free control instead of the bound one — an
  invalid field stayed unstyled and rendered `aria-invalid="false"`
- **added** `search-expandable-tokens.scss`, following the hand-written `:where(.kbq-...)` convention
  the other token files use
- **fixed** the migration rule that dropped `KbqFilterBarSearch`'s `tooltip` alias, which would have
  silently replaced a custom tooltip with the locale default
- **fixed** an unscoped identifier rewrite in the migration that rewrote reads without rewriting the
  declaration, producing `TS2339`
- **fixed** `onEnter()` not calling `preventDefault()` — inside a native `<form>`, Enter both emitted
  the query and submitted the form; the default is now prevented only on the path that actually emits
- **added** two warn patterns to `v20-upgrade` with matching README rows: the attribute form
  `<div kbq-filter-search>` has no counterpart in an element-only selector and would silently match
  nothing after the upgrade, and the emit timeout default moved from `0` to `200`ms

- `packages/schematics/src/migrations/v20-upgrade/data.ts` — the new attribute-form warn uses a
  lookbehind so it does not fire on the element form the migration already rewrites, and there is a
  test pinning exactly that.
- The error-state fix changes which control drives `.kbq-form-field_invalid` and `aria-invalid`.

- The `any` / `unknown` ratchet moves this package from 1 to 2. The added occurrence is
  `InputSignalWithTransform<boolean, unknown>` on `isEmitValueByEnterEnabled` — Angular's own write type
  for `input(false, { transform: booleanAttribute })`, since `booleanAttribute` is declared
  `(value: unknown) => boolean`. It cannot be narrowed without giving up the coercion helper.

---

One of ten pull requests from a single review pass (`review/*`). Every branch carries two commits: the
fixes from the component review, and a second pass over the findings an independent verifier could not
confirm outright.

The branches share files, so they have to be merged one at a time with a rebase in between. Overlap:
`packages/components/core/locales/*` (7 branches), `tools/public_api_guard/components/core.api.md` (7),
`tools/cspell-locales/*.json` (6), `packages/e2e/routes.ts` (4) and
`packages/schematics/src/{collection,migrations}.json` (3).


artembelik (approved):

packages/components/search-expandable/search-expandable.ru.md
```diff
@@ -1,5 +1,7 @@
 Поле поиска, которое разворачивается из компактной кнопки-иконки.
 
+Компонент является элементом формы и не имеет отдельного входного параметра для значения: привяжите его через `[formControl]`, `[formControlName]` или `[(ngModel)]`, иначе при инициализации будет выброшена ошибка. Введённое значение попадает в связанный контрол через `emitValueTimeout` миллисекунд или по нажатию `Enter`, если задан `isEmitValueByEnterEnabled`.
```
NikGurev, line 3: Разнести бы абзац на разные примеры: 
- Привязку к formControl можно добавить к  абзацу ниже
- Добавить отдельный пример про `isEmitValueByEnterEnabled`

NikGurev (approved):

packages/components/search-expandable/search-expandable.ru.md
lskramarov: кажется, что без доп. примеров все понятно.. предлагаю вернуться к этому как только прилетит первый запрос на недостаточность документации

PR #1941, opened 2026-08-26
fix(tooltip): errors following a full review of the component

lskramarov:
## Summary

Tooltip works for pointer users and was structurally silent for everyone else: no `role="tooltip"`, no
`aria-describedby`, no way to dismiss a hover tooltip with Escape, and a default that made every
tooltip transparent to the pointer — a WCAG 1.4.13 *hoverable* failure in the shipped configuration.
The review also fixed the pure-standalone `NullInjectorError` and several `core/pop-up` defects the
whole pop-up family inherits.

The second commit extends the pointer-events compensation to the built-in overflow hints and closes a
latch that could leave an invisible pane attached.

## List of notable changes:

- **fixed** the missing ARIA: the pane now carries `role="tooltip"` and the trigger points
  `aria-describedby` at it, so a screen reader announces the hint together with the control
- **fixed** the `ignoreTooltipPointerEvents` default, which was `true` — the pane could not be reached
  by the pointer at all, so a hint that takes a while to read was unreachable (WCAG 1.4.13)
- **fixed** Escape not dismissing a hover tooltip on a non-focusable element, where the only handler
  sat on the trigger's own `keydown`
- **fixed** the standalone import crashing at directive construction because
  `KBQ_TOOLTIP_SCROLL_STRATEGY` had no factory default
- **fixed** `core/pop-up` defects the family shares: the `hideWithTimeout` watchdog degenerating into
  an `interval(0)`, a pending hide that re-hover could not cancel, and a timer that could fire after
  destroy
- **added** `core/pop-up/pop-up.spec.ts` — the shared base had no spec of its own
- **fixed** the remaining pointer-capturing hints: `kbq-title` and `kbqEllipsisCenter` now opt out the
  way `kbq-option` and `kbq-timezone-option` already did, and a tooltip whose trigger is `manual` or
  `none` opts out unconditionally — hovering such a pane neither opens nor closes it, so a
  pointer-capturing pane there only blocks what it floats over, including the validation hints of
  datepicker, timepicker and inline-edit for their full three seconds
- **fixed** `showForMouseEvent` bypassing the empty-content guard, which latched an invisible pane
  open and left it swallowing document keydowns until the next `hide()`
- **added** the `tooltip-pointer-events-and-types` migration. It reports the *absence* of
  `ignoreTooltipPointerEvents` in a file that renders tooltips, because the markup whose behaviour
  changed is exactly the markup that never mentioned the input — plus the removed `scheduler`, the
  dropped `getMouseLeaveListener` argument, the narrowed `placementChange` payload and the `any` that
  left `content` / `header` / `context` / `modifier`

- **Breaking behaviour change, documented but not auto-migrated.** Flipping the
  `ignoreTooltipPointerEvents` default reverses what unbound markup does. It is described in
  `tooltip.en.md`/`tooltip.ru.md` and in a new "Behavior changes with no code pattern" section of the
  `v20-upgrade` README. No regex can detect an absent binding, so there is deliberately no rewrite rule.
- **New public API:** `ignoreTooltipPointerEvents` now appears in `title.api.md` and
  `ellipsis-center.api.md`.
- `core/pop-up/pop-up-trigger.ts` and `pop-up.ts` are shared by tooltip, popover, app-switcher and
  notification-center — the consumer suites for all of them run green here, but this is the widest
  blast radius in the series.

---

One of ten pull requests from a single review pass (`review/*`). Every branch carries two commits: the
fixes from the component review, and a second pass over the findings an independent verifier could not
confirm outright.

The branches share files, so they have to be merged one at a time with a rebase in between. Overlap:
`packages/components/core/locales/*` (7 branches), `tools/public_api_guard/components/core.api.md` (7),
`tools/cspell-locales/*.json` (6), `packages/e2e/routes.ts` (4) and
`packages/schematics/src/{collection,migrations}.json` (3).


packages/components/tooltip/__screenshots__/01-light.png
NikGurev, line 1: Тултипы "сжались"

Перепроверь пожалуйста

https://github.com/user-attachments/assets/b2fd2270-ff68-4789-b3a4-b50e74f0de48

NikGurev (approved):

NikGurev:
надо будет потом использовать скилл, сейчас очень тяжело читается migration.ru.md

PR #1949, opened 2026-08-27
feat(list,tree,core)!: allow changing multiple after initialization (#DS-4854)

lskramarov:
## Summary

`multiple` on `kbq-list-selection` and `kbq-tree-selection` was a static host attribute read once in the constructor, so `[multiple]` could not be bound and the selection mode was frozen for the lifetime of the component. It also mapped every unrecognized value to `checkbox`, which is why `multiple="false"` meant *multiple* selection.

It is now a real input, resolved by a shared `resolveMultipleMode` helper in core, and the mode can be changed at any time. Changing it rebuilds the CDK `SelectionModel` — CDK freezes multiplicity at construction — keeps as much of the selection as the new mode can hold, and reports the result to the form control.

Two commits: the feature, then the fixes from a review of it.

## List of notable changes:

- **added** `multiple` as an accessor `@Input` on `kbq-list-selection` and `kbq-tree-selection`, backed by a signal so options, checkboxes and `aria-multiselectable` follow a change without manual change detection
- **added** `resolveMultipleMode` and the `KbqMultipleInput` type to `@koobiq/components/core`, next to the existing `MultipleMode` enum
- **added** an explicit single-selection spelling: `multiple="single"`, `multiple="false"` and `[multiple]="false"`
- **updated** value reporting so a mode change reaches the `ControlValueAccessor`: narrowing keeps the first selected item in render order, drops the rest and emits `selectionChange` for each one
- **updated** both components to rebuild from the `SelectionModel` rather than from their content query, so items a virtual scroll has not instantiated are not silently lost
- **updated** `autoSelect` / `noUnselectLast` to derive their default from the mode through a nullable backing field, so an explicitly set value survives a later mode change
- **updated** documentation (en/ru) for both components with the accepted values, the runtime-change behaviour and the caveats below

**The breaking change.** `multiple="false"`, `multiple="single"` and any other unrecognized value now mean single selection instead of `checkbox`. Nothing in this repo relied on the old fallback, but a consumer writing `multiple="multiple"` — valid HTML habit from `<select multiple="multiple">` — silently loses multi-select. It is called out in the `BREAKING CHANGE` footer of the first commit. No `ng-update` migration is included: a codemod cannot recover the intent behind an arbitrary value. Please confirm that is the right call.

**Value shape on the tree.** `KbqTreeSelection.getSelectedValues()` has always returned a bare value in single selection and an array in multiple selection. Now that the mode can change at runtime, the shape the form control holds changes with it, so the tree re-reports on every multiplicity change — without that, `writeValue` throws `getKbqSelectNonArrayValueError()` on the next round trip. The list is unaffected; it always reports an array.

**A tree inside `kbq-tree-select` now throws.** The select shares its `SelectionModel` with the tree and subscribes to that instance, so the select owns how many nodes may be selected. Changing the tree's multiplicity there throws `getKbqTreeSelectionOwnedMultipleError` rather than being ignored: Angular caches the bound value either way, so a silently dropped write leaves the binding and the tree permanently out of step. Switching only between `checkbox` and `keyboard` is still allowed. Worth a look at whether throwing is too aggressive for your taste — the alternative is a dev-mode warning plus a knowingly divergent binding.

**Subscribers of `selectionModel`.** A mode change replaces the `SelectionModel` instance, so any code holding `selectionModel.changed.subscribe(...)` is left on the discarded one. Both fields are documented to that effect and point at `selectionChange` instead. `packages/docs-examples/components/tree/tree-access-rights` subscribes that way today; it is unaffected because it does not bind `[multiple]`, but it is the shape to watch for.

**Known follow-up, not in this PR.** `setMultipleMode` / `rebuildSelectionModel` / `bindSelectionModel` and the mode accessors are near-duplicates across the two components. Extracting a shared helper alongside `resolveMultipleMode` is worth doing, but not in the same change as the correctness fixes — the two bodies still differ in what the model holds, how they reconcile, and what they emit.

## Verification

- `yarn run unit:components` — 4702 passed, 0 failed
- `yarn run eslint`, `yarn run prettier`, `yarn run check-api` — clean; API snapshots regenerated after a fresh `build:components`
- Manually exercised the runtime flip in `dev:list` and `dev:tree` across `checkbox` / `keyboard` / single, confirming checkboxes, `aria-multiselectable`, the collapsed selection and the reported model value, with no console errors
- New tests were mutation-checked: each targeted fix was reverted individually and the corresponding test fails


artembelik (approved):

packages/components/list/list.ru.md
```diff
@@ -29,7 +30,35 @@
...
+| `multiple`, `multiple="true"`, `[multiple]="true"`                          | checkbox        |
+| `multiple="single"`, `multiple="false"`, `[multiple]="false"`, без атрибута | одиночный выбор |
+
+Любое другое значение понимается как одиночный выбор, а в dev-режиме о нём сообщается в консоль.
+
+`multiple` — обычный input, поэтому режим можно связать и менять в любой момент:
+
+<!-- prettier-ignore -->
+```html
+<kbq-list-selection [multiple]="mode()" [(ngModel)]="selected">
```
artembelik, line 50: может пример добавим? чтобы наглядно было

packages/components/list/list.ru.md
```diff
@@ -29,7 +30,28 @@
...
+    <kbq-list-option [value]="item">{{ item }}</kbq-list-option>
+</kbq-list-selection>
+```
+
+Если в списке было выбрано несколько опций, при переходе к одиночному выбору остаётся первая выбранная,
```
NikGurev, line 55: Стоит дать возможность пользователю настраивать, какая опция будет выбрана при multipleChange
Так можно гибко настраивать и не завязываться только на первую выбранную опцию

packages/components/core/selection/constants.ts
```diff
@@ -1,4 +1,59 @@
...
+ * Values accepted by the `multiple` input of a selection list or tree.
+ *
+ * The bare string literals are spelled out next to {@link MultipleMode} because a static attribute reaches
+ * the input as a plain string, which TypeScript does not consider assignable to a string enum.
+ *
+ * `''` is what a bare `multiple` attribute resolves to. `'single'`, `'false'` and `false` all turn multiple
+ * selection off.
+ */
+export type KbqMultipleInput =
+    MultipleMode | 'checkbox' | 'keyboard' | 'single' | 'true' | 'false' | '' | boolean | null | undefined;
```
NikGurev:
```suggestion
export type KbqMultipleInput =
    MultipleMode | KbqEnumValues<MultipleMode> | 'single' | 'true' | 'false' | '' | boolean | null | undefined;
```

NikGurev (approved):

packages/components/list/list.ru.md
lskramarov: это дефолт и после смены режима им все равно придется модель менять ну т.е. пусть сами делают что хотят, на это нет никаких ограничений

packages/components/list/list.ru.md
lskramarov: не, оба режима в примерах есть, а новый функционал не такой важный что бы под него отдельный пример делать, если попросят добавим.

PR #1950, opened 2026-08-27
fix(core): theme service SSR support (#DS-5470)

artembelik:


packages/components/core/services/theme.service.ts
```diff
@@ -334,28 +408,74 @@ export class KbqThemeService<T extends KbqThemeConfig = KbqThemeConfig> {
...
+            return null;
+        }
+    }
+
+    private watchSystemTheme(): void {
+        const media = this.media;
+
+        // A server-side `matchMedia` stub can return a bare object carrying `matches` and nothing else;
+        // subscribing to that would throw in exactly the environment the guard above exists for.
+        if (typeof media?.addEventListener !== 'function') return;
```
lskramarov, line 482: **Correctness — crash.** `fromEvent()` accepts a target only when **both** `addEventListener` and `removeEventListener` are functions (`isEventTarget`), and it throws `TypeError: Invalid event target` **synchronously at call time**, not at subscribe. This guard checks half the pair, so a `MediaQueryList` stub carrying `addEventListener` and nothing else takes down the `KbqThemeService` constructor — the exact environment the guard exists for.

Reproduced on `b36d50d` with:

```ts
{ provide: KBQ_WINDOW, useValue: { matchMedia: () => ({ matches: false, media: '', addEventListener: () => {} }) } }
```

```
TypeError: Invalid event target
    at fromEvent (rxjs/src/internal/observable/fromEvent.ts:288:11)
    at KbqThemeService.watchSystemTheme (theme.service.ts:458:18)
    at new KbqThemeService (theme.service.ts:393:14)
```

`apps/docs/src/config.server.ts` happens to stub both methods, so this repo's own SSR is unaffected; a consumer whose stub is one-sided is not.

```suggestion
        if (typeof media?.addEventListener !== 'function' || typeof media.removeEventListener !== 'function') return;
```

_Checked locally: `prettier --check` clean, all 60 specs pass._

packages/components/core/services/theme.service.ts
```diff
@@ -154,6 +164,21 @@ export class KbqThemeLocalStorageStore implements KbqThemeStore {
...
     private readonly staticThemeStorageKey = `${this.storageKey}-static`;
 
+    /**
+     * `storage` fires only in the *other* documents of the same origin, never in the one that wrote the
+     * value, so this reports exactly the external changes `KbqThemeStore.changes` is meant to report,
+     * with no echo of this instance's own writes. `EMPTY` where listeners can't be registered at all —
+     * a server-side `KBQ_WINDOW` stub, where there are no other tabs to hear from anyway.
+     */
+    readonly changes: Observable<void> =
+        typeof this.window.addEventListener === 'function'
```
lskramarov: **Correctness — crash.** Same one-sided guard, and here the blast radius is larger: because `fromEvent()` throws at call time, the throw happens in this **field initializer**, so `new KbqThemeLocalStorageStore()` fails → the `KBQ_THEME_STORE` factory fails → `KbqThemeService` fails → the app fails to bootstrap.

Reproduced on `b36d50d` with a `KBQ_WINDOW` that has `addEventListener` but no `removeEventListener`:

```
TypeError: Invalid event target
    at fromEvent (rxjs/src/internal/observable/fromEvent.ts:288:11)
    at new KbqThemeLocalStorageStore (theme.service.ts:175:24)
    at Object.factory (theme.service.ts:325:26)          // KBQ_THEME_STORE
    at new KbqThemeService (theme.service.ts:342:36)
```

```suggestion
        typeof this.window.addEventListener === 'function' && typeof this.window.removeEventListener === 'function'
```

_Checked locally: `prettier --check` clean, all 60 specs pass._

packages/components/core/services/theme.service.ts
```diff
@@ -334,28 +408,74 @@ export class KbqThemeService<T extends KbqThemeConfig = KbqThemeConfig> {
...
+    /** The persisted mode, or `config.mode` when nothing valid is stored. Also used to re-read the store. */
+    private readMode(): KbqThemeMode {
         const stored = this.store.getMode();
 
         return stored === 'auto' || stored === 'light' || stored === 'dark' ? stored : this.config.mode;
     }
 
-    private readInitialStaticTheme(): string | null {
+    /** The persisted static theme, or `config.theme`. Initialization only — see `watchStore()`. */
+    private readStaticTheme(): string | null {
```
lskramarov, line 442: **Correctness.** `watchStore()` below explains, correctly, why the `config.theme` fallback must not be re-applied when the store reports nothing pinned — *"a store cannot tell a cleared pin from one that was never set"*. The same hole is still open on the **reload** path, which this method owns.

`setMode()` clears the pin on every call, and `KbqThemeLocalStorageStore.getStaticTheme()` maps the persisted `''` back to `null`. So once the user picks a mode, the next page load re-pins `config.theme` and silently overrides that choice.

Reproduced on `b36d50d` with `kbqThemeProvider({ theme: 'dark' })` and the default store:

```
setMode('light')  ->  localStorage: kbq-theme-mode="light", kbq-theme-mode-static=""
reload            ->  mode()="light"   staticTheme()="dark"   currentTheme()="dark"
```

The user asked for light, `mode()` agrees, and the page renders dark. `KbqThemeSettings.theme` is documented as *"Used only when nothing is persisted yet in the `KBQ_THEME_STORE`"*, which is not what happens here.

Pre-existing (`readInitialStaticTheme()` had the same body), but this PR renames the method and writes the reasoning that identifies the bug one screen below — worth closing here, e.g. by letting the store distinguish "cleared" from "never set", or by applying `config.theme` only when `getMode()` is `null` too.

packages/components/core/services/theme.service.ts
```diff
@@ -334,28 +408,74 @@ export class KbqThemeService<T extends KbqThemeConfig = KbqThemeConfig> {
...
+     * single source of truth, and writes nothing back: the tab that made the change already persisted it.
+     */
+    private watchStore(): void {
+        this.store.changes?.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
+            this.modeState.set(this.readMode());
+            // Deliberately not `readStaticTheme()`: its `config.theme` fallback applies only when nothing is
+            // persisted yet. A store cannot tell a cleared pin from one that was never set — both read back
+            // as `null` — so applying the fallback here would re-pin the configured theme in every other tab
+            // the moment one tab clears it, which `setMode()` does on every call.
+            this.staticThemeState.set(this.store.getStaticTheme());
```
lskramarov: **Correctness.** This propagates a theme *name* between tabs, but the tabs need not have the same `themes()` registered. When the incoming name is not in this instance's set, `currentTheme()` computes to `null` and `applyTheme(null, themes)` removes **every** theme class from `<body>` — the page ends up with no theme at all rather than a fallback.

Reachable in the shipped docs site: `ThemeStaticSelectionExample` registers `Day`/`Night` on the shared root service only while it is mounted, and `selectTheme('Night')` persists that name. Any other open docs tab hears the `storage` event, pins `Night`, finds no such theme, and loses `kbq-light`/`kbq-dark`.

Reproduced on `b36d50d`:

```
before changes.next() -> currentTheme "light",  body.className "kbq-light"
after  changes.next() -> currentTheme null,     body.className ""
```

Cheapest guard: ignore a name that is not in `themesState()` and keep the current pin, so an unknown value degrades to "no change" instead of "no theme". The same hole exists at init via `readStaticTheme()`, but `watchStore()` is what makes it fire live.

packages/components/core/services/theme.service.ts
```diff
@@ -334,28 +408,74 @@ export class KbqThemeService<T extends KbqThemeConfig = KbqThemeConfig> {
     setMode(mode: KbqThemeMode) {
         this.selectTheme(null);
         this.modeState.set(mode);
+        this.store.setMode(mode);
     }
```
lskramarov, line 421: **Correctness.** `setMode()` performs two separate store writes, and `KbqThemeLocalStorageStore` turns each `setItem` into its own `storage` event. Other tabs therefore run `watchStore()` twice and observe the intermediate state: pin already cleared, mode still the old one.

When the other tab has a pin active, that intermediate state is *visible*. Reproduced on `b36d50d` (stored `mode='light'`, `static='dark'`, writer calls `setMode('dark')`):

```
after first storage event  -> currentTheme "light"  body "kbq-light"
after second storage event -> currentTheme "dark"   body "kbq-dark"
```

— the reader flashes light before landing on dark.

Writing the mode first makes the intermediate state harmless: the pin still wins until it is cleared, so the reader makes one transition instead of two.

```suggestion
    setMode(mode: KbqThemeMode) {
        this.modeState.set(mode);
        this.store.setMode(mode);
        this.selectTheme(null);
    }
```

_In-tab behaviour is unchanged — the signal writes batch either way. Checked locally: all 60 specs pass._

packages/components/core/services/theme.service.ts
```diff
@@ -225,21 +256,63 @@ export class KbqThemeCookieStore implements KbqThemeStore {
...
 
     private readCookie(key: string): string | null {
+        // `DOCUMENT.cookie` first: in the browser it is the only source reflecting writes made since the
+        // page loaded, and on the server it lets an app that already populates it keep that value.
+        return (
+            this.parseCookie(this.documentCookies(), key) ??
+            this.parseCookie(this.request?.headers.get('cookie') ?? '', key)
```
lskramarov: **Correctness (plausible).** `this.request?.` guards `request` being `null`, but not `headers` being absent or not a `Headers`. Everything else this PR adds to the class is wrapped so a bad input cannot take the render down — `documentCookies()`, `parseCookie()`'s `decodeURIComponent`, `writeCookie()` — and this is the one line that isn't.

`REQUEST` is documented as a fetch `Request`, but it is a plain `InjectionToken` with a `null` root factory, so apps routinely provide it themselves. An Express `req` (whose `headers` is a plain object) would throw `TypeError: this.request.headers.get is not a function` out of `getMode()`, during `KbqThemeService` construction — i.e. it fails the whole render, which is precisely what the rest of this diff is defending against.

Would be confirmed by any SSR setup providing `REQUEST` from a Node adapter rather than `@angular/ssr`'s own `Request`. The same defensive shape as its neighbours would close it:

```ts
private requestCookies(): string {
    try {
        return this.request?.headers.get('cookie') ?? '';
    } catch {
        return '';
    }
}
```

lskramarov (commented):
### `/code-review max` — theme service SSR support

10 findings, most severe first: two constructor crashes from a half-checked `fromEvent` guard, three state bugs (`config.theme` re-pinned on reload, an unknown theme name stripping every `<body>` class, cross-tab write ordering), one plausible `REQUEST` shape issue, one stale public doc, one altitude note on duck-typed SSR detection, and two test-quality notes.

Everything except the `REQUEST` item was reproduced locally against `b36d50d` with throwaway Jest probes; the two `suggestion` blocks were applied and verified (`prettier --check` clean, all 60 specs green). The probes were deleted — the working tree is untouched.

_Generated with Claude Code_

packages/components/core/services/theme.service.ts
```diff
@@ -201,6 +226,12 @@ export class KbqThemeLocalStorageStore implements KbqThemeStore {
 @Injectable({ providedIn: 'root' })
 export class KbqThemeCookieStore implements KbqThemeStore {
     private readonly document = inject(DOCUMENT);
+    /**
+     * The incoming request during a server render, `null` in the browser. `DOCUMENT.cookie` is empty on the
+     * server unless the app populates it itself, so reading the request's own `Cookie` header is what makes
+     * this store work there without that extra wiring.
+     */
+    private readonly request = inject(REQUEST, { optional: true });
```
lskramarov, line 242: **Conventions.** This field's doc says reading the request header *"is what makes this store work there without that extra wiring"* — but the class JSDoc immediately above (lines 220–224) still states the opposite:

> Requires the app's SSR bootstrap to populate `DOCUMENT.cookie` from the request.

True before this PR, not any more. It is the public-facing doc for `KbqThemeCookieStore`, so it will send consumers off to build wiring they no longer need.

AGENTS.md: *"Update or remove comments that become outdated because of your changes."* Something like "reads the incoming request's `Cookie` header on the server; a `DOCUMENT.cookie` the app populates itself still takes precedence" would match what the code now does.

packages/components/core/services/theme.service.ts
```diff
@@ -334,28 +408,74 @@ export class KbqThemeService<T extends KbqThemeConfig = KbqThemeConfig> {
...
 
+    /**
+     * Read synchronously at construction, so `systemPrefersDark` is right before the first render — deferring
+     * it to e.g. `afterNextRender` would show the wrong theme first and correct it visibly. Guarded by both
+     * the `typeof` check and the try/catch: `matchMedia` is absent from the server's `KBQ_WINDOW`, and can be
+     * present but throw in sandboxed iframes / under a restrictive CSP.
+     */
+    private matchPrefersDark(): MediaQueryList | null {
+        try {
+            return typeof this.window.matchMedia === 'function' ? this.window.matchMedia(PREFERS_DARK_QUERY) : null;
```
lskramarov: **Altitude.** SSR is being detected by duck-typing the shape of `KBQ_WINDOW` — `typeof matchMedia === 'function'` here, `typeof addEventListener === 'function'` in the store, `try/catch` around `document.cookie`. Angular answers this question directly, and this repo already uses that answer for exactly this purpose (`packages/components/file-upload/single-file-upload.component.ts:292`, `multiple-file-upload.component.ts:313`):

```ts
private readonly isBrowser = isPlatformBrowser(inject(PLATFORM_ID));
```

The duck-typed version is already wrong in both directions here:

* the `changes` comment (lines 170–171) says `EMPTY` is what "a server-side `KBQ_WINDOW` stub" gets — but `apps/docs/src/config.server.ts` stubs `addEventListener`/`removeEventListener`, so the server actually takes the `fromEvent` branch and registers a listener that can never fire;
* the two `Invalid event target` crashes flagged elsewhere in this review exist only because the branch is chosen by shape rather than by platform.

Not arguing to drop the `try/catch` — the sandboxed-iframe / CSP case is a genuine *browser* failure and needs it. The suggestion is to make "am I on the server" a platform check, and leave each guard covering only the thing it is actually about.

packages/components/core/services/theme.service.spec.ts
```diff
@@ -84,6 +108,119 @@ describe('KbqThemeService', () => {
...
+        // A consumer overriding `provideServerWindow` to reflect e.g. a `Sec-CH-Prefers-Color-Scheme` request
+        // header would reassign `matches` to `true` on this same stub shape, so the server renders dark
+        // straight away. Asserted without `TestBed.tick()` — the resolution must be synchronous, since a
+        // server render never gets an `afterNextRender`-style follow-up pass to correct a wrong first guess.
+        TestBed.configureTestingModule({
+            providers: [
+                {
+                    provide: KBQ_WINDOW,
+                    useValue: {
+                        matchMedia: () => ({
```
lskramarov: **Simplification.** This 9-property stub is a character-for-character copy of the one ~30 lines above; the only difference is `matches: true` vs `matches: false`. A helper next to `fakeMediaQueryList()` removes the duplication and makes the two tests state what they actually vary:

```ts
/** Mirrors the server-side `KBQ_WINDOW` from apps/docs/src/config.server.ts. */
function fakeServerWindow(matches: boolean) {
    return {
        matchMedia: () => ({
            addEventListener: () => {},
            dispatchEvent: () => false,
            removeEventListener: () => {},
            matches,
            media: '',
            onchange: null,
            addListener: () => {},
            removeListener: () => {}
        })
    };
}
```

packages/components/core/services/theme.service.spec.ts
```diff
@@ -557,12 +868,61 @@ describe('KbqThemeLocalStorageStore', () => {
...
+
+        store.changes.subscribe(() => emitted++);
+
+        events.emit('kbq-theme-mode');
+        events.emit('some-unrelated-key');
+
+        expect(emitted).toBe(0);
+    });
+
+    it('has an empty changes stream where listeners cannot be registered (e.g. on the server)', () => {
```
lskramarov: **Test coverage.** `{ addEventListener: undefined }` is a no-op: `setup()` builds the window as `{ ...window, ...windowOverrides }`, and spread copies only *own enumerable* properties — `addEventListener` lives on `EventTarget.prototype`, so `{ ...window }` never carried it. Against this repo's jsdom:

```
Object.hasOwn(window, 'addEventListener')  ->  false
typeof {...window}.addEventListener        ->  undefined
```

The test is not vacuous — without the guard, `fromEvent` throws `Invalid event target` — but the override reads as though it removes something, and the real consequence is that **every** test in this `describe` that does not pass `fakeStorageEvents().overrides` runs against the `EMPTY` branch. Nothing in the suite exercises `changes` against a target that has a real `removeEventListener`, which is why the one-sided-guard crash flagged in `theme.service.ts` is invisible from here.

Suggest dropping the dead override (`setup({})` already gives a listener-less window) and adding a case that asserts `changes` is wired when both listener methods are present.

packages/components/core/services/theme.service.ts
artembelik: Убрал `fromEvent()` — слушатель вешается вручную в `new Observable`, так что валидировать цель больше нечему, а ветка выбирается по `isPlatformBrowser`.

packages/components/core/services/theme.service.ts
artembelik: Здесь то же: `fromEvent()` убран, слушатель вешается напрямую и снимается через `destroyRef.onDestroy`, проверка `addEventListener` осталась односторонней.

packages/components/core/services/theme.service.ts
artembelik: Применил второй вариант — `config.theme` подставляется, только когда не сохранён и режим; случай `selectTheme(null)` без `setMode()` остался открыт и описан в JSDoc `KbqThemeSettings.theme`.

packages/components/core/services/theme.service.ts
artembelik: В сторах перешёл на `isPlatformBrowser` (945e07c8a), а `matchPrefersDark()` оставил на проверке самого `matchMedia`: платформенная проверка там сломала бы серверный стаб, отражающий `Sec-CH-Prefers-Color-Scheme`.

lskramarov (approved):

PR #1951, opened 2026-08-27
fix(dl): wrap text (#DS-3323)

artembelik:


packages/components/dl/dl.scss
```diff
@@ -6,7 +6,7 @@
     column-gap: var(--kbq-description-list-size-horizontal-content-gap-horizontal);
     row-gap: var(--kbq-description-list-size-horizontal-gap-vertical);
 
-    grid-template-columns: repeat(4, 1fr);
+    grid-template-columns: repeat(4, minmax(0, 1fr));
```
lskramarov, line 9: **Pinning every track to `minmax(0, 1fr)` shreds ordinary words mid-word.**

`1fr` is `minmax(auto, 1fr)`, and that `auto` minimum is what let the term column grow to the item’s min-content (its longest word). With `minmax(0, 1fr)` the track is pinned to exactly a quarter of the list and the item can no longer push it open.

The cost is visible in the two baselines this PR re-approves. In `E2eDlStates` (250px cells):

| | before | after |
|---|---|---|
| `Multiline term` | `Multiline` / `term` | `Multilin` / `e term` |
| `Multiline description term` | `Multiline` / `description` / `term` | `Multilin` / `e` / `descrip` / `tion` / `term` |

Five lines, one of them a single character. Any consumer whose term is longer than a quarter of the list now gets ordinary prose split mid-word — the rare unbroken-identifier case was fixed by making the common case worse, and neither the docs nor the PR description mention the trade-off.

Worth considering scoping the aggressive breaking (a modifier class or an input) instead of applying it list-wide, so lists made of normal text keep their current layout.

_Generated with Claude Code_

packages/components/dl/dl.scss
```diff
@@ -24,6 +24,12 @@
         justify-items: end;
     }
 
+    .kbq-dt,
+    .kbq-dd {
+        min-inline-size: 0;
```
lskramarov, line 29: **`min-inline-size: 0` breaks the auto-measured resize minimum.**

`measureDtWidth()` (`dl.component.ts:222`) sets `measuredDtWidth` from `terms()[0].getBoundingClientRect().width`, and that value is the documented default for both `dtMinWidth` and `ddMinWidth` (“defaults to the rendered term width”) as well as the separator’s `aria-valuemin`.

Before this change the first term rendered at `max(track share, its own min-content)` because `min-width: auto` held it open, so the derived minimum was always wide enough for the term. With `min-inline-size: 0` the term renders at exactly the 1/4 track share, already broken mid-word — so the measurement now describes the column, not the term.

Concretely: a `<kbq-dl resizable>` whose first term is `Registration timestamp` in a 480px list reports an `aria-valuemin` of ~110px measured off a term wrapped onto two lines, and dragging down to that minimum no longer fits the term — which is the guarantee the auto default existed to provide.

`E2eDlResizableAuto` does not catch this: its 600px host makes the 1/4 share (~142px) wider than `Incident`’s min-content anyway, so the assertion still holds.

_Generated with Claude Code_

packages/docs-examples/components/dl/dl-long-text/dl-long-text-example.ts
```diff
@@ -0,0 +1,25 @@
...
+@Component({
+    selector: 'dl-long-text-example',
+    imports: [KbqDlModule, KbqLinkModule],
+    template: `
+        <kbq-dl [vertical]="false">
+            <kbq-dt>Identifierwithoutseparators</kbq-dt>
+            <kbq-dd>0198EB25C90370D5A8B57414ED82C4D0198EB25C90370D5A8B5B7414ED82C4D</kbq-dd>
+            <kbq-dt>Link</kbq-dt>
+            <kbq-dd>
+                <a kbq-link target="_blank" rel="noopener noreferrer">
```
lskramarov, line 17: **This anchor has no `href`, so it is not a link.**

An `<a>` without `href` has no implicit `link` role, is not activatable, and gets no pointer cursor; `target` and `rel` are inert on it. But `KbqLink` unconditionally binds `[attr.tabindex]="tabIndex"` (`link.component.ts:45`), so the element still enters the tab order — a keyboard user lands on something a screen reader announces as generic text with nothing to activate.

AGENTS.md is explicit here:

> It MUST pass all AXE checks.
> It MUST follow all WCAG AA minimums, including focus management, color contrast, and ARIA attributes.

Every other link example in the repo (`link-overview`, `link-external`, `link-icons`, `link-application`, `link-caption`) uses `href="…" target="_blank" class="kbq-link_external"`. Docs examples are copy-paste templates, so this ships the broken pattern to consumers.

```suggestion
                <a
                    class="kbq-link_external"
                    href="https://github.com/koobiq/angular-components/blob/main/packages/components/dl/index.ts"
                    kbq-link
                    multiline
                    rel="noopener noreferrer"
                    target="_blank"
                >
```

_Generated with Claude Code_

lskramarov (commented):
## Code review — `xhigh`

11 findings. The two that matter most:

1. Pinning the tracks to `minmax(0, 1fr)` removes all min-content pressure from the columns, so **ordinary multi-word terms are now broken mid-word** — visible in the `E2eDlStates` baselines this PR re-approves (`Multiline description term` → `Multilin`/`e`/`descrip`/`tion`/`term`).
2. `min-inline-size: 0` changes what `measureDtWidth()` measures, so the auto-derived `dtMinWidth`/`ddMinWidth` and `aria-valuemin` of a resizable list no longer guarantee the term fits.

The rest cover a hrefless anchor in the docs example, `overflow-wrap` inheritance into projected components, vacuous e2e assertions, coverage gaps, and a stale `example-module.ts`.

_Generated with Claude Code_

packages/components/dl/dl.scss
```diff
@@ -24,6 +24,12 @@
...
+        overflow-wrap: anywhere;
```
lskramarov, line 30: **`overflow-wrap` is inherited, so this leaks into everything projected into a `kbq-dd`.**

The rule sits on the `kbq-dt`/`kbq-dd` host elements, and `overflow-wrap` is an inherited property. A `<button kbq-button>Сохранить изменения</button>` or a `<kbq-tag>` nested in a description now reports a min-content width of one character and its label breaks mid-word in a narrow column, instead of the control holding its natural width. The `E2eDlStates` fixture already nests a `kbq-badge` in a `kbq-dd`; it only escapes because that column happens to be wide enough.

The codebase already separates the two cases:

- `code-block.scss:112` → `overflow-wrap: anywhere` (code, where breaking anywhere is correct)
- `modal.scss:129`, `tree-option.scss:19` → `overflow-wrap: break-word` (prose)

Scoping the rule to the text run rather than the projection host would keep nested components sizing themselves.

_Generated with Claude Code_

packages/components/dl/e2e.playwright-spec.ts
```diff
@@ -13,6 +13,27 @@ test.describe('KbqDlModule', () => {
...
+            const list = component.locator('.kbq-dl');
+            const shortTerm = page.getByTestId('e2eDlShortTerm');
+            const shortDescription = page.getByTestId('e2eDlShortDescription');
+            const longTerm = page.getByTestId('e2eDlLongTerm');
+            const longDescription = page.getByTestId('e2eDlLongDescription');
+
+            expect(await getHeight(longTerm)).toBeGreaterThanOrEqual((await getHeight(shortTerm)) * 2);
+            expect(await getHeight(longDescription)).toBeGreaterThanOrEqual((await getHeight(shortDescription)) * 2);
+            expect(await list.evaluate((element) => element.scrollWidth <= element.clientWidth)).toBe(true);
+            expect(await longTerm.evaluate((element) => element.scrollWidth <= element.clientWidth)).toBe(true);
```
lskramarov, line 32: **These two assertions pass with or without the fix.**

Revert `dl.scss` and re-run: with `repeat(4, 1fr)` the track’s `auto` minimum grows to the long word’s min-content (~380px), the `kbq-dt` is laid out at that full 380px, and `scrollWidth === clientWidth` on the item — green. The overflow lands on the grid *container*, not on the item, so only the `list` assertion on line 31 (plus the two height assertions) actually guards the regression.

As written, lines 32–33 give false confidence that per-element wrapping is covered. Asserting the item width against the container width, or against the expected track share, would make them load-bearing.

_Generated with Claude Code_

packages/components/dl/e2e.ts
```diff
@@ -113,6 +113,34 @@ import { KbqDlModule } from './dl.module';
 })
 export class E2eDlStates {}
 
+@Component({
+    selector: 'e2e-dl-long-text',
+    imports: [KbqDlModule],
+    template: `
+        <kbq-dl [vertical]="false">
```
lskramarov, line 120: **The fixture misses the other layouts this diff changes.**

`dl.scss` also rewrites the `.kbq-dl_wide` track (line 130) and the `.kbq-dl_vertical` track (line 150), and `min-inline-size: 0` newly affects `.kbq-dl_resizable` — but this fixture renders a single plain `<kbq-dl [vertical]="false">`. A future change that reverts either of those two edits, or reintroduces overflow in the resizable path, passes the whole dl suite.

The coverage asymmetry is worth noting: the only visual baseline touched here (`E2eDlStates`) was re-approved to absorb a regression, while the behavior the PR adds gets no baseline at all.

_Generated with Claude Code_

packages/docs-examples/components/dl/index.ts
```diff
@@ -18,7 +20,8 @@ const EXAMPLES = [
     DlHorizontalOverviewExample,
     DlVerticalOverviewExample,
     DlSmallExample,
-    DlResizableExample
+    DlResizableExample,
+    DlLongTextExample
```
lskramarov, line 24: **`example-module.ts` was not regenerated, so the new doc section renders nothing.**

`packages/docs-examples/example-module.ts` is a committed generated file holding both the `EXAMPLE_COMPONENTS` manifest and the `loadExample` switch. It has `dl-overview`, `dl-resizable`, `dl-small`… but no `dl-long-text`.

`docs-live-example-viewer.ts:137`:

```ts
if (!EXAMPLE_COMPONENTS[exampleName]) {
    console.error(`Could not find example: ${exampleName}`);

    return;
}
```

So the `<!-- example(dl-long-text) -->` placeholder added to `dl.en.md`/`dl.ru.md` renders an empty viewer. CI’s `build-docs` action runs `yarn run build:docs-examples-module` first, so the deployed site is fine — but `docs:build`, `docs:start:prod` and anything reading the committed module see the stale file.

```bash
yarn run build:docs-examples-module
```

and commit the result, as 3abffac63 did for the list examples.

_Generated with Claude Code_

packages/components/dl/dl.en.md
```diff
@@ -23,3 +23,9 @@ The `resizable` flag allows changing column widths by dragging the separator.
 Minimum column widths are set with `dtMinWidth` and `ddMinWidth`. Two-way binding with `[(dtWidth)]` is used to set and preserve the width programmatically and synchronize it across lists.
 
 <!-- example(dl-resizable) -->
+
+### Long text
+
+Long strings without spaces, such as identifiers or URLs, wrap automatically and do not affect column widths.
```
lskramarov, line 29: **This documents the improvement but not the regression.**

“Long strings without spaces, such as identifiers or URLs, wrap automatically and do not affect column widths” tells a consumer the change is scoped to separator-free strings. It is not — `overflow-wrap: anywhere` plus `minmax(0, 1fr)` apply to every `kbq-dt`/`kbq-dd`, so a plain term like `Multiline term` in a narrow list now renders as `Multilin` / `e term`, exactly what the re-approved `01-light.png` shows.

A consumer reading this section has no way to anticipate the layout change to their existing lists. Same for `dl.ru.md:29`.

_Generated with Claude Code_

packages/components/dl/dl.scss
```diff
@@ -121,7 +127,7 @@
     }
 
     &.kbq-dl_wide {
-        grid-template-columns: repeat(2, 1fr);
+        grid-template-columns: repeat(2, minmax(0, 1fr));
```
lskramarov, line 130: **These three track edits are redundant given `min-inline-size: 0` on the items.**

Per CSS Grid §6.6 the automatic minimum size applies only while the item’s computed `min-width` is `auto`. `min-inline-size: 0` sets it to `0`, so the minimum contribution is `0` and `1fr` (= `minmax(auto, 1fr)`) already resolves to `minmax(0, 1fr)` for every in-flow item. The only in-flow grid items are `.kbq-dt`/`.kbq-dd` — `.kbq-dl__resize-track` is absolutely positioned and does not size tracks — so lines 9, 130 and 150 change nothing on their own.

They are still worth keeping as a guard for raw elements a consumer projects straight into `<kbq-dl>`, but that should be said in a comment: as it stands two mechanisms express one intent and a future reader cannot tell which one is load-bearing.

_Generated with Claude Code_

packages/components/dl/e2e.playwright-spec.ts
```diff
@@ -13,6 +13,27 @@ test.describe('KbqDlModule', () => {
...
+            const component = page.getByTestId('e2eDlLongText');
+            const list = component.locator('.kbq-dl');
+            const shortTerm = page.getByTestId('e2eDlShortTerm');
+            const shortDescription = page.getByTestId('e2eDlShortDescription');
+            const longTerm = page.getByTestId('e2eDlLongTerm');
+            const longDescription = page.getByTestId('e2eDlLongDescription');
+
+            expect(await getHeight(longTerm)).toBeGreaterThanOrEqual((await getHeight(shortTerm)) * 2);
+            expect(await getHeight(longDescription)).toBeGreaterThanOrEqual((await getHeight(shortDescription)) * 2);
+            expect(await list.evaluate((element) => element.scrollWidth <= element.clientWidth)).toBe(true);
```
lskramarov, line 31: **`toBe(true)` hides the numbers when this fails.**

The comparison happens in-page, so a CI failure prints only `Expected: true / Received: false` — no indication of how far the list overflowed. Returning the pair and asserting on it keeps the measurements in the report:

```ts
const getOverflow = (locator: Locator) =>
    locator.evaluate((element) => ({ scrollWidth: element.scrollWidth, clientWidth: element.clientWidth }));

// …
const { scrollWidth, clientWidth } = await getOverflow(list);

expect(scrollWidth).toBeLessThanOrEqual(clientWidth);
```

The same expression is copy-pasted on lines 31, 32 and 33 — a local helper next to `getHeight` covers all three.

_Generated with Claude Code_

packages/docs-examples/components/dl/dl-long-text/dl-long-text-example.ts
```diff
@@ -0,0 +1,25 @@
...
+                    https://github.com/koobiq/angular-components/blob/main/packages/components/dl/index.ts
```
lskramarov, line 18: **A wrapping-link example should use the `multiline` input.**

`KbqLink.multiline` is documented as “Whether the link has an increased hit area for multiline usage” and toggles `.kbq-link_multiline`. `link-multi-line-example.ts` is the canonical usage: `<a kbq-link class="kbq-link_external" multiline>` wrapping a `<span class="kbq-link__text">`.

This example puts a ~90-character URL into a narrow `kbq-dd` — guaranteed to wrap — and gets the single-line hit area, so the wrapped fragments are harder to click. Since demonstrating wrapping links is the whole point of the example, it should show the component’s own multiline affordance rather than leaving readers to find it elsewhere.

_Generated with Claude Code_

rmnturov (approved):

packages/docs-examples/components/dl/dl-long-text/dl-long-text-example.ts
artembelik: это сделано нарочно, чтобы не уводить пользователя со страницы

lskramarov (approved):

packages/components/dl/dl.scss
artembelik: откатил, более не актуально

packages/components/dl/dl.scss
artembelik: откатил, более не актуально

rmnturov:
В коммите e260f0db2 доработал пример длинного текста и связанное описание:

- добавил длинный обычный текст, чтобы показать перенос по словам;
- добавил длинные идентификаторы, путь и URL, чтобы показать перенос строк без пробелов;
- добавил многострочную ссылку на MITRE ATT&CK;
- исправил SHA-256: 64 шестнадцатеричных символа в нижнем регистре и моноширинном начертании;
- вычитал русское описание по редполитике и синхронизировал английскую версию.

Проверки: ESLint, Prettier, git diff --check и production-сборка документации.

PR #1952, opened 2026-08-27
feat(dropdown): added secondary action (#DS-3869)

NikGurev:
## Summary

Secondary action in dropdown item with keyboard navigation
Handles overlay properly, including progress state


lskramarov (commented):
### Code review — `xhigh` (recall-oriented)

15 findings, most severe first. The top three are behavioural regressions that the new specs and screenshots do not cover; the rest are smaller correctness gaps, test-quality issues and conventions.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item.scss
```diff
@@ -48,10 +63,27 @@
...
+
+    & > .kbq-dropdown-item__action {
+        flex: 0 0 auto;
+    }
+
+    // Hidden by default so the action is clickable — but `progress` wins: see the `.kbq-progress`
+    // override below, which has higher specificity (3 classes vs. 2) and restores `display` so the
+    // shimmer (and the full-cover click-block that comes with it) takes over the whole row again,
+    // action included.
+    & > .kbq-dropdown-item-overlay {
```
lskramarov: **Hiding the overlay removes the disabled click guard.**

`.kbq-dropdown-item-overlay` is not only the progress shimmer — it is the element that swallows clicks on a disabled item (`(click)="haltDisabledEvents($event)"` → `stopImmediatePropagation()`), which is what keeps the event from ever reaching the host listeners.

With `display: none` here, a click on a disabled item now reaches the host. `KbqDropdownItem.checkDisabled` only calls `stopPropagation()`, and that does **not** stop other listeners on the *same* element — so for

```html
<div kbq-dropdown-item disabled [kbqDropdownTriggerFor]="sub">
    Disabled
    <a kbq-dropdown-item-action href="/settings">Settings</a>
</div>
```

clicking the item padding (outside the disabled `<button>`) still runs `KbqDropdownTrigger.handleClick` (`dropdown-trigger.directive.ts:424`, which has no disabled guard) and **opens the submenu of a disabled item**. `handleHover` guards on `!active.disabled`, `handleClick` never did — it relied on the overlay.

Suggestion: keep the overlay rendered and raise the action above it (`.kbq-dropdown-item__action { position: relative; z-index: 1 }`) instead of switching the guard off.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item.component.ts
```diff
@@ -42,23 +53,33 @@ import { KBQ_DROPDOWN_PANEL, KbqDropdownPanel } from './dropdown.types';
         '[class.kbq-dropdown-item_highlighted]': 'highlighted',
         '[class.kbq-disabled]': 'disabled',
         '[class.kbq-progress]': 'progress()',
+        '[class.kbq-dropdown-item_has-action]': '!!itemAction()',
 
         '[attr.disabled]': 'disabled || null',
-        '[attr.tabindex]': 'getTabIndex()',
+        '[attr.tabindex]': 'itemAction() ? null : getTabIndex()',
```
lskramarov: **Nested trigger + action: focus restoration silently fails.**

Dropping `tabindex` from the host makes it unfocusable, but `KbqDropdownTrigger` still focuses the raw host element:

```ts
// dropdown-trigger.directive.ts:355
focus(origin?: FocusOrigin, options?: FocusOptions) {
    if (this.focusMonitor && origin) {
        this.focusMonitor.focusVia(this.elementRef.nativeElement, origin, options);
    } else {
        this.elementRef.nativeElement.focus();
    }
}
```

For the nested-trigger-with-action case this PR explicitly supports (`e2eDropdownItemActionNested`, `ActionDropdown` item 3): open the submenu, press <kbd>Escape</kbd> or <kbd>←</kbd>. `destroy()` calls `this.focus(this.openedBy)` (`:460`) on a `<div>` with no `tabindex`, so focus falls back to `<body>` — the parent panel stops receiving keydown events and arrow/Escape navigation is dead from there on.

The trigger should delegate to `dropdownItemInstance.focus(...)` when it has one, rather than focusing `elementRef.nativeElement` directly.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item.html
```diff
@@ -1,8 +1,20 @@
-<ng-content select="[kbq-icon]" />
+<ng-template #primaryContent>
+    <ng-content select="[kbq-icon]" />
```
lskramarov: **The leading icon is no longer a direct child of the item, which breaks its sizing rules.**

With an action present this `ng-content` renders inside `<button class="kbq-dropdown-item__primary">`, so `dropdown-item.scss:20` no longer matches:

```scss
.kbq-dropdown-item {
    & > .kbq-icon:first-child {
        &:before,
        & > svg { align-self: center; display: flex; height: var(--kbq-size-xl); width: 16px; align-items: center; }
    }
}
```

So `<div kbq-dropdown-item><i kbq-icon="…"></i>Text<a kbq-dropdown-item-action/></div>` renders the leading icon unsized/misaligned. No example, e2e fixture or spec in this PR combines a leading `[kbq-icon]` with an action, so the new screenshots do not cover it.

Either keep the icon outside the primary button, or relax the selector to a descendant one.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item.component.ts
```diff
@@ -96,9 +117,10 @@ export class KbqDropdownItem implements KbqTitleTextRef, IFocusableOption, After
         if (this.focusMonitor) {
             // Start monitoring the element so it gets the appropriate focused classes. We want
             // to show the focus style for menu items only when the focus was not caused by a
-            // mouse or touch interaction.
-
-            this.focusMonitor.monitor(this.elementRef, false);
+            // mouse or touch interaction. When `itemAction` is present, focus never lands on the
+            // host itself (it lands on `primaryAction`/`itemAction`), so children must be checked
+            // too for the focused classes to still be applied to the host.
+            this.focusMonitor.monitor(this.elementRef, !!this.itemAction());
```
lskramarov, line 117: **`checkChildren` is snapshotted once and never updated.**

`itemAction()` is a content-query signal, and every other consumer of it is reactive (host classes, `tabindex`, the template `@if`, `focus()`). This one read is frozen at `ngAfterViewInit`.

For a conditionally projected action:

```html
<div kbq-dropdown-item>
    Text
    @if (canEdit()) { <a kbq-dropdown-item-action href="/settings">Settings</a> }
</div>
```

when `canEdit()` flips to `true` afterwards, `checkChildren` stays `false`: focus lands on the generated `<button>`/action, the host never gets `.cdk-focused`, and the item loses its hover/focus highlight (`_dropdown-theme.scss:26`). The reverse flip leaves `checkChildren: true` on an item that has no focusable children left.

Re-applying `monitor()` from an `effect()` on `itemAction()` would keep it in sync.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item-action.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Always stops the click from bubbling to the item's own primary click handler.
+     * Additionally, blocks the action's own default behavior when the parent item is
+     * disabled/loading — `<a>` has no native `disabled`.
+     */
+    onClick(event: MouseEvent): void {
+        event.stopPropagation();
+
+        if (this.isInactive()) {
```
lskramarov, line 82: **The `progress` half of this guard is unreachable by mouse.**

When the item is in `progress`, `.kbq-dropdown-item.kbq-progress .kbq-dropdown-item-overlay` restores `display: block`, and that overlay covers the whole row — including the action. So the click never reaches this listener.

Clicking the gear on `<div kbq-dropdown-item progress><a kbq-dropdown-item-action href="/settings"/></div>` (the `e2eDropdownItemActionProgress` fixture) hits the overlay instead: `haltDisabledEvents` is a no-op (the item is not `disabled`), the event bubbles to `.kbq-dropdown__panel`’s `(click)="close()"`, and **the whole dropdown closes** — the opposite of the documented “unavailable”.

The unit test at `dropdown.spec.ts:1330` dispatches the click directly on the action element, so jsdom’s lack of hit-testing hides this. Worth an e2e/browser-level check, or blocking the click on the overlay when `progress()` is set.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item.html
```diff
@@ -1,8 +1,20 @@
...
-<div #kbqTitleText class="kbq-dropdown-item__text">
-    <ng-content />
-</div>
+    <div #kbqTitleText class="kbq-dropdown-item__text">
+        <ng-content />
+    </div>
+</ng-template>
+
+@if (itemAction()) {
+    <button #primaryAction type="button" class="kbq-dropdown-item__primary" [disabled]="disabled">
```
lskramarov: **`kbq-title` measures the row, not this truncating box.**

`.kbq-dropdown-item__primary` is the element that actually clips the text now, but it is not marked `#kbqTitleContainer`, so `KbqTitleDirective.parent` still falls back to the item host (`title.directive.ts:157-163`) whose `offsetWidth` includes the action and the gap.

The docs example added in this PR hits exactly that: `<div kbq-dropdown-item kbq-title>` + action. With a 250px row, a 24px action and an 8px gap, text needing 210px is visibly ellipsised inside the 194px primary box, yet `isHorizontalOverflown` (`250 < 210`) is `false` → no tooltip.

```suggestion
    <button #primaryAction #kbqTitleContainer type="button" class="kbq-dropdown-item__primary" [disabled]="disabled">
```

_Generated with Claude Code_

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -1229,6 +1229,155 @@ describe('KbqDropdown', () => {
...
+            const item = getItems()[0];
+            const primaryEl = item.querySelector('.kbq-dropdown-item__primary') as HTMLElement;
+
+            dispatchKeyboardEvent(primaryEl, 'keydown', TAB);
+            fixture.detectChanges();
+
+            expect(fixture.componentInstance.closeCallback).not.toHaveBeenCalled();
+        });
+
+        it('should still recognize Tab via `event.key` when `event.keyCode` is unreliable', () => {
```
lskramarov: **This test cannot fail.**

Remove `|| event.key === 'Tab'` from `handleActionKeydown` and it still passes. `KbqDropdown.handleKeydown` (`dropdown.component.ts:520`) switches on `event.keyCode`, and CDK’s `ListKeyManager.onKeydown` matches `case TAB` on `keyCode` too — so a `keyCode: 0` event never reaches `tabOut`, and `closeCallback` is never called regardless of what the item does.

That also points at a real inconsistency: the panel and the key manager stay keyCode-only, so the `event.key` fallback in the item does not actually make Tab work end-to-end in a browser that reports `keyCode: 0` — it only makes this one handler agree with a path that never fires.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item-action.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+ * element to be non-interactive (e.g. `<div kbq-dropdown-item>`), since interactive content
+ * cannot validly nest inside another interactive element.
+ */
+@Directive({
+    selector: '[kbq-dropdown-item-action]',
+    host: {
+        class: 'kbq-dropdown-item__action',
+        '[class.kbq-disabled]': 'isInactive()',
+        '[attr.aria-disabled]': 'isInactive() || null',
+        '[attr.tabindex]': "isInactive() ? '-1' : null",
```
lskramarov: **This binding clobbers `KbqIconButton`’s `tabindex`.**

`KbqIconButton` — the composition every example and e2e fixture here uses — binds `'[attr.tabindex]': 'tabindex'` and returns `0` when enabled (`icon-button.component.ts:35`, `:63-69`). The component’s host bindings run before this directive’s, so writing `null` here removes the attribute again on every *active* action, and a custom `[tabindex]` passed to `kbq-icon-button` is silently discarded.

It only looks fine because the examples use `<a href>`, which is focusable without a `tabindex`; `<div kbq-icon-button kbq-dropdown-item-action>` would not be reachable at all.

Consider only *adding* `-1` when inactive (e.g. via `inert`/an explicit stored value) rather than binding `null` for the active case.

_Generated with Claude Code_

packages/components/dropdown/dropdown.en.md
```diff
@@ -12,6 +12,22 @@ Add the `progress` attribute to `kbq-dropdown-item` to show a loading shimmer on
...
 
+### Secondary action
+
+You can add an extra button inside a menu item (for example, to go to settings). It works independently: clicking it doesn't close the menu, and you can reach it with the Tab key.
+
+The menu item itself should be a regular block, not a button.
+
+If the item is in a loading state, the extra button is hidden and unavailable.
+
+<!-- example(dropdown-item-action) -->
```
lskramarov, line 23: **`packages/docs-examples/example-module.ts` was not regenerated.**

The two new markers reference example ids that do not exist in the committed registry — `grep -n "dropdown-item-action\|dropdown-item-link" packages/docs-examples/example-module.ts` on this branch returns nothing.

That file is committed, and every prior example-adding commit regenerated it (e.g. `3abffac63 feat(list): drag and drop reordering`). CI does run `build:docs-examples-module` before building docs, so the site is fine — but the committed artifact is now stale on `main`: a local `docs:build` without running the generator renders nothing for these markers, and the next unrelated PR that regenerates will carry a diff it did not author.

Run `yarn run build:docs-examples-module` and commit the result.

_Generated with Claude Code_

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -1229,6 +1229,155 @@ describe('KbqDropdown', () => {
...
+        });
+
+        it('should add the has-action modifier when an action is projected', () => {
+            const item = getItems()[0];
+
+            expect(item.classList).toContain('kbq-dropdown-item_has-action');
+            expect(item.classList).not.toContain('kbq-progress');
+        });
+
+        it('should move focus from the primary button to the action on Tab without closing the dropdown', () => {
```
lskramarov: **Name ↔ body mismatch: neither of these tests checks focus.**

This one and `should move focus from the action back to the primary button on Shift+Tab…` (`:1286`) assert only `expect(closeCallback).not.toHaveBeenCalled()`. A synthetic `keydown` in jsdom moves no focus at all, so the only thing actually under test is `stopPropagation`.

A change that keeps the dropdown open but breaks the primary↔action pairing (e.g. narrowing `movesToAction`’s `target === primaryEl` check) passes both. Either rename them to what they assert, or focus the elements explicitly and assert the resulting `document.activeElement`.

_Generated with Claude Code_

packages/components/dropdown/dropdown.en.md
```diff
@@ -12,6 +12,22 @@ Add the `progress` attribute to `kbq-dropdown-item` to show a loading shimmer on
 
 <!-- example(dropdown-item-progress) -->
 
+### Secondary action
+
+You can add an extra button inside a menu item (for example, to go to settings). It works independently: clicking it doesn't close the menu, and you can reach it with the Tab key.
+
+The menu item itself should be a regular block, not a button.
+
+If the item is in a loading state, the extra button is hidden and unavailable.
```
lskramarov: **The action is not hidden while loading.**

Same for `dropdown.ru.md:22` («дополнительная кнопка скрыта и недоступна»). `kbq-progress-overlay()` paints `rgba(255,255,255,.1)` / `rgba(0,0,0,.04)` stripes (`core/styles/common/_animation.scss:14-15`), so the gear stays fully legible — the new `02-light.png` baseline shows it. Only pointer and tab access are removed.

Suggest wording along the lines of “the extra button stays visible but becomes unavailable”.

_Generated with Claude Code_

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
```diff
@@ -0,0 +1,61 @@
...
+                <a kbq-icon-button kbq-dropdown-item-action aria-label="Hash sum settings" href="/settings/hash-sum">
+                    <i kbq-icon="kbq-gear_16" [color]="'contrast-fade'"></i>
+                </a>
+            </div>
+
+            <div kbq-dropdown-item disabled>
+                Not available
+                <a
+                    disabled
+                    kbq-icon-button="kbq-gear_16"
```
lskramarov, line 49: **Two markup forms in one example, plus copy-pasted `aria-label`s.**

This item uses the `kbq-icon-button="kbq-gear_16"` + `color="contrast-fade"` attribute form while its three siblings nest `<i kbq-icon [color]="'contrast-fade'">`. The two forms produce different DOM — `_dropdown-theme.scss:7` (`.kbq-icon.kbq-empty:not(.kbq-icon-button, .kbq-icon-item)`) explicitly excludes `.kbq-icon-button` — so the disabled row’s icon can theme differently from the others for no stated reason.

Also, `aria-label="Hash sum settings"` is copied onto the “Not available” item (line 40) and the “Nested” item (line 48). A screen reader announces “Hash sum settings” for rows that are about something else; the example is what consumers copy.

_Generated with Claude Code_

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -2421,6 +2570,44 @@ class FooterDropdown {
...
+            </div>
+        </kbq-dropdown>
+
+        <kbq-dropdown #submenu="kbqDropdown">
+            <button kbq-dropdown-item>Sub item</button>
+        </kbq-dropdown>
+    `
+})
+class ActionDropdown {
+    readonly trigger = viewChild.required<KbqDropdownTrigger>('trigger');
```
lskramarov: **Dead `viewChild` in the fixture.**

`trigger` (and the `#trigger="kbqDropdownTrigger"` template ref it reads) is never used — every test in the `item action` block reaches the trigger through `triggerEl()`.

Also `createComponent(ActionDropdown, [], [])` at `:1240` passes the two arguments that already default to `[]`; other describes in this file just call `createComponent(X)`.

_Generated with Claude Code_

packages/components/dropdown/e2e.ts
```diff
@@ -306,3 +306,68 @@ export class E2eDropdownTitleOverflow {
     protected readonly longValue =
         'Just a text and a long text and a long text and a long text and a long text and a long text and a long text';
 }
+
+@Component({
+    selector: 'e2e-dropdown-item-action',
+    imports: [KbqDropdownModule, KbqButtonModule, KbqIconModule, KbqIconButton],
```
lskramarov: `KbqIconModule` already exports `KbqIconButton` (`packages/components/icon/icon.module.ts` lists it in both `imports` and `exports`), so the standalone entry is redundant here and in `dropdown-item-action-example.ts:12`.

_Generated with Claude Code_

packages/components/dropdown/dropdown-item-action.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+    protected isInactive(): boolean {
+        return !!this.dropdownItem?.disabled || !!this.dropdownItem?.progress();
+    }
+
+    /**
+     * Always stops the click from bubbling to the item's own primary click handler.
+     * Additionally, blocks the action's own default behavior when the parent item is
+     * disabled/loading — `<a>` has no native `disabled`.
+     */
+    onClick(event: MouseEvent): void {
```
lskramarov: **Host-only callback declared public.**

AGENTS.md: “Use `protected` for template bindings”. `onClick` is referenced only from the `host` metadata, as is `KbqDropdownItem.handleActionKeydown`. Both land in `tools/public_api_guard/components/dropdown.api.md`, so they become supported API that cannot be renamed or removed without a breaking-change entry.

`isInactive()` right above is correctly `protected`, so the class is internally inconsistent too.

_Generated with Claude Code_

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
```diff
@@ -0,0 +1,61 @@
...
+                    kbq-dropdown-item-action
+                    aria-label="Hash sum settings"
+                    href="/settings/hash-sum"
+                    color="contrast-fade"
+                ></a>
```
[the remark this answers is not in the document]
NikGurev: that's mostly used for kbq-icon-button styling

packages/components/dropdown/dropdown.en.md
NikGurev: Low

packages/components/dropdown/__screenshots__/02-dark.png
artembelik, line 1: в тесте бы показать, что будет если текст слева не влезет в блок, будет троеточие?

packages/components/dropdown/dropdown-item-action.ts
artembelik, line 1: префикс directive не рекомендуется использовать (по angular styleguide), темболее в имени класа его нет

packages/components/dropdown/dropdown-item-action.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+);
+
+/**
+ * Marks a secondary, independently-focusable/clickable element (e.g. a settings link) projected
+ * inside a `kbq-dropdown-item` alongside the item's own primary action. Requires the item's host
+ * element to be non-interactive (e.g. `<div kbq-dropdown-item>`), since interactive content
+ * cannot validly nest inside another interactive element.
+ */
+@Directive({
+    selector: '[kbq-dropdown-item-action]',
```
artembelik: по стайлгайду:
```suggestion
    selector: '[kbqDropdownItemAction]',
```

packages/components/dropdown/dropdown-item.html
```diff
@@ -1,8 +1,20 @@
...
 
-<div #kbqTitleText class="kbq-dropdown-item__text">
-    <ng-content />
-</div>
+    <div #kbqTitleText class="kbq-dropdown-item__text">
+        <ng-content />
+    </div>
+</ng-template>
+
+@if (itemAction()) {
```
artembelik: может быть сразу реализуем подход префикс+суффикс аналогичный `KbqButtonSuffix`/`KbqButtonPrefix` в кнопках, `KbqPrefix`/`KbqSuffix` в фф, `KbqTagPrefix`/`KbqTagSuffix` в тегах?

@NikGurev @lskramarov что думаете?

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
```diff
@@ -0,0 +1,61 @@
...
+            </div>
+
+            <div kbq-dropdown-item disabled>
+                Not available
+                <a
+                    disabled
+                    kbq-icon-button="kbq-gear_16"
+                    kbq-dropdown-item-action
+                    aria-label="Hash sum settings"
+                    href="/settings/hash-sum"
```
artembelik: давай уберем все href они уводят на 404 страницу, либо добавить открытие в новом окне и с валидной ссылкой

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
artembelik, line 1: давай новые примеры сделаем аналогично (высота блока и кнопка-триггер по центру)  "Состояние загрузки" и "Футер" которые добавлены недавно:
Image

packages/components/dropdown/dropdown-item-action.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    /** Returns the host DOM element. */
+    getHostElement(): HTMLElement {
+        return this.elementRef.nativeElement;
+    }
+
+    /** `progress` takes priority over the action, same as `disabled`. */
+    protected isInactive(): boolean {
+        return !!this.dropdownItem?.disabled || !!this.dropdownItem?.progress();
+    }
```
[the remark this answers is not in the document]
NikGurev: Сделали так, что если использовать эту кнопку-действие не внутри пункта меню, приложение сразу покажет ошибку — а не будет вести себя непонятно.

packages/components/dropdown/dropdown-item.component.ts
```diff
@@ -96,9 +117,10 @@ export class KbqDropdownItem implements KbqTitleTextRef, IFocusableOption, After
         if (this.focusMonitor) {
             // Start monitoring the element so it gets the appropriate focused classes. We want
             // to show the focus style for menu items only when the focus was not caused by a
-            // mouse or touch interaction.
-
-            this.focusMonitor.monitor(this.elementRef, false);
+            // mouse or touch interaction. When `itemAction` is present, focus never lands on the
```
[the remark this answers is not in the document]
NikGurev: Пока не стали добавлять отдельную проверку в коде — в комментариях к директиве и так написано, что она рассчитана на обычный блок, а не кнопку. Если понадобится жёсткая проверка — вынесем отдельной задачей.

packages/docs-examples/components/dropdown/dropdown-item-link/dropdown-item-link-example.ts
```diff
@@ -0,0 +1,26 @@
...
+    template: `
+        <button kbq-button [kbqDropdownTriggerFor]="actionsDropdown">
+            Actions
+            <i kbq-icon="kbq-chevron-down-s_16"></i>
+        </button>
+
+        <kbq-dropdown #actionsDropdown="kbqDropdown">
+            <a kbq-dropdown-item href="https://angular.dev/" target="_blank" rel="noopener">Angular</a>
+            <a kbq-dropdown-item href="https://www.typescriptlang.org/" target="_blank" rel="noopener">Typescript</a>
+            <a kbq-dropdown-item href="https://storybook.js.org/" target="_blank" rel="noopener">Storybook</a>
```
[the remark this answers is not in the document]
NikGurev: Поправили — добавили noreferrer и исправили название TypeScript (с большой S).

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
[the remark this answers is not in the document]
NikGurev: Убрали некорректный атрибут у ссылки, теперь состояние "недоступно" задаётся правильным способом через Angular.

packages/components/dropdown/public-api.ts
```diff
@@ -1,6 +1,7 @@
 export * from './dropdown-animations';
 export * from './dropdown-content.directive';
 export * from './dropdown-errors';
+export * from './dropdown-item-action.directive';
```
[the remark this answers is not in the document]
NikGurev: Скрыли лишние технические детали, которые не должны быть видны снаружи библиотеки — наружу теперь торчит только сама кнопка-действие.

packages/components/dropdown/dropdown-item.scss
NikGurev: Починили: раньше у отключённого пункта с вложенным подменю клик по пустому месту всё равно открывал подменю. Теперь клик по отключённому или загружающемуся пункту действительно ничего не делает.

packages/components/dropdown/dropdown-item.component.ts
NikGurev: Переделали так, что фокус после закрытия подменю всегда возвращается на исходный пункт меню, даже если у него есть кнопка-действие.

packages/components/dropdown/dropdown-item.html
NikGurev: Убрали лишний вложенный элемент, из-за которого иконка слева могла отображаться не того размера.

packages/components/dropdown/dropdown-item.component.ts
NikGurev: Теперь подсветка пункта при фокусе корректно работает, даже если кнопка-действие появляется или пропадает динамически.

packages/components/dropdown/dropdown-item-action.ts
NikGurev: Починили вместе с проблемой выше (клик по кнопке у пункта в состоянии загрузки закрывал всё меню) — теперь клик просто ничего не делает, как и должно быть.

packages/components/dropdown/dropdown-item.html
NikGurev: Эта проблема ушла сама после того, как убрали лишний вложенный элемент — подсказка при обрезанном тексте снова считается правильно.

packages/components/dropdown/dropdown.spec.ts
NikGurev: Убрали тест, который ничего реально не проверял, и заодно упростили сам код — так надёжнее.

packages/components/dropdown/dropdown-item-action.directive.ts
NikGurev: Починили: раньше при использовании готовой кнопки-иконки вместе с этим действием у неё пропадала возможность попасть туда через Tab. Теперь всё работает.

packages/components/dropdown/dropdown.en.md
NikGurev: Перегенерировали файл со списком примеров — новые примеры теперь подхватываются как надо.

packages/components/dropdown/dropdown.spec.ts
NikGurev: Переименовали тесты, чтобы название соответствовало тому, что они реально проверяют.

packages/components/dropdown/dropdown.en.md
NikGurev: Поправили текст в документации — кнопка не прячется во время загрузки, а просто становится неактивной, так и написали.

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
NikGurev: Привели пример к единому виду и сделали подписи для скринридера разными — раньше они были одинаковые для разных пунктов.

packages/components/dropdown/dropdown.spec.ts
NikGurev: Убрали неиспользуемый код в тестах.

packages/components/dropdown/e2e.ts
NikGurev: Убрали лишний импорт, он и так был доступен через общий модуль с иконками.

packages/components/dropdown/dropdown-item-action.directive.ts
NikGurev: Закрыли внутренний метод от использования снаружи, как и другие похожие методы в проекте.

packages/components/dropdown/__screenshots__/02-dark.png
NikGurev: Добавили отдельный пункт меню с длинным текстом и кнопкой рядом — тест теперь проверяет, что при наведении показывается подсказка с полным текстом, если он не помещается. Скриншоты-эталоны обновил.

packages/components/dropdown/dropdown-item-action.ts
NikGurev: Переименовали файл, убрали лишнее слово directive в названии.

packages/components/dropdown/dropdown-item-action.directive.ts
NikGurev: Название атрибута уже поправили на этот вариант, спасибо.

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
NikGurev: Заменили ссылки-заглушки на настоящие страницы документации, чтобы не вели в никуда.

packages/docs-examples/components/dropdown/dropdown-item-action/dropdown-item-action-example.ts
NikGurev: Поправили — оба новых примера теперь выглядят так же, как остальные (по центру, с одинаковой высотой блока).

packages/components/dropdown/dropdown-item.html
NikGurev: Не хотелось бы новые директивы плодить, не знаю насколько оправдано

dropdown action  еще и клик обрабатывает

NikGurev: Предлагаю отдельно сделать, заодно дизайн обновим и примеры добавим

lskramarov (approved):

artembelik (approved):

PR #1953, opened 2026-08-27
feat(file-upload)!: removed deprecated outputs (#DS-5229)

NikGurev:


packages/schematics/src/migrations/file-upload-deprecated-outputs/data.ts
```diff
@@ -0,0 +1,22 @@
...
+ * `.fileQueueChanged` property access / `.subscribe(...)` call in `.ts`.
+ */
+
+export interface Replacement {
+    from: string;
+    to: string;
+}
+
+export const replacements: Replacement[] = [
+    { from: '\\bfileQueueChanged\\b', to: 'filesChange' },
```
lskramarov, line 20: **The rename is not scoped to koobiq usage.**

These two regexes are applied to the full text of every `.ts` and `.html` file in the workspace, with no gate on a `@koobiq` import, on the `kbq-file-upload` element, or on being inside a template/property-access position. So they also rewrite string literals, HTML text and attribute values, and any identifier of the consumer's own that happens to carry the same name.

Concretely, in a consumer app all of these are silently rewritten with no compile error:

```ts
analytics.track('fileQueueChanged', payload);   // -> 'filesChange'  (event name changes)
t('fileQueueChanged');                          // -> 'filesChange'  (i18n key stops resolving)
@Output('fileQueueChanged') done = ...;         // -> 'filesChange'  (their own public alias)
```

```html
<div class="fileQueueChanged">                  <!-- -> class="filesChange", CSS stops matching -->
```

The README's claim that the names "don't collide with any other Angular or TypeScript identifier" is a statement about library identifiers; it says nothing about the consumer's own code, which is what this schematic edits.

The sibling migrations shipped in this same release solve exactly this by scoping edits to the koobiq element: `tag-slots`, `dropdown-demote-overlay` and `button-supported-colors` all go through `parseTemplate` + `collectInlineTemplateRanges` from `utils/typescript.ts`. Consider doing the same here, and restricting the `.ts` property-access rewrite to member accesses rather than raw text.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+            );
+
+            const updated = (await run(first)).readText(ts);
+
+            expect(updated).toContain('this.upload().fileChange.subscribe(');
+            expect(updated).not.toContain('fileQueueChange');
+        });
+    });
+
+    describe('idempotency and no-op', () => {
```
lskramarov, line 145: **The safety claims this migration rests on are untested.**

The whole argument in `data.ts` and `README.md` is "`\b` word boundaries make the rename safe, and `node_modules`/`dist` are skipped". None of that is asserted anywhere:

- no test that `onFileQueueChanged(...)`, `fileQueueChangedAt` or `myFileQueueChange` are left alone;
- no test that a file under `node_modules/` or `dist/` is skipped;
- no test that `--project` scoping actually limits the walk (`createTestApp` builds a single project, so that branch never runs);
- no test of the dry-run log output.

Drop the `\b` from either entry in `data.ts` and the entire suite still passes.

Also, `it('does not touch a file that already uses the canonical outputs')` below writes `(filesChange)` and asserts it is unchanged — that passes even if the schematic were a complete no-op. A negative test with a near-miss identifier in the same file would carry real signal.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
```diff
@@ -0,0 +1,65 @@
...
+            touched++;
+
+            if (fix) {
+                tree.overwrite(filePath, content);
+            } else {
+                logMessage(context.logger, [`${LABEL} would update ${filePath} (run with --fix to apply)`]);
+            }
+        });
+
+        logMessage(context.logger, [
```
lskramarov, line 72: **Nothing is reported back to the user.**

The final message is a bare file count, and the per-file dry-run line names only the path. Combined with the unscoped rename, that makes these the least reviewable edits in the collection — a user sees `updated 37 file(s)` with no way to tell which of those were their own unrelated symbols.

Every comparable migration in this release does more:

- `filter-bar-rename-action`, `locale-configuration-providers` and `button-supported-colors` append a `BEHAVIOUR_NOTE` telling the user what to check by hand;
- `css-selectors` prints the actual `from -> to` pairs it matched, per file.

At minimum, a closing note saying the rewrite is textual and the diff must be reviewed before committing would be worth adding — the migration guide already tells readers that about regex-based schematics.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+function isMigratableFile(filePath: string): boolean {
+    return filePath.endsWith(TS_EXT) || filePath.endsWith(HTML_EXT);
+}
+
+function applyReplacements(content: string): string {
+    let result = content;
+
+    for (const { from, to } of replacements) {
+        result = result.replace(new RegExp(from, 'g'), to);
```
lskramarov: **Cheap pre-check dropped, and the regexes are recompiled per file.**

This file is otherwise a close copy of `scrollbar-deprecated-path/index.ts`, but that one guards the regex with a substring test first (`index.ts:48`):

```ts
if (!originalContent || !originalContent.includes('@koobiq/components/scrollbar')) return;
```

Here there is no guard, so every `.ts` and `.html` file in the workspace gets two freshly compiled `RegExp`s run over its full text. One substring check covers both names, since `fileQueueChange` is a prefix of `fileQueueChanged` — adding this next to the existing emptiness check in the visitor would skip almost every file:

```ts
if (!originalContent.includes('fileQueueChange')) return;
```

And the two `RegExp` objects can be built once at module scope instead of once per file.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/data.ts
```diff
@@ -0,0 +1,22 @@
...
+ * `KbqSingleFileUploadComponent.fileQueueChange` were deprecated in favor of
+ * `filesChange` / `fileChange` and are removed in this release. Both names
+ * are unambiguous — they don't collide with any other Angular or TypeScript
+ * identifier — so a plain `\b…\b` word-boundary rename is safe everywhere
+ * the identifier can appear: an `(fileQueueChanged)="…"` template binding
+ * (external `.html` or inline `template:` string), or a programmatic
+ * `.fileQueueChanged` property access / `.subscribe(...)` call in `.ts`.
+ */
+
+export interface Replacement {
```
lskramarov, line 14: **This type and its driver already exist in the package.**

`utils/typescript.ts:37` exports exactly this shape:

```ts
export type Replacement = { from: string; to: string };
```

and `v20-upgrade/index.ts:19` already implements the "ordered list of regex-source replacements" driver that `applyReplacements` re-implements here. That version also carries an optional `note` per entry, used to report each replacement in dry-run mode — which is precisely the reporting this migration is missing.

With this addition there are now three local copies of the same two-field interface in `packages/schematics` (`filter-bar-signals/data.ts:35`, `v20-upgrade/data.ts:11`, and this one) while the shared type sits unused in `utils/typescript.ts`.

_Generated with Claude Code_

docs/guides/migration.en.md
```diff
@@ -955,6 +956,30 @@ Run it even if you upgrade by hand: a `{ provide: KBQ_<X>_CONFIGURATION, useValu
...
+
+```bash
+ng g @koobiq/components:file-upload-deprecated-outputs --project <your project>
+```
+
+#### What is fixed automatically
+
+Every `(fileQueueChanged)` and `(fileQueueChange)` binding is renamed to `(filesChange)` / `(fileChange)`,
+in templates and in TypeScript code (for example `.fileQueueChanged.subscribe(...)` becomes
+`.filesChange.subscribe(...)`). No manual steps are needed.
```
lskramarov: **"No manual steps are needed" overstates what the schematic can do.**

The schematic is an unscoped text replace over every `.ts`/`.html` file, so it can rewrite string literals, HTML attribute values and the consumer's own identifiers that share the name. The closing paragraph of this very guide says the opposite about regex-based migrations:

> The migration is regex-based and does not rewrite aliased imports, local variables, or re-exports — **review the diff before committing**

Section 17 is the only numbered section here that claims the migration is complete — §15 has a "What you need to fix manually" block and §16 closes by naming what surfaces as compile errors. Suggest replacing the last sentence with a note that the rewrite is textual and the diff should be reviewed, especially in projects that use these names for their own symbols.

_Generated with Claude Code_

docs/guides/migration.ru.md
```diff
@@ -959,6 +960,31 @@ ng g @koobiq/components:locale-configuration-providers --project <your project>
 переименованные типы и два сужения — проявляется ошибками компиляции, сообщения которых сами называют
 исправление.
 
+### 17. Удаление устаревших событий file-upload (20.3.0)
+
+Событие `fileQueueChanged` у множественной загрузчика файлов (`kbq-multiple-file-upload`) и `fileQueueChange`
```
lskramarov: **Grammar.** The adjective disagrees with its noun, and a singular subject takes a plural predicate:

```suggestion
События `fileQueueChanged` у множественного загрузчика файлов (`kbq-multiple-file-upload`) и `fileQueueChange`
```

_Generated with Claude Code_

lskramarov (commented):
Automated review (`/code-review xhigh`) — 11 findings on the new `file-upload-deprecated-outputs` schematic, the migration guide, and the dev app. Ranked most severe first in the inline comments.

The component/API side of the change is clean: every removed `fileQueueChanged`/`fileQueueChange` emit sat next to an identical `filesChange`/`fileChange` emit, so the guide's "pure rename, no behaviour change" claim checks out, the api.md snapshot matches, and no stale references are left in the repo. The findings are concentrated in the schematic.

_Generated with Claude Code_

docs/guides/migration.ru.md
```diff
@@ -959,6 +960,31 @@ ng g @koobiq/components:locale-configuration-providers --project <your project>
...
+
+```bash
+ng g @koobiq/components:file-upload-deprecated-outputs --project <your project>
+```
+
+#### Что исправляется автоматически
+
+Каждый `(fileQueueChanged)` и `(fileQueueChange)` переименовывается в `(filesChange)` / `(fileChange)` — как
+в шаблонах, так и в TypeScript-коде (например, `.fileQueueChanged.subscribe(...)` становится
+`.filesChange.subscribe(...)`). Ручных действий не требуется.
```
lskramarov: **Section 17 is now wedged in front of a block that belongs to §15.**

The `### После миграции` / `### After the migration` heading that follows is about dropping the `overlayscrollbars` dependency — that is the tail of the Scrollbar section (§15), not a closing note for the whole document. A reader who finishes §17 (file-upload outputs) lands straight on "npm uninstall overlayscrollbars" and reads it as a step of this migration.

The seam already existed after §16, but this PR edits exactly that boundary, so it is a good moment to fold the overlayscrollbars block back under §15 (or retitle it) in both language files.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/schema.json
```diff
@@ -0,0 +1,20 @@
+{
+    "$schema": "http://json-schema.org/schema",
+    "$id": "koobiq-components-file-upload-deprecated-outputs",
+    "title": "Koobiq components file-upload deprecated outputs removal",
+    "type": "object",
+    "properties": {
+        "project": {
+            "type": "string",
+            "description": "Name of the project to migrate. If omitted, the migration runs over the whole tree.",
```
lskramarov: **This description contradicts the `$default` right below it.**

`$default: { $source: "projectName" }` means the Angular CLI always fills `project` in with the workspace's default project, so "If omitted, the migration runs over the whole tree" is unreachable from `ng g`. Whole-tree only happens under `ng update`, which passes no options at all.

A user in a multi-project workspace running `ng g @koobiq/components:file-upload-deprecated-outputs` without `--project` will expect every project to be migrated, get only the default one, and find the rest broken at build time. (Copied from the sibling schemas, but worth fixing here rather than propagating — either drop the `$default`, or reword the description to say the default project is used.)

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
```diff
@@ -0,0 +1,65 @@
...
+        // declares no schema, so the schema default never reaches us — applying the
+        // fix is the intended behaviour there.
+        const fix = options.fix ?? true;
+        const projectDefinition = await setupOptions(project, tree);
+        const root = projectDefinition?.root ?? '';
+        const rootDir = root ? tree.getDir(root as Path) : tree.root;
+        let touched = 0;
+
+        rootDir.visit((filePath: Path, entry) => {
+            if (filePath.includes('node_modules') || filePath.includes('/dist/')) return;
```
lskramarov: **The skip filter misses generated directories, and matches substrings rather than path segments.**

Only `node_modules` and `/dist/` are excluded. Jest's HTML coverage report (`coverage/lcov-report/**.html`) embeds the highlighted source, so it contains these identifiers as text and gets rewritten; the same applies to anything left in `.angular/cache` or `out-tsc`. Those show up in the user's `git status` and inflate the `updated N file(s)` count with generated files.

In the other direction, `includes('node_modules')` is a bare substring test, so a genuine source directory whose name merely contains that string is silently skipped and never migrated. Matching on path segments (`/node_modules/`) and adding the usual generated-output directories would tighten both ends.

_Generated with Claude Code_

packages/components-dev/file-upload/template.html
```diff
@@ -159,7 +159,7 @@
         <div devCustomText class="dev-container">
             <div class="kbq-subheading">Multiple file-upload, default UX validation</div>
             <br />
-            <kbq-file-upload multiple (fileQueueChanged)="onFileQueueChange($event)">
+            <kbq-file-upload multiple (filesChange)="onFileQueueChange($event)">
```
lskramarov: The binding was renamed but the handler still carries the name of the output this PR deletes (`onFileQueueChange`, declared at `packages/components-dev/file-upload/module.ts:421`).

Renaming both (here and in `module.ts`) to something like `onFilesChange` would finish the job — otherwise `fileQueueChange` stays greppable in the repo as a live symbol after the release that removes it.

_Generated with Claude Code_

packages/schematics/src/migrations/file-upload-deprecated-outputs/data.ts
NikGurev: Мне кажется это излишним

packages/schematics/src/migrations/file-upload-deprecated-outputs/data.ts
NikGurev: Пусть каждая миграция несет отдельный тип

artembelik (approved):

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.spec.ts
NikGurev: Добавили тесты: проверяем, что похожие, но другие названия (например `onFileQueueChanged`, `fileQueueChangedAt`) не трогаются, что файлы в `node_modules` и `dist` пропускаются, и что в режиме без записи (`fix: false`) в лог выводится, какие файлы изменились бы.

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
NikGurev: Пока не стали добавлять отдельное сообщение в вывод миграции — то же предупреждение о том, что нужно проверить diff, уже добавили в гайд по обновлению (шаг 17), поэтому решили не дублировать его в консоли, чтобы не усложнять.

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
NikGurev: Поправили: теперь сначала проверяем, есть ли в файле подстрока `fileQueueChange`, и только если она есть — применяем замену. Регулярные выражения теперь собираются один раз, а не при каждом файле.

docs/guides/migration.en.md
NikGurev: Убрали фразу «ручных действий не требуется» — теперь честно написано, что замена текстовая и не привязана именно к компонентам koobiq, поэтому перед коммитом стоит посмотреть diff.

docs/guides/migration.ru.md
NikGurev: Поправили, спасибо!

docs/guides/migration.ru.md
NikGurev: Перенесли блок про удаление overlayscrollbars обратно в раздел 15, к которому он относится, чтобы после раздела 17 (про file-upload) не возникало путаницы.

packages/schematics/src/migrations/file-upload-deprecated-outputs/schema.json
NikGurev: Поправили описание — теперь честно написано, что по умолчанию берётся проект из workspace при запуске через `ng g`, а через `ng update` миграция идёт по всему дереву.

packages/schematics/src/migrations/file-upload-deprecated-outputs/index.ts
NikGurev: Поправили: теперь сравниваем не подстроку, а сегмент пути (`/node_modules/`, `/dist/`), и добавили в список пропуска ещё `coverage`, `.angular` и `out-tsc`.

packages/components-dev/file-upload/template.html
NikGurev: Переименовали обработчик в демо-приложении в `onFilesChange` — привязка в шаблоне и сам метод теперь называются одинаково.

lskramarov (approved):

PR #1957, opened 2026-08-28
feat(docs): added fullscreen mode for examples (#DS-4019)

artembelik:


rmnturov (changes requested):
Поменяй верстку кнопок-иконок, пожалуйста. Разделители не нужны, отступы изменились. Между иконками-кнопками и ссылкой разные отступы. Ссылка на Стекблитз должна быть написана тем же размером, что и «Показать код». Кстати, там надо формулировку заменить на более короткую, чтобы на мобильной версии в строку помещалось.

https://www.figma.com/design/29UkHBWsbAnWyuGOpR55m8/koobiq-%C2%B7-markdown?node-id=40174-5578&t=QTnTMzTfPGk2NeVj-4

image

apps/docs/e2e.playwright-spec.ts
```diff
@@ -85,6 +85,70 @@ test.describe('docs app', () => {
...
+
+        const viewer = page.locator('docs-live-example-viewer').first();
+        const fullscreenButton = viewer.getByRole('button', { name: 'Enter fullscreen' });
+        const fullscreenEnabled = await page.evaluate(() => document.fullscreenEnabled);
+
+        if (!fullscreenEnabled) {
+            await expect(fullscreenButton).toHaveCount(0);
+
+            return;
+        }
```
lskramarov: **Test coverage** — this guard makes the whole test pass vacuously rather than skipping.

If CI Chrome ever reports `document.fullscreenEnabled === false` (old headless mode, an enterprise/sandbox policy disabling the Fullscreen API, or a frame without `allow="fullscreen"`), the test asserts only `toHaveCount(0)` and returns green — while nothing about entering fullscreen, overlay reparenting, or the split layout is exercised. This is the feature’s only integration coverage, so the loss would be invisible.

`test.skip()` fails loudly in the reporter instead:

```suggestion
        test.skip(!fullscreenEnabled, 'Fullscreen API unavailable in this browser');
```

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.spec.ts
```diff
@@ -68,6 +108,40 @@ describe(DocsLiveExampleViewerComponent.name, () => {
         expect(toggle().textContent?.trim()).toBe('Hide example code');
     });
 
+    it('enters fullscreen mode for the whole viewer', async () => {
+        fullscreenButton().click();
+        await fixture.whenStable();
+
+        expect(requestFullscreen).toHaveBeenCalledWith({ navigationUI: 'auto' });
```
lskramarov: **Test coverage** — this assertion never checks *which* element went fullscreen.

`requestFullscreen` is mocked on `HTMLElement.prototype`, so swapping `this.host.requestFullscreen(...)` for `this.document.body.requestFullscreen(...)` (or the inner `.docs-live-example__container`) still records one call with the same argument — the test keeps passing while the `:fullscreen` styles, which key off `.docs-live-example-viewer`, silently stop applying. And `navigationUI: 'auto'` is the Fullscreen API default, so the only asserted argument carries no information.

Jest 30 is installed, so `mock.contexts` is available:

```suggestion
        expect(requestFullscreen).toHaveBeenCalledTimes(1);
        expect(requestFullscreen.mock.contexts[0]).toBe(fixture.nativeElement);
```

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -99,12 +107,38 @@ export class DocsLiveExampleViewerComponent extends DocsLocaleState {
         // Load whenever the key changes (replaces a side-effecting `@Input` setter). The effect only
         // reacts to an actual change, so `reload()` re-invokes the loader directly instead.
         effect(() => this.loadExample(this.example()));
+
+        afterNextRender(() => {
+            this.fullscreenAvailable.set(Boolean(this.document.fullscreenEnabled && this.host.requestFullscreen));
```
lskramarov: **Test coverage** — the capability guard itself is untested.

Every spec forces `document.fullscreenEnabled` to `true`, and the e2e’s unavailable branch is unreachable in CI Chrome. Weakening this expression to a constant `true` keeps the suite green — the button would then render on platforms with no element Fullscreen API (iOS Safari supports it for video only), where the click rejects and the rejection is swallowed by the `catch` in `toggleFullscreen`, leaving a control that looks live and does nothing.

A spec that defines `fullscreenEnabled` as `false` before `TestBed.createComponent` and asserts `fullscreenButton()` is null would pin it.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.html
```diff
@@ -38,6 +38,26 @@
...
             <kbq-divider class="docs-live-example-footer__divider" [vertical]="true" />
+            @if (fullscreenAvailable()) {
+                <button
+                    kbq-button
+                    color="theme"
+                    kbqStyle="transparent"
+                    kbqPlacement="bottom"
+                    [kbqTooltip]="isFullscreen() ? t('exitFullscreen') : t('enterFullscreen')"
+                    [attr.aria-label]="isFullscreen() ? t('exitFullscreen') : t('enterFullscreen')"
+                    [attr.aria-pressed]="isFullscreen()"
```
lskramarov: **Accessibility** — the button changes both its accessible name and `aria-pressed`, encoding the same state twice.

In fullscreen a screen reader announces *"Exit fullscreen, toggle button, pressed"*: the name says the button leaves fullscreen, the state says that action is already on. The WAI-ARIA APG button pattern notes that a toggle whose label already reflects its state does not need `aria-pressed`.

Either keep a stable name (`Fullscreen`) with `aria-pressed`, or keep the changing name and drop the attribute.

Flagging rather than asserting: `packages/components/form-field/password-toggle.ts:65-66` already does both, so this may be a deliberate house pattern — probably better settled explicitly than by precedent.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.spec.ts
```diff
@@ -41,9 +64,26 @@ describe(DocsLiveExampleViewerComponent.name, () => {
         fixture = TestBed.createComponent(DocsLiveExampleViewerComponent);
         httpMock = TestBed.inject(HttpTestingController);
         fixture.detectChanges();
+        await fixture.whenStable();
+        fixture.detectChanges();
     });
 
-    afterEach(() => httpMock.verify());
+    afterEach(() => {
+        httpMock.verify();
```
lskramarov: **Test hygiene** — `verify()` runs before the descriptor restore, so a failure here skips the restore entirely.

Any spec that leaves an outstanding request (e.g. a future example whose file list drifts from the two `expectOne(...)` flushes in `with a registered example`) makes this throw. The loop below never runs, `HTMLElement.prototype.requestFullscreen` and the three `document` properties stay patched for the rest of the file, and the reported failure becomes the verify error plus whatever the leaked stubs break downstream.

Restoring first — or wrapping `verify()` in `try/finally` — makes teardown unconditional.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -99,12 +107,38 @@ export class DocsLiveExampleViewerComponent extends DocsLocaleState {
         // Load whenever the key changes (replaces a side-effecting `@Input` setter). The effect only
         // reacts to an actual change, so `reload()` re-invokes the loader directly instead.
         effect(() => this.loadExample(this.example()));
+
+        afterNextRender(() => {
```
lskramarov: **Efficiency** — this probes an app-global fact once per viewer instance.

`/en/components/select/examples` renders one viewer per `<!-- example(...) -->` marker in `packages/components/select/examples.select.en.md` (~29), so the page ends up with ~29 identical `document.fullscreenEnabled` probes and ~29 document-level `fullscreenchange` listeners, each running an identity compare on every transition.

A root-provided service exposing one `fullscreenElement` signal plus a single `fullscreenAvailable` value would reduce each viewer to a `computed()` and the page to one listener.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -99,12 +107,38 @@ export class DocsLiveExampleViewerComponent extends DocsLocaleState {
...
     toggleSourceView() {
         this.isSourceShown.update((isShown) => !isShown);
     }
 
+    protected async toggleFullscreen(): Promise<void> {
+        try {
+            if (this.document.fullscreenElement === this.host) {
+                await this.document.exitFullscreen();
+            } else {
+                await this.host.requestFullscreen({ navigationUI: 'auto' });
```
lskramarov: **Simplification** — `navigationUI: 'auto'` is the Fullscreen API default, so the options object changes no behaviour.

It does give the unit test an argument to assert on, which is exactly what lets that test pass without checking which element requested fullscreen (see the comment on `docs-live-example-viewer.spec.ts:115`). Dropping it removes the dead option and forces the test to assert something real.

```suggestion
                await this.host.requestFullscreen();
```

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.scss
```diff
@@ -19,6 +19,37 @@
...
+            display: flex;
+            flex-direction: column;
+
+            & > .docs-live-example__example {
+                flex: 1 1 auto;
+            }
+        }
+    }
+
+    &::backdrop {
```
lskramarov: **Simplification** — this rule can never be seen.

`&:fullscreen` above sets `margin: 0; width: 100%; height: 100%` with an opaque `background: var(--kbq-background-bg)`, so the originating element covers its own backdrop completely in every state.

Independently: `--kbq-background-bg` is defined under `.kbq-light`/`.kbq-dark`, and `::backdrop` only inherits custom properties from its originating element in Chrome 122+ / Safari 17.4+ — on older engines the `var()` would not resolve and this would fall back to transparent anyway.

Deleting the rule changes nothing rendered.

apps/docs/src/app/config.ts
```diff
@@ -26,6 +27,7 @@ export const appConfig: ApplicationConfig = {
         provideHttpClient(withFetch()),
         provideClientHydration(withEventReplay()),
         provideAnimations(),
+        { provide: OverlayContainer, useClass: FullscreenOverlayContainer },
```
lskramarov, line 33: **Convention** — this global swap deserves a one-line rationale.

This single line is what keeps select panels, modals, sidepanels and toasts visible when a `docs-live-example-viewer` is the fullscreen element (they all route through `overlay.create()`), but nothing at the call site says so — the dependency is only asserted in an e2e file three directories away.

AGENTS.md: *"Add concise comments only to explain non-obvious rationale or constraints"*, and this component documents exactly this class of decision everywhere else (`docs-live-example-viewer.ts:107-108`, `:218-219`; `docs-live-example-viewer.scss:11-12`, `:87-90`). Without a note, a later reader trimming providers has no signal that removing this line silently breaks overlays in fullscreen examples.

lskramarov (commented):

Nine findings, none blocking: three test-coverage gaps, one accessibility question, and cleanup.

Several candidates were checked and dropped rather than reported:

- **`Escape` in the e2e does not exit fullscreen.** Verified against Chrome: a CDP-injected `Escape` reaches the page but leaves `document.fullscreenElement` set, so the select-close step is safe.
- **The tooltip is not left stranded on entering fullscreen.** `FlexibleConnectedPositionStrategy` re-applies on `ViewportRuler.change()`, which the fullscreen resize triggers.
- **`afterNextRender` does run for portal-attached viewers.** `DomPortalOutlet` is given a `viewContainerRef`, so the component joins the app view tree.
- **Signal writes from outside the zone do reach the template.** `provideZoneChangeDetection()` without `ignoreChangesOutsideZone` leaves the hybrid scheduler enabled, so exiting fullscreen via the browser’s own `Escape` still updates the button.
- **No SSR or theming break.** `FullscreenOverlayContainer` resolves no event name on the server DOM, and the moved container stays a descendant of `body.docs.koobiq`, so token inheritance holds.

_Generated with Claude Code_

lskramarov (approved):

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -99,12 +107,38 @@ export class DocsLiveExampleViewerComponent extends DocsLocaleState {
...
+            this.fullscreenAvailable.set(Boolean(this.document.fullscreenEnabled && this.host.requestFullscreen));
+
+            this.ngZone.runOutsideAngular(() => {
+                fromEvent(this.document, 'fullscreenchange')
+                    .pipe(
+                        map(() => this.document.fullscreenElement === this.host),
+                        distinctUntilChanged(),
+                        takeUntilDestroyed(this.destroyRef)
+                    )
+                    .subscribe((isFullscreen) => this.isFullscreen.set(isFullscreen));
```
NikGurev: Можем ли мы в toggleFullscreen переключать isFullscreen для простоты?

apps/docs/e2e.playwright-spec.ts
artembelik: поправил

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.spec.ts
artembelik: поправил

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.html
artembelik: поправил

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
artembelik: поправил

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.scss
artembelik: справедливо, поправил

apps/docs/src/app/config.ts
artembelik: добавил комментарий

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
artembelik: нет, изменил реализацию, вынес логику в сервис и добавил метод toggle

artembelik:
> Поменяй верстку кнопок-иконок, пожалуйста. Разделители не нужны, отступы изменились. Между иконками-кнопками и ссылкой разные отступы. Ссылка на Стекблитз должна быть написана тем же размером, что и «Показать код». Кстати, там надо формулировку заменить на более короткую, чтобы на мобильной версии в строку помещалось.

поправил

NikGurev (approved):

apps/docs/src/app/services/i18n.ts
```diff
@@ -76,9 +76,11 @@ export const DOCS_TRANSLATIONS = {
 
     // example / live-example viewers
     loadingDocument: { ru: 'Загрузка документа...', en: 'Loading document...' },
-    showExampleCode: { ru: 'Показать пример кода', en: 'Show example code' },
-    hideExampleCode: { ru: 'Скрыть пример кода', en: 'Hide example code' },
+    showExampleCode: { ru: 'Показать код', en: 'Show code' },
+    hideExampleCode: { ru: 'Скрыть код', en: 'Hide code' },
     resetState: { ru: 'Сбросить состояние', en: 'Reset state' },
+    enterFullscreen: { ru: 'На весь экран', en: 'Enter fullscreen' },
+    exitFullscreen: { ru: 'Выйти из полного экрана', en: 'Exit fullscreen' },
```
rmnturov:
```suggestion
    exitFullscreen: { ru: 'Выйти из полноэкранного режима', en: 'Exit fullscreen' },
```

rmnturov: «Из полного экрана» — это была будто калька с английского

rmnturov: Пояснение к suggestion:

Предлагаю писать `full screen` раздельно в пользовательских текстах. Такое написание используют Apple (`Full Screen`, `Exit Full Screen`), YouTube (`Full screen`, `Exit full screen`), Firefox (`Full Screen mode`, `exit full screen`) и Chrome (`Full screen`, `full-screen mode`).

[Apple](https://support.apple.com/en-ie/guide/mac-help/mchl9c21d2be/mac), [YouTube](https://support.google.com/youtube/answer/72689?hl=en), [Firefox](https://support.mozilla.org/en-US/kb/how-make-firefox-and-websites-go-full-screen), [Chrome](https://support.google.com/chrome/answer/96810).

Слитная форма `fullscreen` характерна прежде всего для технических названий: `Fullscreen API`, `requestFullscreen()`, `fullscreenElement` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API)). Поэтому для тултипов и `aria-label` предлагаю использовать `Enter full screen` и `Exit full screen`.

В блоке `suggestion` также учтена согласованная русская правка: GitHub заменяет строку целиком.

```suggestion
exitFullscreen: { ru: 'Выйти из полноэкранного режима', en: 'Exit full screen' },
```

rmnturov (approved):

apps/docs/src/app/services/i18n.ts
artembelik: поправил

artembelik: поправил

PR #1958, opened 2026-08-28
feat(cli): expose @koobiq/cli release helpers (#DS-5212)

NikGurev:
expose changelog version diff extraction to be used separately

Planning to use in @koobiq/icons and @koobiq/data-grid in conjuction with NX

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
         "build:components-experimental": "ng build components-experimental",
         "build:docs-examples-module": "ts-node --project ./tools/example-module/tsconfig.json ./tools/example-module/index.ts",
         "build:docs-examples": "ng build docs-examples",
-        "build:cli": "rimraf dist/cli && rollup -c ./packages/cli/rollup.config.js && tsc -p packages/cli/tsconfig.release.json && node ./packages/cli/scripts/copy-meta-to-dist.js",
+        "build:cli": "rimraf dist/cli packages/cli/dist/release && rollup -c ./packages/cli/rollup.config.js && tsc -p packages/cli/tsconfig.release.json && node ./packages/cli/scripts/copy-meta-to-dist.js",
         "build:tokens:data": "node ./tools/tokens/docs/index.js",
         "build:docs-content": "ts-node --project ./tools/markdown-to-html/tsconfig.json ./tools/markdown-to-html/transform-markdown.ts && yarn run docs:api-gen",
         "build:package-docs-content": "ts-node --project ./tools/package-docs-content/tsconfig.json ./tools/package-docs-content/package-docs-content.ts",
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
         "build:components-experimental": "ng build components-experimental",
         "build:docs-examples-module": "ts-node --project ./tools/example-module/tsconfig.json ./tools/example-module/index.ts",
         "build:docs-examples": "ng build docs-examples",
-        "build:cli": "rimraf dist/cli && rollup -c ./packages/cli/rollup.config.js && node ./packages/cli/scripts/copy-meta-to-dist.js",
+        "build:cli": "rimraf dist/cli && rollup -c ./packages/cli/rollup.config.js && tsc -p packages/cli/tsconfig.release.json && node ./packages/cli/scripts/copy-meta-to-dist.js",
         "build:tokens:data": "node ./tools/tokens/docs/index.js",
         "build:docs-content": "ts-node --project ./tools/markdown-to-html/tsconfig.json ./tools/markdown-to-html/transform-markdown.ts && yarn run docs:api-gen",
         "build:package-docs-content": "ts-node --project ./tools/package-docs-content/tsconfig.json ./tools/package-docs-content/package-docs-content.ts",
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -22,6 +22,19 @@ npx koobiq-cli stage-commit
 npx koobiq-cli publish-ci-gitlab
 ```
 
+## Reading release notes in your own scripts
+
+If you just want the release notes for a version — for example, to post a release announcement
+somewhere — you don't need the full CLI.
+
+`@koobiq/cli/release` reads them straight out of your project's `CHANGELOG.md` file:
+
+```ts
+import { extractReleaseNotes } from '@koobiq/cli/release';
+
+const notes = extractReleaseNotes('./CHANGELOG.md', '1.2.3');
+```
+
 ## Use npm pack to test packages locally
 
 Build Package
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -27,13 +27,37 @@ npx koobiq-cli publish-ci-gitlab
 If you just want the release notes for a version — for example, to post a release announcement
 somewhere — you don't need the full CLI.
 
-`@koobiq/cli/release` reads them straight out of your project's `CHANGELOG.md` file:
+`@koobiq/cli/release` reads them straight out of your project's `CHANGELOG.md` file. It throws if
+the file doesn't exist, and returns `null` if the version isn't in it, so guard both:
 
 ```ts
 import { extractReleaseNotes } from '@koobiq/cli/release';
 
 const notes = extractReleaseNotes('./CHANGELOG.md', '1.2.3');
+
+if (!notes) {
+    throw new Error('Version 1.2.3 was not found in the changelog');
+}
+
+console.log(notes.releaseTitle, notes.releaseNotes);
 ```
+
+### Resolving a changelog from a git tag
+
+`parseTag` and `resolveChangelogPath` turn a release tag into the right changelog path. They
+support both a plain `{version}` tag and Nx's independent-release `{projectName}@{version}` tag,
+so the same code works for single-project and per-package releases:
+
+```ts
+import { extractReleaseNotes, parseTag, resolveChangelogPath } from '@koobiq/cli/release';
+
+const tag = process.env.GIT_TAG!; // e.g. '1.2.3' or '@scope/my-package@1.2.3'
+const changelogPath = resolveChangelogPath(process.cwd(), parseTag(tag));
+const notes = extractReleaseNotes(changelogPath, parseTag(tag).version);
+```
+
+`resolveChangelogPath` throws if the tag names a project but no `packages/{project}/CHANGELOG.md`
+exists for it.
 
 ## Use npm pack to test packages locally
 
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -12,7 +12,8 @@
         "./release": {
             "types": "./dist/release/index.d.ts",
             "default": "./dist/release/index.js"
-        }
+        },
+        "./package.json": "./package.json"
     },
     "engines": {
         "node": ">=24.16",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -7,6 +7,13 @@
         "koobiq-cli": "./bin/cli"
     },
     "main": "./dist/cli.js",
+    "exports": {
+        ".": "./dist/cli.js",
+        "./release": {
+            "types": "./dist/release/index.d.ts",
+            "default": "./dist/release/index.js"
+        }
+    },
     "engines": {
         "node": ">=24.16",
         "npm": ">=11",
--- a/packages/cli/scripts/copy-meta-to-dist.js
+++ b/packages/cli/scripts/copy-meta-to-dist.js
@@ -21,7 +21,7 @@ copyFileSync(getSrcPckgPath('LICENSE'), getDistPath(`${distCLIPath}/LICENSE`));
 copyFileSync(getSrcPckgPath('bin/cli'), getDistPath(`${distCLIPath}/bin/cli`));
 copyFileSync(getSrcPckgPath('dist/cli.js'), getDistPath(`${distCLIPath}/dist/cli.js`));
 copyFileSync(getSrcPckgPath('dist/cli.js.map'), getDistPath(`${distCLIPath}/dist/cli.js.map`));
-cpSync(getSrcPckgPath('dist/release'), getDistPath(`${distCLIPath}/release`), { recursive: true });
+cpSync(getSrcPckgPath('dist/release'), getDistPath(`${distCLIPath}/dist/release`), { recursive: true });
 
 // Update package version
 const currentVersion = JSON.parse(readFileSync('package.json', 'utf8')).version;
--- a/packages/cli/scripts/copy-meta-to-dist.js
+++ b/packages/cli/scripts/copy-meta-to-dist.js
@@ -1,4 +1,4 @@
-const { copyFileSync, existsSync, mkdirSync, readFileSync, writeFileSync } = require('fs');
+const { copyFileSync, cpSync, existsSync, mkdirSync, readFileSync, writeFileSync } = require('fs');
 const { resolve } = require('path');
 
 const getSrcPckgPath = (root) => resolve(__dirname, '..', root);
@@ -21,6 +21,7 @@ copyFileSync(getSrcPckgPath('LICENSE'), getDistPath(`${distCLIPath}/LICENSE`));
 copyFileSync(getSrcPckgPath('bin/cli'), getDistPath(`${distCLIPath}/bin/cli`));
 copyFileSync(getSrcPckgPath('dist/cli.js'), getDistPath(`${distCLIPath}/dist/cli.js`));
 copyFileSync(getSrcPckgPath('dist/cli.js.map'), getDistPath(`${distCLIPath}/dist/cli.js.map`));
+cpSync(getSrcPckgPath('dist/release'), getDistPath(`${distCLIPath}/release`), { recursive: true });
 
 // Update package version
 const currentVersion = JSON.parse(readFileSync('package.json', 'utf8')).version;
--- a/packages/cli/src/release/extract-release-notes.spec.ts
+++ b/packages/cli/src/release/extract-release-notes.spec.ts
@@ -1,5 +1,7 @@
-import fs from 'fs';
-import { extractReleaseNotes } from './extract-release-notes';
+import fs, { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { extractReleaseNotes, parseTag, resolveChangelogPath } from './extract-release-notes';
 
 describe(extractReleaseNotes.name, () => {
     it('should extract release notes for a version between two other versions', () => {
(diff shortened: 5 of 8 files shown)


packages/cli/scripts/copy-meta-to-dist.js
```diff
@@ -21,6 +21,7 @@ copyFileSync(getSrcPckgPath('LICENSE'), getDistPath(`${distCLIPath}/LICENSE`));
 copyFileSync(getSrcPckgPath('bin/cli'), getDistPath(`${distCLIPath}/bin/cli`));
 copyFileSync(getSrcPckgPath('dist/cli.js'), getDistPath(`${distCLIPath}/dist/cli.js`));
 copyFileSync(getSrcPckgPath('dist/cli.js.map'), getDistPath(`${distCLIPath}/dist/cli.js.map`));
+cpSync(getSrcPckgPath('dist/release'), getDistPath(`${distCLIPath}/release`), { recursive: true });
```
lskramarov: **The published `@koobiq/cli/release` entry point does not resolve.**

This copies the compiled output to `<pkg>/release`, but `packages/cli/package.json` declares the subpath as `"./dist/release/index.js"`. Every other artifact in this script goes to `${distCLIPath}/dist/...` — this one drops the `dist/` segment.

Reproduced against a fixture package mirroring the shipped layout:

```
require('@koobiq/cli/release')  ->  MODULE_NOT_FOUND
  Cannot find module .../@koobiq/cli/dist/release/index.js
```

Copying to `dist/release` instead resolves OK. As it stands the feature this PR adds is dead in the published package — including the README example added in this same diff.

```suggestion
cpSync(getSrcPckgPath('dist/release'), getDistPath(`${distCLIPath}/dist/release`), { recursive: true });
```

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.ts
```diff
@@ -48,3 +50,53 @@ export function extractReleaseNotes(changelogPath: string, versionName: string):
         releaseTitle
     } satisfies ChangelogReleaseNotes;
 }
```
lskramarov, line 62: **Substring version matching merges notes from several releases** (about line 34, outside this diff — flagging it because this PR turns the function into a documented public API).

Line 34 matches with `line.includes(versionName)`, and because that branch `continue`s, the `break` on line 39 never fires when a *later* heading also matches. The result is a title from the wrong release and notes concatenated across all of them.

Verified by running the loop against a changelog holding a release and its release candidate:

```ts
extractReleaseNotes(path, '20.2.0')
// title : "# 20.2.0-rc.1 (2026-07-20)"      <- wrong release
// notes : final release entries + rc-only entries, merged
```

Same for `'1.2.3'` against `# 1.2.30`, and `'8.10.0'` against `# 18.10.0`. Prerelease labels are already a supported flow here (`src/release/prompt/prerelease-labels.ts`), and the five internal callers (`publish-release-*.ts`, `stage-release-commit.ts`) are exposed too.

Matching the version against the parsed heading rather than by substring — or at minimum anchoring on a word boundary and `break`ing before re-matching — would fix it. Worth doing in this PR, since README:35 documents exactly the `extractReleaseNotes(path, '1.2.3')` shape.

_Generated with Claude Code_

lskramarov (commented):
## Code review — `max` effort

13 findings, most severe first. Two are blocking in my view:

1. **The published `@koobiq/cli/release` subpath does not resolve** — the build copies to `<pkg>/release` while `exports` points at `dist/release/index.js`. Verified: `MODULE_NOT_FOUND`. The feature is dead in the packed artifact, README example included.
2. **`extractReleaseNotes` matches versions by substring**, so a release and its `-rc` (or `1.2.3` vs `1.2.30`) collide, the `break` guard never fires, and notes from several releases come back merged under the wrong title. Pre-existing, but this PR makes it a documented public API and the repo's own release pipeline already runs through it.

The rest cover the scoped-tag path mismatch, the `exports` map shadowing `package.json`, `parseTag` validation, and build/test/doc cleanups. Every claim below was reproduced locally rather than read off the diff.

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.ts
```diff
@@ -48,3 +50,53 @@ export function extractReleaseNotes(changelogPath: string, versionName: string):
...
+ * (`packages/{project}/CHANGELOG.md`) for a scoped tag, falling back to the workspace root
+ * changelog when there's no project scope, or no changelog exists for that project.
+ *
+ * @param workspaceRoot
+ * @param parsedTag
+ * @see parseTag
+ */
+export function resolveChangelogPath(workspaceRoot: string, { project }: ParsedTag): string {
+    if (project) {
+        const projectChangelog = join(workspaceRoot, 'packages', project, CHANGELOG_FILE_NAME);
```
lskramarov: **Scoped project names resolve to a path that can never exist.**

The spec at line 103 asserts that `parseTag` keeps scoped npm names intact (`'should split on the last @ so a scoped npm package name still works'`), but feeding that result here produces a directory no monorepo has:

```ts
resolveChangelogPath(root, parseTag('@koobiq/components@20.2.0'))
// builds <root>/packages/@koobiq/components/CHANGELOG.md   <- never exists
// real directory is <root>/packages/components
```

`existsSync` then fails and the function silently returns the **root** changelog, so the caller extracts and publishes notes for the wrong project with no error at all.

Stripping the npm scope (`project.replace(/^@[^/]+\//, '')`) before joining, or taking an explicit project→directory mapping, would honour what `parseTag` already promises.

_Generated with Claude Code_

packages/cli/package.json
```diff
@@ -7,6 +7,13 @@
         "koobiq-cli": "./bin/cli"
     },
     "main": "./dist/cli.js",
+    "exports": {
+        ".": "./dist/cli.js",
+        "./release": {
+            "types": "./dist/release/index.d.ts",
+            "default": "./dist/release/index.js"
+        }
+    },
```
lskramarov, line 17: **Adding `exports` silently breaks `require('@koobiq/cli/package.json')`.**

There was no `exports` field before this diff, so consumers could read the manifest. Now:

```
require('@koobiq/cli/package.json')  ->  ERR_PACKAGE_PATH_NOT_EXPORTED
```

(verified against a fixture). Version-check scripts, bundler plugins and dependency tooling that read the manifest break on upgrade — an unannounced breaking change in a PR whose stated scope is adding one helper entry point.

```suggestion
    "exports": {
        ".": "./dist/cli.js",
        "./release": {
            "types": "./dist/release/index.d.ts",
            "default": "./dist/release/index.js"
        },
        "./package.json": "./package.json"
    },
```

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.ts
```diff
@@ -48,3 +50,53 @@ export function extractReleaseNotes(changelogPath: string, versionName: string):
...
+
+        if (existsSync(projectChangelog)) {
+            return projectChangelog;
+        }
+    }
+
+    return join(workspaceRoot, CHANGELOG_FILE_NAME);
```
lskramarov: **The root fallback hides a failed project lookup.**

Two different situations end up here: `project === null`, where the root changelog is genuinely the right answer, and a project scope that failed to resolve — a typo, a scoped name (see the comment on line 94), or a monorepo that doesn't use a `packages/<name>` layout. The second is a misconfiguration, and returning a valid-looking path for it means the caller publishes another project's release notes without any signal that something went wrong. The spec at line 137 currently pins that as intended behaviour.

Worth separating the two: fall back only when there is no project scope, and surface a failed lookup (throw, or return `null`). Note that no package in this repo has its own `CHANGELOG.md`, so the fallback is the only branch this repo ever exercises — the project branch is effectively untested in practice.

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.ts
```diff
@@ -48,3 +50,53 @@ export function extractReleaseNotes(changelogPath: string, versionName: string):
...
+ *
+ * Supports both a plain `{version}` tag (single-project or fixed-group releases) and a
+ * `{projectName}@{version}` tag (Nx's independent-release tag pattern, e.g.
+ * `ag-grid-angular-theme@34.5.1`), so the same tag can drive changelog extraction across
+ * monorepos using either convention.
+ *
+ * @param tag
+ */
+export function parseTag(tag: string): ParsedTag {
+    const at = tag.lastIndexOf('@');
```
lskramarov, line 84: **`parseTag` accepts anything, and two of its outputs are unusable.**

```ts
parseTag('@koobiq/components')  // { project: '', version: 'koobiq/components' }
parseTag('v20.2.0')             // { project: null, version: 'v20.2.0' }
```

1. The empty-string project contradicts the documented `string | null` contract. A consumer branching on `project === null` to detect an unscoped tag takes the wrong path, while `resolveChangelogPath`'s `if (project)` reads the same value as unscoped — two readings of one value.
2. `v`-prefixed versions pass straight through, and `extractReleaseNotes` can never match `v20.2.0` against a `# 20.2.0 (…)` heading — it just returns `null`. The `v` prefix is the dominant git tag convention for the plain `{version}` form the JSDoc says it supports.

Returning `null` for the project when the split yields an empty string, and tolerating an optional `v` prefix on the version, would make the documented contract hold.

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.ts
```diff
@@ -48,3 +50,53 @@ export function extractReleaseNotes(changelogPath: string, versionName: string):
...
+ *
+ * @param workspaceRoot
+ * @param parsedTag
+ * @see parseTag
+ */
+export function resolveChangelogPath(workspaceRoot: string, { project }: ParsedTag): string {
+    if (project) {
+        const projectChangelog = join(workspaceRoot, 'packages', project, CHANGELOG_FILE_NAME);
+
+        if (existsSync(projectChangelog)) {
```
lskramarov: **The `project` segment is joined into a path with no validation, so `..` escapes `workspaceRoot`.**

```ts
resolveChangelogPath('/ws', parseTag('../../../../etc/passwd@1.0.0'))
// -> '/etc/passwd/CHANGELOG.md'   (outside the workspace)
```

Low severity while tags come from your own git history, but this is now a published API that third parties call with tag strings taken from CI environment variables or webhook payloads. Rejecting a `project` that is not a plain path segment, or asserting the resolved path stays under `workspaceRoot`, closes it cheaply.

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.spec.ts
```diff
@@ -85,3 +87,56 @@ source code didn't change.
...
+
+    it('should split on the last @ so a scoped npm package name still works', () => {
+        expect(parseTag('@koobiq/components@18.10.0')).toEqual({
+            project: '@koobiq/components',
+            version: '18.10.0'
+        });
+    });
+});
+
+describe(resolveChangelogPath.name, () => {
```
lskramarov, line 150: **This suite runs after a describe that permanently replaces `fs.readFileSync`.**

Line 66 does `fs.readFileSync = jest.fn(() => CHANGELOG_CONTENT) as any` — a direct assignment to the module export, not `jest.spyOn`, so nothing ever restores it and the stub stays live for the rest of the file.

Nothing fails today (all 44 tests pass — I ran `npx jest --roots packages/cli`) because this new suite happens not to call `readFileSync`. But the diff lands right on the edge of the trap: the next test appended to this file that reads a real changelog will silently receive the hard-coded `CHANGELOG_CONTENT` and assert against the wrong data.

Switching line 66 to `jest.spyOn(fs, 'readFileSync').mockReturnValue(CHANGELOG_CONTENT)` with a `mockRestore` in `afterEach` removes the hazard.

_Generated with Claude Code_

packages/cli/src/release/extract-release-notes.spec.ts
```diff
@@ -85,3 +87,56 @@ source code didn't change.
...
+    });
+
+    it('should resolve to the root changelog for an unscoped tag', () => {
+        const path = resolveChangelogPath(workspaceRoot, { project: null, version: '18.10.0' });
+
+        expect(path).toBe(join(workspaceRoot, 'CHANGELOG.md'));
+    });
+
+    it('should resolve to the project changelog when it exists', () => {
+        const path = resolveChangelogPath(workspaceRoot, { project: 'ag-grid-angular-theme', version: '34.5.1' });
```
lskramarov, line 171: **Nothing composes `parseTag` with `resolveChangelogPath`** — the exact pairing the `@see parseTag` JSDoc describes, and the only way these two are meant to be used together.

Every case here passes a hand-built literal that sidesteps `parseTag`, which is why the scoped-name mismatch is invisible. One test composing them fails immediately and exposes it:

```ts
it('should resolve a scoped project tag to its package changelog', () => {
    const path = resolveChangelogPath(workspaceRoot, parseTag('@scope/components@20.2.0'));

    expect(path).toBe(join(workspaceRoot, 'packages', 'components', 'CHANGELOG.md'));
});
```

_Generated with Claude Code_

packages/cli/tsconfig.release.json
```diff
@@ -0,0 +1,16 @@
+{
```
lskramarov, line 1: **This duplicates `packages/cli/tsconfig.json` instead of extending it, and already diverges.**

Both configs now compile the same `src/release/*.ts` sources under different settings: `target` ES2017 vs ES2022, `moduleResolution` `bundler` vs `node`, `sourceMap` false vs true, and `composite` silently dropped.

Any future change to the base config — a new strictness flag, a lib bump, a target move — applies to the rollup bundle but not to the published `./release` entry, so two builds of identical code drift apart with nothing to catch it. `extends` keeps them in step:

```json
{
    "extends": "./tsconfig.json",
    "compilerOptions": {
        "module": "commonjs",
        "declaration": true,
        "sourceMap": true,
        "composite": false,
        "rootDir": "src/release",
        "outDir": "dist/release"
    },
    "include": ["src/release/index.ts"]
}
```

_Generated with Claude Code_

package.json
```diff
@@ -175,7 +175,7 @@
         "build:components-experimental": "ng build components-experimental",
         "build:docs-examples-module": "ts-node --project ./tools/example-module/tsconfig.json ./tools/example-module/index.ts",
         "build:docs-examples": "ng build docs-examples",
-        "build:cli": "rimraf dist/cli && rollup -c ./packages/cli/rollup.config.js && node ./packages/cli/scripts/copy-meta-to-dist.js",
+        "build:cli": "rimraf dist/cli && rollup -c ./packages/cli/rollup.config.js && tsc -p packages/cli/tsconfig.release.json && node ./packages/cli/scripts/copy-meta-to-dist.js",
```
lskramarov: **The new `tsc` output directory is never cleaned.**

`rimraf dist/cli` removes the *destination* tree, but `tsc -p packages/cli/tsconfig.release.json` writes to `packages/cli/dist/release`, which nothing cleans and which `tsc` never prunes.

Rename or delete a file under `src/release/` and its orphaned `.js` / `.d.ts` / `.js.map` stay there indefinitely — after which the recursive `cpSync` in `copy-meta-to-dist.js` copies them into the published package, shipping dead modules and stale type declarations that consumers can still import.

```suggestion
        "build:cli": "rimraf dist/cli packages/cli/dist/release && rollup -c ./packages/cli/rollup.config.js && tsc -p packages/cli/tsconfig.release.json && node ./packages/cli/scripts/copy-meta-to-dist.js",
```

_Generated with Claude Code_

packages/cli/README.md
```diff
@@ -22,6 +22,19 @@ npx koobiq-cli stage-commit
...
+
+If you just want the release notes for a version — for example, to post a release announcement
+somewhere — you don't need the full CLI.
+
+`@koobiq/cli/release` reads them straight out of your project's `CHANGELOG.md` file:
+
+```ts
+import { extractReleaseNotes } from '@koobiq/cli/release';
+
+const notes = extractReleaseNotes('./CHANGELOG.md', '1.2.3');
```
lskramarov, line 36: **The example omits both failure modes, and three of the four exported helpers are undocumented.**

`extractReleaseNotes` returns `ChangelogReleaseNotes | null` (line 46 of the source) and lets `readFileSync` throw when the path is wrong. A reader who copies this line and dereferences `notes.releaseNotes` gets a `TypeError` whenever the version is not in the changelog, and an uncaught `ENOENT` when the path is off.

Separately, `parseTag`, `resolveChangelogPath` and `isVersionLine` are exported by `src/release/index.ts` but never mentioned here — the monorepo tag flow is the actual substance of this PR and of its title, and it ships undocumented. Worth showing the tag → path → notes chain and the `null` guard.

_Generated with Claude Code_

packages/cli/src/release/index.ts
```diff
@@ -0,0 +1,8 @@
+export {
+    extractReleaseNotes,
+    isVersionLine,
```
lskramarov: **Is `isVersionLine` meant to be public?**

It is an internal predicate encoding one changelog heading layout (`X.Y.Z … (YYYY-MM-DD)`) with no stated use case outside `extractReleaseNotes`'s own loop. Publishing it means any correction to heading detection — including the substring/prerelease fix needed on line 34 — becomes a breaking change for anyone who depends on it.

The README and the PR title describe `extractReleaseNotes`, `parseTag` and `resolveChangelogPath`. Keeping `isVersionLine` module-private leaves room to fix the matching bug without a major bump.

_Generated with Claude Code_

lskramarov (approved):

packages/cli/scripts/copy-meta-to-dist.js
NikGurev: Да, тут была ошибка — файлы копировались не в ту папку, и новая возможность просто не заработала бы у пользователей пакета. Поправил путь и пересобрал пакет — теперь всё подключается как надо.

packages/cli/src/release/extract-release-notes.ts
NikGurev: Согласен, баг настоящий — при частичном совпадении номера версии в заметки могли попасть куски от других релизов. Поправил сравнение так, чтобы номер версии совпадал полностью, и добавил тест на такой случай.

packages/cli/src/release/extract-release-notes.ts
NikGurev: Верно подмечено. Для пакетов со scope (вида @scope/имя) путь к файлу с заметками строился неправильно. Убрал scope перед поиском папки пакета и добавил тест.

packages/cli/package.json
NikGurev: Согласен, это могло сломать сторонние скрипты и инструменты, которые читают package.json. Добавил его в список доступных путей.

packages/cli/src/release/extract-release-notes.ts
NikGurev: Согласен, раньше при ошибке в названии пакета мы молча показывали заметки не от того релиза. Теперь в такой ситуации будет понятная ошибка вместо подмены.

packages/cli/src/release/extract-release-notes.ts
NikGurev: Про пустое имя пакета согласен, поправил — теперь в этом случае возвращается null, как и написано в описании функции. Версии с буквой v в начале не стал трогать: сейчас функция и не обещает такое понимать, это уже отдельная возможность, ее стоит обсудить отдельно, если она правда нужна.

packages/cli/src/release/extract-release-notes.ts
NikGurev: Согласен, это стоило закрыть. Добавил проверку — имя пакета должно быть просто именем, без ../ и подобных вещей, иначе будет ошибка.

packages/cli/src/release/extract-release-notes.spec.ts
NikGurev: Точно, подмена могла остаться и повлиять на другие тесты. Заменил на аккуратный вариант с восстановлением после каждого теста.

packages/cli/src/release/extract-release-notes.spec.ts
NikGurev: Добавил тест, который проверяет именно связку этих двух функций вместе, как вы и предложили.

packages/cli/tsconfig.release.json
NikGurev: Согласен, дублировать настройки — плохая идея, из-за этого сборки могли постепенно разъехаться. Сделал через наследование от основного tsconfig.

package.json
NikGurev: Верно, папка сборки не очищалась перед пересборкой. Добавил ее в очистку.

packages/cli/README.md
NikGurev: Дополнил README — добавил про возможные ошибки при использовании и описал остальные функции, которые раньше не были упомянуты.

packages/cli/src/release/index.ts
NikGurev: Согласен, это внутренняя деталь, наружу ее отдавать незачем. Убрал из списка публичных функций.

PR #1964, opened 2026-08-31
feat(cli): add dist tag policy resolution on publish (#DS-4781)

NikGurev:
## Summary

Добавлен единый механизм разрешения публикуемого тега.

Как работает: 
Если больше или равно текущей версии - тегу выставляется latest
В других случаях `v${version}-lts`

Надо применить в:
- [x] @koobiq/angular-components
- [x] @koobiq/icons (https://github.com/koobiq/icons/pull/147) - влить после выпуска версии @koobiq/cli
- [ ] ~~@koobiq/design-tokens~~ - первоначально не нужно,не часто версии меняются
- [ ] ~~@koobiq/data-grid~~


packages/cli/src/release/npm/npm-client.ts
```diff
@@ -52,6 +52,22 @@ export function npmPublish(packagePath: string, distTag: string): string | undef
...
+/** Returns the version currently published under the given dist-tag, or null if none exists. */
+export function npmViewDistTag(packageName: string, tag: string): string | null {
+    const result = spawnSync('npm', ['view', packageName, `dist-tags.${tag}`], {
+        shell: true,
+        env: npmClientEnvironment
+    });
+
+    if (result.status !== 0) {
+        return null;
+    }
```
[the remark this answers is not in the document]
NikGurev: Добавил проверку E404 ошибки, на остальное выбрасываю ошибку

packages/cli/src/release/publish-release-github-ci.ts
```diff
@@ -81,6 +82,20 @@ export class PublishReleaseCIGithubTask extends BaseReleaseTask {
...
+    /**
+     * Resolves the npm dist-tag for every package before publishing any of them, so a failure
+     * resolving one package's tag (e.g. an ambiguous npm view error) aborts before anything has
+     * been published, instead of leaving a partial release across packages.
+     */
+    private resolveNpmDistTags(): Map<string, string> {
+        return new Map(
+            this.packageJson.release.packages.map((packageName: string) => [
+                packageName,
+                this.config.tagName ?? resolveNpmDistTag(packageName, this.currentVersion)
```
lskramarov: **Blocking — `release.packages` holds dist *directory* names, not npm package names, so the whole policy is a no-op here.**

The root `package.json` declares:

```json
"release": { "packages": ["components", "components-experimental", "angular-luxon-adapter", "angular-moment-adapter", "cli"] }
```

These are directories under `dist/` — that is how they are used one method below (`npmPublish(join(this.releaseOutputPath, packageName), ...)`). The published names are `@koobiq/components`, `@koobiq/cli`, and so on. `resolveNpmDistTag` treats the same string as an npm package name and runs `npm view <name> dist-tags.latest`.

Checked against the live registry:

| queried | resolves to |
| --- | --- |
| `components` | `0.1.0` — an unrelated public package |
| `cli` | `1.0.1` — an unrelated public package |
| `components-experimental`, `angular-luxon-adapter`, `angular-moment-adapter` | `E404` → `null` |

Consequences:

1. The LTS guard never fires. Publishing `19.9.1` from the 19.x branch after 20.x shipped compares `19.9.1` against `0.1.0`, so `compareVersions(...) >= 0` and the tag resolves to `latest` — exactly the regression this PR exists to prevent.
2. koobiq’s dist-tag now depends on third-party packages. If someone publishes `components@21.0.0`, the next koobiq `20.x` release resolves to `v20-lts` and `latest` stops moving.

The same line appears in `publish-release-ci.ts:111`, `publish-release-from-dist.ts:92` and `publish-release-gitlab-ci.ts:96`. The dist package name is available at `dist/<dir>/package.json` → `.name`, or a `dir → npm name` map could live next to the `release` config.

Note that `resolve-publish-tag.spec.ts` passes `'@koobiq/icons'`, which encodes the (correct) assumption that callers hand over npm names — the publish tasks are the ones that break it.

packages/cli/src/release/resolve-npm-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+    projectDir: string;
+    /** npm package name to check the currently published `latest` against. */
+    packageName?: string;
+}
+
+/**
+ * Standalone command wrapping `resolveNpmDistTag` so any koobiq repo's CI can resolve the
+ * correct npm dist-tag (`latest` vs `v<major>-lts`) without reimplementing the semver-against-
+ * registry comparison itself. Prints only the resolved tag to stdout, so callers can do
+ * `TAG=$(koobiq-cli resolve-npm-tag --package-name @koobiq/icons)`.
```
lskramarov, line 20: **Blocking — `dotenv` writes to stdout, so the documented `$(...)` capture is broken under the version this package declares.**

`packages/cli/package.json` declares `"dotenv": "^17.4.2"`, and `cli.ts` calls `dotenvConfig()` before any subcommand runs. Installed `dotenv@17.4.2` and captured stdout only:

```
$ OUT=$(node -e "require('dotenv').config()" 2>/dev/null); echo "[$OUT]"
[◇ injected env (1) from .env // tip: ⌘ override existing { override: true }]

# and with no .env file present at all:
[◇ injected env (0) from .env // tip: ⌘ custom filepath { path: '/custom/path/.env' }]
```

So in `@koobiq/icons`, `TAG=$(koobiq-cli resolve-npm-tag --package-name @koobiq/icons)` captures the banner **and** the tag, and `npm publish --tag "$TAG"` gets garbage.

This repo happens to resolve `dotenv@16.6.1` (hoisted, silent), so its own CI would never surface it — but a standalone install of `@koobiq/cli` gets 17.x.

Fix is one option in `cli.ts`:

```ts
dotenvConfig({ quiet: true });
```

packages/cli/src/release/npm/resolve-publish-tag.ts
```diff
@@ -0,0 +1,38 @@
+import { Version, parseVersionName } from '../version-name/parse-version';
+import { npmViewDistTag } from './npm-client';
+
+/** Positive when `a` outranks `b`, negative when `b` outranks `a`, zero when equal. */
+function compareVersions(a: Version, b: Version): number {
+    if (a.major !== b.major) return a.major - b.major;
+    if (a.minor !== b.minor) return a.minor - b.minor;
+    if (a.patch !== b.patch) return a.patch - b.patch;
+    // A prerelease always ranks below its own stable release.
+    if (!a.prereleaseLabel !== !b.prereleaseLabel) return a.prereleaseLabel ? -1 : 1;
```
lskramarov: **`compareVersions` never compares the prerelease label, only the number.**

Line 10 establishes “a prerelease ranks below its stable release”, but once *both* sides have a label the function falls straight through to the numeric comparison, so `alpha`, `beta` and `rc` rank as equal. Ran the function as written against `parse-version.ts`:

| published `latest` | publishing | resolved | should be |
| --- | --- | --- | --- |
| `20.3.0-beta.5` | `20.3.0-rc.0` | `v20-lts` | `latest` |
| `20.3.0-alpha.9` | `20.3.0-beta.0` | `v20-lts` | `latest` |
| `20.3.0-rc.0` | `20.3.0-alpha.0` | `latest` | `v20-lts` |

The third row is the damaging one: `latest` silently moves **back** onto an older prerelease.

Prereleases really do reach this path — the `publish.yml` tag filter `*.*.*` matches `20.3.0-alpha.0`, and `versionNameRegex` accepts `alpha|beta|rc`.

The straightforward fix is to drop the hand-rolled comparator for `semver.compare` (see the separate comment on line 5).

packages/cli/src/release/npm/resolve-publish-tag.ts
```diff
@@ -0,0 +1,38 @@
...
+ * as `latest` on the registry (not off branch names or publish order).
+ *
+ * @throws {NpmViewError} if the registry can't be queried for a reason other than the package
+ * having never been published (E404) — a network blip, registry 5xx, or auth error must not be
+ * silently treated as "nothing published yet", or it could regress `latest`. Callers should let
+ * this fail the publish rather than falling back to a tag.
+ */
+export function resolveNpmDistTag(packageName: string, version: Version): string {
+    const currentLatestRaw = npmViewDistTag(packageName, 'latest');
+    const currentLatest = currentLatestRaw ? parseVersionName(currentLatestRaw) : null;
```
lskramarov: **An unparseable published `latest` silently degrades to “nothing published yet” — the very hazard `npmViewDistTag` throws to avoid.**

`npmViewDistTag` goes to real lengths to distinguish “genuinely never published” (E404 → `null`) from “we couldn’t check” (throw). Line 29 then discards an equally ambiguous signal: if `parseVersionName` cannot parse what the registry reported, `currentLatest` becomes `null` and line 33 returns `latest`.

`versionNameRegex` is deliberately narrow — `^(\d+)\.(\d+)\.(\d+)(?:-(alpha|beta|rc)\.(\d+))?$`. It rejects any other prerelease label and any build metadata, both of which are valid semver and both of which npm will happily serve:

```
published latest = 20.2.0-next.1   publishing 19.9.1  ->  latest   # regresses the dist-tag from 20.x to 19.x
published latest = 20.2.0+build.7  publishing 19.9.1  ->  latest
```

A parse failure here should throw (or reuse `NpmViewError`) rather than fall through to `latest`. Comparing with `semver.compare` on the raw string sidesteps it entirely.

.github/workflows/publish.yml
```diff
@@ -46,9 +46,12 @@ jobs:
         run: yarn run check-npm-resolution
 
       - name: Publish package
+        # No `-t` here: `@koobiq/cli` now resolves the dist-tag itself per package (`latest` vs
+        # `v<major>-lts`) against what's currently published, so a patch on an older major's
+        # maintenance branch (which reuses this same workflow file) can't regress `latest`.
```
lskramarov: **The maintenance branches do not reuse this file, and their copy still passes `-t latest`.**

Line 11 of this same file already says the opposite: *“since that branch keeps its own copy of this file”*. Verified:

```
$ git show origin/19.x:.github/workflows/publish.yml | grep release:publish
          npm run release:publish:ci-github -- -t latest
```

`origin/19.x` also carries its own pre-change `packages/cli`. So until this PR is backported to 19.x (and 18.x), a 19.x patch tagged after 20.x ships **still** overwrites `latest` — while this comment tells the next reader the case is covered.

Suggested wording, and it is probably worth adding the backport to the PR checklist:

```suggestion
        # No `-t` here: `@koobiq/cli` now resolves the dist-tag itself per package (`latest` vs
        # `v<major>-lts`) against what's currently published. Each maintenance branch keeps its
        # own copy of this file and its own `packages/cli`, so this must be backported there
        # before a patch on an older major stops regressing `latest`.
```

lskramarov (commented):
## Code review — `max` effort

15 findings on the dist-tag policy, ranked most severe first as inline comments. Two are blocking:

1. **`release.packages` holds dist directory names, not npm package names** — `npm view components dist-tags.latest` resolves an unrelated public package (`0.1.0`), so the LTS guard never fires and the feature is a no-op for this repo.
2. **`dotenv@17` prints a banner to stdout** — the documented `TAG=$(koobiq-cli resolve-npm-tag …)` capture is broken under the version `@koobiq/cli` declares.

Also worth a look before merge: `compareVersions` ignores `alpha`/`beta`/`rc` ordering, an unparseable published `latest` silently falls back to `latest`, and the `publish.yml` comment claims a guarantee the 19.x branch does not have (its copy still passes `-t latest`).

Verified locally: `tsc -p packages/cli/tsconfig.lib.json --noEmit` clean, `jest --roots packages/cli` 60/60 green, `yarn run eslint` clean — the findings are behavioural, not build breaks.

_Generated with Claude Code_

packages/cli/src/release/npm/resolve-publish-tag.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    // `>= 0`, not `> 0`: republishing the exact version already tagged `latest` (a retry of a
+    // failed CI run, for instance) must stay `latest`, not fall through to `v<major>-lts`.
+    if (!currentLatest || compareVersions(version, currentLatest) >= 0) {
+        return 'latest';
+    }
+
+    return `v${version.major}-lts`;
```
lskramarov: **`v<major>-lts` is written unconditionally, so it can be regressed the same way `latest` can.**

Only `latest` is compared against its current value. Scenario: `v19-lts` points at `19.9.5`, then a hotfix is cut from an older `19.4.x` support branch and `19.4.2` is published. The comparison is against `latest` (20.x), so this returns `v19-lts`, and `npm publish --tag v19-lts` moves the LTS pointer **back** from `19.9.5` to `19.4.2`. Anyone pinned to `@koobiq/components@v19-lts` silently downgrades five minors.

Same bug class the PR fixes for `latest`, one level down — the resolved tag should also be checked against its own currently published version before being reused.

packages/cli/src/release/publish-release-ci.ts
```diff
@@ -98,9 +99,23 @@ export class PublishReleaseCITask extends BaseReleaseTask {
         // console.info(green(`  ✓   GitHub release is posted.`));
     }
 
+    /**
+     * Resolves the npm dist-tag for every package before publishing any of them, so a failure
+     * resolving one package's tag (e.g. an ambiguous npm view error) aborts before anything has
+     * been published, instead of leaving a partial release across packages.
+     */
+    private resolveNpmDistTags(): Map<string, string> {
```
lskramarov: **`resolveNpmDistTags()` is pasted byte-for-byte into all four publish tasks.**

Identical 15 lines (5-line JSDoc + 10-line body) at:

- `publish-release-ci.ts:102-114`
- `publish-release-from-dist.ts:83-95`
- `publish-release-github-ci.ts:85-97`
- `publish-release-gitlab-ci.ts:87-99`

All four extend `BaseReleaseTask`, and all four declare `currentVersion: Version` with the same initializer, so this fits cleanly as a single `protected resolveNpmDistTags(version: Version): Map<string, string>` on the base class.

As it stands, each of the other findings on this PR — wrong package identifier, empty `-t ''`, error reporting — has to be fixed in four places. It also repeats the same rationale comment four times, which `AGENTS.md` pushes against (*“Prefer clear naming and structure over comments”*).

packages/cli/src/release/resolve-npm-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        const versionString = this.readVersionFromPackageJson(projectDir);
+        const version = parseVersionName(versionString);
+
+        if (!version) {
+            console.error(red(`  ✘   Cannot parse "${versionString}" as a Semver version.`));
+            process.exit(1);
+        }
+
+        console.log(resolveNpmDistTag(packageName, version));
```
lskramarov: **`NpmViewError` is never caught, so the error the module was designed around surfaces as a raw stack.**

Two paths, both unhandled:

- Here (sync): a registry 5xx prints a Node stack trace and exits 1. A caller using the documented `TAG=$(koobiq-cli resolve-npm-tag …)` **without** `set -e` gets `TAG=''` and then runs `npm publish --tag ''`.
- In the four publish tasks: `resolveNpmDistTags()` throws inside `async run()`, and `cli.ts` calls `.run()` without a `.catch()`. It becomes an unhandled rejection — fatal on Node 24, so the publish does abort, but with no actionable message.

Every other failure in these files uses `console.error(red(...))` + `process.exit(1)`. Given the `@throws` doc says *“Callers should let this fail the publish”*, catching `NpmViewError` and reporting it in that style would make the intent hold end-to-end.

packages/cli/src/release/npm/resolve-publish-tag.ts
```diff
@@ -0,0 +1,38 @@
...
+ * silently treated as "nothing published yet", or it could regress `latest`. Callers should let
+ * this fail the publish rather than falling back to a tag.
+ */
+export function resolveNpmDistTag(packageName: string, version: Version): string {
+    const currentLatestRaw = npmViewDistTag(packageName, 'latest');
+    const currentLatest = currentLatestRaw ? parseVersionName(currentLatestRaw) : null;
+
+    // `>= 0`, not `> 0`: republishing the exact version already tagged `latest` (a retry of a
+    // failed CI run, for instance) must stay `latest`, not fall through to `v<major>-lts`.
+    if (!currentLatest || compareVersions(version, currentLatest) >= 0) {
```
lskramarov: **A prerelease that outranks the published `latest` is given the `latest` dist-tag.**

Published `latest` is `20.2.0`; tagging `20.3.0-alpha.0` resolves to `latest`, so `npm i @koobiq/components` starts installing an alpha.

The `github-release` job in this PR's own `publish.yml` takes the opposite position for the same tag:

```bash
case "$TAG" in
  *-*)
    prerelease=(--prerelease) # a prerelease can never be latest
```

This is inherited from the old hardcoded `-t latest` rather than introduced here, but this PR is the point where tag selection becomes a stated policy — worth making the two agree (prerelease → `next`, or its own label).

packages/cli/src/release/npm/npm-client.ts
```diff
@@ -52,6 +52,41 @@ export function npmPublish(packagePath: string, distTag: string): string | undef
...
+ * so those cases throw instead of silently returning null.
+ */
+export function npmViewDistTag(packageName: string, tag: string): string | null {
+    const result = spawnSync('npm', ['view', packageName, `dist-tags.${tag}`], {
+        shell: true,
+        env: npmClientEnvironment
+    });
+
+    if (result.status !== 0) {
+        const stderr = result.stderr.toString();
```
lskramarov: **`result.stderr` is dereferenced without a null check, and `result.error` is never inspected.**

`spawnSync` sets `status: null` and can leave `stdout`/`stderr` as `null` when the spawn itself fails (EACCES on the shell, EMFILE under a parallel release, a killed process). `status !== 0` is true for `null`, so line 74 is reached and throws `TypeError: Cannot read properties of null (reading 'toString')` — a type no caller expects, carrying none of the diagnostics the `NpmViewError` message was written for.

`@types/node` types `stderr` as non-nullable, so `tsc` will not flag it. Even on the reachable path the message renders as `failed (exit null)`.

Worth checking `result.error` first and defaulting `stderr` (`result.stderr?.toString() ?? ''`).

packages/cli/src/release/base-release-task.ts
```diff
@@ -15,7 +15,8 @@ export interface IReleaseTaskConfig {
     projectDir: string;
     repoToken: string;
     distDir: string;
-    tagName: string;
+    /** npm dist-tag to publish under. Auto-resolved per package when omitted. */
+    tagName?: string;
```
lskramarov, line 21: **Now that `tagName` is optional, `??` lets an empty string through to `npm publish`.**

The four call sites use `this.config.tagName ?? resolveNpmDistTag(...)`. Nullish coalescing only catches `null`/`undefined`, so `koobiq-cli publish-ci-github -t ''` gives commander `options.tagName === ''`, that evaluates to `''`, and `npmPublish` runs `npm publish --access public --tag ''` — rejected by npm, per package, *after* `checkReleaseOutput` has already passed.

The codebase already treats the empty string as “unset” for this exact field — `changelog-writer-options.spec.ts:43` builds an `IReleaseTaskConfig` with `tagName: ''`.

Either use `||`, or normalise empty → `undefined` once at the CLI boundary.

packages/cli/src/release/npm/resolve-publish-tag.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        expect(resolveNpmDistTag('@koobiq/icons', parseVersionName('12.1.1')!)).toBe('latest');
+    });
+
+    it('tags a patch on an older major as "v<major>-lts", not "latest"', () => {
+        npmViewDistTag.mockReturnValue('11.3.0');
+
+        expect(resolveNpmDistTag('@koobiq/icons', parseVersionName('10.10.1')!)).toBe('v10-lts');
+    });
+
+    it('tags a lower version of the current major as "v<major>-lts"', () => {
```
lskramarov: **Test name and body disagree, and the body duplicates the test above it.**

The name promises “a lower version of **the current major**”, but the body publishes `11.7.2` against a published `latest` of `12.1.1` — a *different* major, which is exactly what the preceding test (`11.3.0` / `10.10.1`) already covers.

So the case the name claims is untested: a lower version of the *same* major, which resolves to `v12-lts` and creates an LTS tag on the still-current major — arguably surprising given the JSDoc says *“Older majors get their own `v<major>-lts` tag”*.

```suggestion
    it('tags a lower version of the current major as "v<major>-lts"', () => {
        npmViewDistTag.mockReturnValue('12.1.1');

        expect(resolveNpmDistTag('@koobiq/icons', parseVersionName('12.0.9')!)).toBe('v12-lts');
    });
```

packages/cli/src/release/npm/npm-client.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    }) as unknown as ReturnType<typeof spawnSync>;
+
+describe(npmViewDistTag.name, () => {
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('returns the published version on success', () => {
+        spawnSyncMock.mockReturnValue(mockResult({ status: 0, stdout: '12.1.1\n' }));
+
```
lskramarov, line 26: **No test asserts what command is actually built.**

All six cases stub `spawnSync` and assert only on the return value, so the suite stays green if the identifier passed in is wrong (see the `release.packages` comment on `publish-release-github-ci.ts`), if the field were misspelled `dist-tag.${tag}`, or if `npmClientEnvironment` were dropped. One assertion closes that:

```ts
expect(spawnSyncMock).toHaveBeenCalledWith(
    'npm',
    ['view', '@koobiq/icons', 'dist-tags.latest'],
    expect.objectContaining({ shell: true })
);
```

Related gaps in `resolve-publish-tag.spec.ts`: no prerelease case (would have caught the alpha/beta/rc ordering bug), no unparseable-`latest` case, and nothing verifying that a thrown `NpmViewError` propagates out of `resolveNpmDistTag` — the single behaviour the module's JSDoc calls load-bearing.

packages/cli/src/release/npm/resolve-publish-tag.ts
```diff
@@ -0,0 +1,38 @@
+import { Version, parseVersionName } from '../version-name/parse-version';
+import { npmViewDistTag } from './npm-client';
+
+/** Positive when `a` outranks `b`, negative when `b` outranks `a`, zero when equal. */
+function compareVersions(a: Version, b: Version): number {
```
lskramarov: **`semver` is already in the repo and handles the cases this comparator gets wrong.**

`semver@7.8.5` is resolved today (root `package.json:143`). `semver.compare(a, b)` implements the full precedence rules — `alpha < beta < rc`, numeric vs alphanumeric identifiers, build metadata ignored — so comparing the raw registry string with `semver.compare` / `semver.gte` removes both the prerelease-ordering bug (line 10) and the unparseable-`latest` fallback (line 29) at once, instead of routing the registry's answer through `parseVersionName`'s deliberately narrow *release* grammar.

Caveat: `semver` is currently a devDependency, so it would need adding to `packages/cli/package.json` `dependencies`.

packages/cli/src/release/resolve-npm-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+    packageName?: string;
+}
+
+/**
+ * Standalone command wrapping `resolveNpmDistTag` so any koobiq repo's CI can resolve the
+ * correct npm dist-tag (`latest` vs `v<major>-lts`) without reimplementing the semver-against-
+ * registry comparison itself. Prints only the resolved tag to stdout, so callers can do
+ * `TAG=$(koobiq-cli resolve-npm-tag --package-name @koobiq/icons)`.
+ */
+export class ResolveNpmTagTask {
```
lskramarov, line 22: **Two new files with near-identical names.**

`release/resolve-npm-tag.ts` (this CLI task) and `release/npm/resolve-publish-tag.ts` (the resolution function) are hard to tell apart, and the exported symbol matches the *other* file: line 4 here reads `import { resolveNpmDistTag } from './npm/resolve-publish-tag';`.

Grepping `resolve-npm-tag` lands on the wrapper; grepping `resolveNpmDistTag` lands on `resolve-publish-tag.ts`. Naming the task after its command (`resolve-npm-tag-task.ts`) or the core file after its export (`npm/resolve-npm-dist-tag.ts`) removes the collision.

packages/cli/src/release/publish-release-github-ci.ts
NikGurev: Точно, тут был баг. Раньше проверка брала имя папки сборки (например "components"), а не настоящее имя пакета в npm (@koobiq/components) — из-за этого сравнение версий уходило в чужой, случайный пакет и по факту не работало. Поправил: имя пакета теперь читается из package.json внутри собранной папки, и заодно вся эта логика переехала в один общий метод (в base-release-task.ts), а не дублируется в четырёх местах.

packages/cli/src/release/resolve-npm-tag.ts
NikGurev: Поправил — при запуске cli теперь всегда включаем тихий режим для dotenv, так что этот баннер больше не попадёт в переменную с тегом.

packages/cli/src/release/npm/resolve-publish-tag.ts
NikGurev: Согласен, порядок alpha/beta/rc не учитывался. Убрал своё сравнение версий и перешёл на библиотеку semver, которая уже есть в проекте и умеет сравнивать версии правильно, включая пререлизы.

packages/cli/src/release/npm/resolve-publish-tag.ts
NikGurev: Теперь, если реестр вернул что-то, что не похоже на нормальную версию, функция не притворяется, что ничего не публиковалось, а сразу падает с понятной ошибкой. Это часть перехода на semver из соседнего комментария.

.github/workflows/publish.yml
NikGurev: Поправил текст комментария — теперь он честно говорит, что старые ветки (19.x и более ранние) пока используют свою копию файла, и что перед патчем на такой ветке эти изменения нужно туда перенести отдельно.

packages/cli/src/release/npm/resolve-publish-tag.ts
NikGurev: Хорошее замечание. Теперь тег v<major>-lts тоже защищён: если публикуемая версия старше той, что уже висит на этом теге, публикация останавливается с понятной ошибкой, а не молча откатывает тег назад.

packages/cli/src/release/publish-release-ci.ts
NikGurev: Вынес общую функцию в базовый класс (base-release-task.ts) — теперь она в одном месте, а не скопирована в четыре файла публикации.

packages/cli/src/release/resolve-npm-tag.ts
NikGurev: Теперь эта ошибка перехватывается и печатается как обычное понятное сообщение (так же, как остальные ошибки в этих скриптах), а не падает сырым стектрейсом.

packages/cli/src/release/npm/resolve-publish-tag.ts
NikGurev: Поправил политику: любая пререлизная версия (alpha/beta/rc) теперь всегда публикуется под отдельным тегом next и никогда не может случайно стать latest.

packages/cli/src/release/npm/npm-client.ts
NikGurev: Поправил — если сам процесс npm не смог запуститься, это тоже превращается в понятную ошибку, а не в TypeError, и обращение к stderr больше не упадёт, если он окажется пустым.

packages/cli/src/release/base-release-task.ts
NikGurev: Дублирование ушло вместе с переносом функции в базовый класс, и там уже используется ||, так что пустая строка в теге больше не проскочит.

packages/cli/src/release/npm/resolve-publish-tag.spec.ts
NikGurev: Поправил тест — теперь он действительно проверяет то, что написано в названии (более старая версия внутри той же мажорной ветки), а не дублирует соседний тест.

packages/cli/src/release/npm/npm-client.spec.ts
NikGurev: Добавил тест, который проверяет саму команду, а не только результат — теперь если имя пакета или нужный флаг потеряются, тест это заметит.

packages/cli/src/release/npm/resolve-publish-tag.ts
NikGurev: Согласен, так надёжнее — переписал сравнение версий на semver вместо самодельной реализации.

packages/cli/src/release/resolve-npm-tag.ts
NikGurev: Переименовал файл с самой функцией в resolve-npm-dist-tag.ts (по названию экспортируемой функции resolveNpmDistTag) — теперь его легко отличить от resolve-npm-tag.ts.

lskramarov (approved):

NikGurev:
Dry-run режим 

Новая версия (>20.2.0)
image

Патч версия предыдущего минора (20.1.1)

image

PR #1967, opened 2026-08-31
feat(schematics): added ng-add automated set up (#DS-4683)

NikGurev:
## Summary

Добавлены
- Установка темы (`light | dark | auto`) вместе со стилями и сервисом KbqThemeService (если выбрано `'auto'`)
- Обновление index.html в зависимости от выбранной темы
- импорт provideAnimations , в зависимости от решения пользователя  
- Вывод сообщения об установке шрифтов, так как способы подключения шрифтов могут отличаться.
- Вывод сообщения о сохранении настроек темы если проект использует SSR

image

image

image


artembelik (approved):

packages/schematics/src/ng-add/index.ts
```diff
@@ -19,13 +22,77 @@ const VERSIONS = {
...
 
+/**
+ * Whether any `.ts` file under `projectRoot` already contains `needle`. `addRootProvider` has no
+ * built-in guard against inserting the same expression twice, so callers use this to skip
+ * insertion on re-run instead of accumulating duplicate provider calls.
+ */
+function projectSourceContains(tree: Tree, projectRoot: string, needle: string): boolean {
+    let found = false;
+
+    tree.getDir(projectRoot).visit((filePath) => {
```
lskramarov: **`node_modules` is inside the scan, so both guards false-positive and no provider is added.**

`ng new` writes `"root": ""` for the app it creates (`@schematics/angular/ng-new/index.js:27` passes `projectRoot: ''`). `HostTree._normalizePath` is `normalize('/' + path)`, so `getDir('')` resolves to `/`, and the `ng add` workflow tree is a `ScopedHost` on the workspace root with no `node_modules` exclusion.

`ng add` installs the package *before* running the schematic, so by the time this runs the tree contains:

- `node_modules/@angular/platform-browser/animations/index.d.ts:81` — `declare function provideAnimations(): Provider[];`
- `node_modules/@koobiq/components/core/index.d.ts` — 4 occurrences of `kbqThemeProvider(` in JSDoc

So on a plain `ng new my-app && ng add @koobiq/components`, **neither `kbqThemeProvider({ mode })` nor `provideAnimations()` is inserted**, and nothing is logged either — the fallback messages only fire from `safeRule`'s catch, and this is the "already present" path. Every animating component then throws NG05105, which is exactly what the unconditional warning this PR removed was there to prevent. Reproduced independently three times against a real `HostTree`.

Secondary cost: `HostDirEntry.visit` materializes the whole recursive file list before calling the visitor, so `if (found) return` cannot short-circuit the walk. Measured on a real workspace root: 99,243 files visited, 22,335 `.ts` read, 113 MB, **23.9 s per walk** — and there are four walks per project on the default path.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -69,13 +136,88 @@ export default function ngAdd(options: Schema): Rule {
...
+                        messages.themeProviderFallback(wired.projectName, theme)
+                    )
+                );
+            }
+
+            // `kbqThemeProvider({ mode: 'auto' })` only configures `KbqThemeService` — the service
+            // is `providedIn: 'root'` and nothing in the library injects it eagerly, so without an
+            // initializer forcing that first injection, the OS color scheme is never actually
+            // applied until something else (e.g. a theme-toggle button) happens to inject it.
+            if (theme === 'auto' && !projectSourceContains(tree, wired.projectRoot, 'inject(KbqThemeService)')) {
```
lskramarov: **This initializer crashes the SSR build.**

`addProviderToStandaloneBootstrap` targets `findAppConfig(...)?.filePath` — i.e. `app.config.ts`, not `main.ts` — and the generated `app.config.server.ts` does `mergeApplicationConfig(appConfig, serverConfig)`. So `ApplicationInitStatus` runs this initializer on the server too.

`KbqThemeService` constructs `matchMedia` in a field initializer with no platform guard:

```ts
// packages/components/core/services/theme.service.ts:274
private readonly media = this.window.matchMedia('(prefers-color-scheme: dark)');
```

`grep -c matchMedia` is 0 across `@angular/platform-server/fesm2022/*.mjs` and `@angular/ssr`, so `ng new my-app --ssr` + `ng add @koobiq/components` (default `theme: 'auto'`) fails the next `ng build` with `TypeError: this.window.matchMedia is not a function` at prerender.

Nothing in `@koobiq/components` injects `KbqThemeService` eagerly today, so this rule is what turns a latent opt-in limitation into a guaranteed build failure on the default path. The loop already knows the answer — `wired.isSsr` is computed at line 164 and used only for a cosmetic suggestion.

Worth considering whether the deeper fix belongs in the library: a `provideEnvironmentInitializer(() => inject(KbqThemeService))` inside `kbqThemeProvider` (plus a platform guard on `media`) would fix both this path and the documented manual setup, and would let the schematic stop generating boilerplate it can never remove without a migration.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -19,13 +22,77 @@ const VERSIONS = {
...
+
+    tree.getDir(projectRoot).visit((filePath) => {
+        if (found || !filePath.endsWith('.ts')) return;
+
+        const content = tree.read(filePath)!.toString('utf-8');
+
+        if (!THEME_PROVIDER_CALL.test(content)) return;
+
+        found = true;
+        const updated = content.replace(THEME_PROVIDER_CALL, `kbqThemeProvider({ mode: '${theme}' })`);
```
lskramarov: **Mutates the first matching file anywhere under the project root, then reports success.**

`found = true` is set before the write, the visitor early-returns for every remaining file, and the regex has no `g` flag — so exactly one occurrence in one arbitrarily-chosen file is rewritten.

Verified: a project with `kbqThemeProvider({ mode: 'light' })` in a `*.spec.ts` fixture, run with `--theme=dark`, gets **the spec file** rewritten to `mode: 'dark'` (corrupting a deliberate test fixture) while `app.config.ts` ends up with no theme provider at all and no warning — the guard at line 167 short-circuits on the `true` return.

Which file wins depends on traversal order (`_recurse` emits `subfiles` before descending into `subdirs`), so the outcome differs across platforms. Combined with the `root: ""` scope problem it can `tree.overwrite` a file inside `node_modules`; in a multi-app workspace it rewrites a sibling project's `app.config.ts` on this project's behalf.

Scoping this to the file `addRootProvider` will actually edit — `findAppConfig` / `findProvidersLiteral` from `@schematics/angular/utility/standalone/` — removes the whole class of problem.

_Generated with Claude Code_

packages/schematics/src/utils/html-config.ts
```diff
@@ -0,0 +1,48 @@
...
+    const classMatch = CLASS_ATTR.exec(attrs);
+    const existingClasses = classMatch ? classMatch[1].split(/\s+/).filter(Boolean) : [];
+    const keptClasses = existingClasses.filter((cls) => !KNOWN_THEME_CLASSES.includes(cls));
+    const newClasses =
+        theme === 'auto'
+            ? [...keptClasses, 'kbq-app-background']
+            : [...keptClasses, 'kbq-app-background', `kbq-${theme}`];
+    const newClassValue = newClasses.join(' ');
+
+    const newAttrs = classMatch
```
lskramarov: **A single-quoted `class` produces a duplicate attribute and the theme silently never applies.**

`CLASS_ATTR` only matches double quotes, so `<body class='app'>` misses and falls into the append branch. Verified by running the function:

```
in:  <body class='app'>                 (theme: light)
out: <body class='app' class="kbq-app-background kbq-light">
```

The HTML5 tokenizer drops the duplicate attribute, so the browser keeps only `class='app'` — no `kbq-app-background`, no theme — while `ng add` reports success. And it never self-heals: a second run matches the appended double-quoted attribute, computes an identical tag, and returns `changed: false`. `<body class=app>` behaves the same.

_Generated with Claude Code_

lskramarov (commented):

15 findings, ranked most-severe first. Every claim below was reproduced against real source or a real `HostTree`; the empirical output is quoted inline.

Two of them break the default path outright:

1. **`index.ts:33`** — the `.ts` guards walk `tree.getDir("")`, which is the whole workspace *including `node_modules`*, whenever `project.root` is `""` (what `ng new` writes). Angular's and koobiq's own shipped `.d.ts` files satisfy the needles, so on a plain `ng new` + `ng add` **neither `kbqThemeProvider` nor `provideAnimations()` is inserted, and nothing is logged** — NG05105 on the first dropdown.
2. **`index.ts:184`** — the `auto`-theme `provideAppInitializer(() => inject(KbqThemeService))` lands in `app.config.ts`, which the server config merges, and `KbqThemeService` calls `matchMedia` in a field initializer. `ng new --ssr` + `ng add` fails the next `ng build` at prerender.

The rest cluster into: the substring/regex guards matching the wrong thing (`index.ts:41/62/199`), `html-config.ts`'s regex not being HTML-aware (four distinct corruption modes, all reproduced), `safeRule` leaving orphaned imports behind, and a couple of gaps where a failure produces no message at all.

The test suite can't currently see any of this — `newProjectRoot: 'projects'` means no fixture ever has `root: ""`, and there is no `node_modules` in the tree.

Worth flagging as a possible deeper fix: several findings dissolve if `kbqThemeProvider` itself carried a `provideEnvironmentInitializer(() => inject(KbqThemeService))` (with a platform guard on `media`), and if the HTML/workspace editing went through `@angular/cdk/schematics`' `addBodyClass` / `getProjectIndexFiles` / `getProjectBuildTargets` instead of hand-rolled regex and a forked builder allowlist.
_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -69,13 +136,88 @@ export default function ngAdd(options: Schema): Rule {
...
+                            const kbqThemeService = external('KbqThemeService', '@koobiq/components/core');
+
+                            return code`${provideAppInitializer}(() => { ${inject}(${kbqThemeService}); })`;
+                        }),
+                        messages.themeServiceFallback(wired.projectName)
+                    )
+                );
+            }
+
+            if (animations && !projectSourceContains(tree, wired.projectRoot, 'provideAnimations(')) {
```
lskramarov: **A substring match over every `.ts` file treats a comment or a spec as "already configured".**

Verified on this suite's own `projects/app` fixture:

- appending `// TestBed.configureTestingModule({ providers: [provideAnimations()] });` to `app.spec.ts` → `app.config.ts` gets no `provideAnimations`, **and no warning**
- prefixing `app.ts` with `// TODO: configure kbqThemeProvider( later` → same for the theme provider

A `providers: [provideAnimations()]` line in a spec is very common in Angular apps. There is no "already present, skipping" message on this path — `animationsManualSetup(projectName)` only fires from `safeRule`'s catch — so the user has no way to notice.

The `'inject(KbqThemeService)'` needle at line 184 has the mirror problem: it misses `inject(KbqThemeService, { optional: true })`, a line-wrapped call, and the schematic's own output when `code_block.js:66` has to alias the import to `KbqThemeService_alias` — in which case every re-run appends another `provideAppInitializer`.

`findProvidersLiteral(config)` in `@schematics/angular/utility/standalone/util.js` — the module `addRootProvider` is built on — answers this against the actual `providers` array.

_Generated with Claude Code_

packages/schematics/src/utils/html-config.ts
```diff
@@ -0,0 +1,48 @@
...
+        ? attrs.replace(CLASS_ATTR, `class="${newClassValue}"`)
+        : `${attrs} class="${newClassValue}"`;
+    const newTag = `<body${newAttrs}>`;
+
+    if (newTag === fullTag) return { content: html, changed: false };
+
+    return { content: html.replace(BODY_TAG, newTag), changed: true };
```
lskramarov: **`$` sequences in the existing `<body>` attributes are expanded as replacement patterns.**

Both `.replace()` calls (this line and line 41) pass a computed string, so `$&`, `` $` ``, `$'`, `$n` and `$$` carried over from the user's markup are interpreted rather than inserted. Verified:

```
in:  <body data-q="a$&b">              (theme: light)
out: <body data-q="a<body data-q="a$&b">b" class="kbq-app-background kbq-light">
```

The whole matched tag is spliced into the attribute value and unparseable HTML is written back to the user's project. `<body data-format="$1.00">` expands `$1` to capture group 1; `class="a$$b"` is silently rewritten to `class="a$b …"`. `BODY_TAG` has a capture group, so `$<…>` syntax fires too.

A replacer function (or escaping `$` → `$$` in the replacement) fixes this specific hazard.

_Generated with Claude Code_

packages/schematics/src/utils/html-config.ts
```diff
@@ -0,0 +1,48 @@
+import { KoobiqTheme } from './workspace-styles';
+
+const KNOWN_THEME_CLASSES = ['kbq-app-background', 'kbq-light', 'kbq-dark'];
+const BODY_TAG = /<body\b([^>]*)>/i;
+const CLASS_ATTR = /class\s*=\s*"([^"]*)"/i;
```
lskramarov: **Neither regex is anchored to what it means to match.**

`CLASS_ATTR` has no boundary before `class`, so it matches the tail of any attribute name ending in `class`. Verified:

```
in:  <body data-class="x" id="r">      (theme: dark)
out: <body data-class="x kbq-app-background kbq-dark" id="r">
```

An unrelated attribute is corrupted and `<body>` never gets a `class` at all. Same for `ng-class`.

`BODY_TAG` takes the first textual `<body` in the document, including one inside a comment — e.g. the Google Tag Manager install snippet, which is routinely pasted into `index.html`:

```
in:  <!-- put after <body> tag --><html><body class="real">
out: <!-- put after <body class="kbq-app-background kbq-light"> tag -->  (real body untouched)
```

In both cases `changed: true`, the file is written, no theme is applied, and a re-run reports `changed: false` so it never recovers. `[^>]*` also truncates on a `>` inside an attribute value (`<body data-x="a>b">`), splicing the emitted tag mid-attribute.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -19,13 +22,77 @@ const VERSIONS = {
...
+/**
+ * Runs `rule` against `tree` and swallows any error it throws (e.g. `addRootProvider` failing to
+ * statically analyze a non-standard bootstrap), logging `fallbackMessage` instead — one project's
+ * unusual setup shouldn't abort the rest of `ng add`.
+ */
+function safeRule(rule: Rule, fallbackMessage: string[]): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        try {
+            await firstValueFrom(callRule(rule, tree, context));
+        } catch {
```
lskramarov, line 138: **The catch fires after `addRootProvider` has already written its import, and it flattens every error class into one message.**

`addProviderToStandaloneBootstrap` returns `chain([...rules, () => …insertStandaloneRootProvider…])`, where `rules` come from `CodeBlock.transformPendingCode` — which calls `tree.commitUpdate(recorder)` to write the `import` **before** the insertion runs. `insertStandaloneRootProvider` then throws (`standalone/rules.js:148`).

Verified with `bootstrapApplication(App, someConfig, extraArg);` in `main.ts`: the run leaves

```ts
import { kbqThemeProvider } from '@koobiq/components/core';
import { provideAnimations } from '@angular/platform-browser/animations';
```

with no corresponding providers — up to 5 orphaned imports on the `auto` path, failing the consumer's `noUnusedLocals` / `no-unused-vars` while `ng add` exits 0. The warning tells them to add providers but never to remove the imports.

Because the catch is untyped, a corrupt `angular.json`, an `EACCES`, or a bug in the koobiq callback all surface as "Could not automatically register", so real regressions ship silently. The expected case is separable: `isStandaloneApp` + `findBootstrapApplicationCall` are a pre-flight check available before anything is written.

_Generated with Claude Code_

packages/schematics/src/utils/workspace-styles.ts
```diff
@@ -0,0 +1,152 @@
...
+const THEME_TOKENS_PATTERN = /^node_modules\/@koobiq\/design-tokens\/web\/css-tokens-(light|dark)\.css$/;
+
+/**
+ * The style entries documented in the installation guide, in the order they must be loaded.
+ * `'auto'` includes both token files, since `KbqThemeService` switches between them at runtime —
+ * a static `'light'`/`'dark'` choice only needs its own.
+ */
+export function getKoobiqStyleEntries(theme: KoobiqTheme): string[] {
+    const [icons, tokens, prebuiltTheme] = KOOBIQ_STATIC_STYLE_ENTRIES;
+    const themeTokens = theme === 'auto' ? [
```
lskramarov, line 31: **`light`/`dark` ship one token file, but the service can still switch to the other.**

The schematic passes only `mode`, never `themes`, so `theme.service.ts:277` keeps `themes: KBQ_DEFAULT_THEMES` — both light and dark stay registered. Two ways that bites after `--theme=light`:

- `readInitialMode()` (`theme.service.ts:349-353`) returns the **stored** value first and falls back to `config.mode` only when storage is empty, so a stale `kbq-theme-mode: 'dark'` in `localStorage` from a previous session silently overrides the configured mode
- `toggle()` (`:345`) is what `docs/guides/theming.en.md:50` documents as *the* switching API

Either way `applyTheme` puts `kbq-dark` on `document.body`, but `css-tokens-dark.css` — the only file declaring anything under `.kbq-dark` — was never added to `angular.json`, so every `--kbq-*` colour resolves to nothing and the app renders unstyled.

`docs/guides/theming.en.md:18` explicitly says to include all three token files. Either always ship both, or pin `themes` to the single registered theme so the service can't switch to one that has no tokens.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -69,13 +136,88 @@ export default function ngAdd(options: Schema): Rule {
...
+                        addRootProvider(wired.projectName, ({ code, external }) => {
+                            return code`${external('provideAnimations', '@angular/platform-browser/animations')}()`;
+                        }),
+                        messages.animationsManualSetup(wired.projectName)
+                    )
+                );
+            }
+        }
+
+        if (!animations) {
```
lskramarov, line 335: **The animations instruction is now unreachable in the one case where it mattered most.**

On `main` this warning was unconditional. Now it only fires when the user *opts out*, so when `wiredProjects` is empty the guidance disappears entirely.

Verified with `ng add @koobiq/components --project=my-lib` (also: any project whose `build` builder is outside `APPLICATION_BUILDERS`, or a `--no-create-application` workspace). The `noProject` check passes, `applyKoobiqWorkspaceStyles` returns `[]`, the loop body never runs, and `animations` is `true` — so the only output is `noWiredProjects()`, which talks about "styles and the theme class" and never mentions animations, plus `fontsSuggestion()`.

The user ends up with `@angular/animations` in `package.json`, no `provideAnimations()`, and no hint that they need one.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -19,13 +22,77 @@ const VERSIONS = {
...
+
+    tree.getDir(projectRoot).visit((filePath) => {
+        if (found || !filePath.endsWith('.ts')) return;
+        if (tree.read(filePath)!.toString('utf-8').includes(needle)) found = true;
+    });
+
+    return found;
+}
+
+const THEME_PROVIDER_CALL = /kbqThemeProvider\(\{\s*mode:\s*'(?:light|dark|auto)'\s*\}\)/;
```
lskramarov: **Quote-style-fragile: a re-run flips the styles but silently leaves the provider stale.**

The pattern hard-codes single quotes and an exact single-property object. First run writes `kbqThemeProvider({ mode: 'light' })`; the consumer's Prettier pass (default `singleQuote: false`) rewrites it to `"light"`, or a hand-edit adds a trailing comma — `{ mode: 'dark', }` fails the regex, verified.

Then `ng add --theme=dark`:

- `updateThemeProviderMode` returns `false` (regex miss)
- `projectSourceContains(…, 'kbqThemeProvider(')` still returns `true`
- the two-clause guard at 166-169 is therefore false → no rule pushed, no `else`, **no `logMessage`**
- meanwhile `applyKoobiqWorkspaceStyles` has already swapped the token file and `setKoobiqThemeBodyClass` the `<body>` class

Result: dark styles, light provider, zero diagnostic. (Multi-line formatting does still match — `\s*` spans newlines — so the triggers are quote style, trailing comma, or any extra option.)

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
```diff
@@ -69,13 +136,88 @@ export default function ngAdd(options: Schema): Rule {
...
-        logMessage(context.logger, [
-            'Angular animations have to be provided by the application.',
-            "Add `provideAnimations()` from '@angular/platform-browser/animations' to the providers",
-            'of `bootstrapApplication`, otherwise every component that animates (dropdown, select,',
-            'tooltip, toast, datepicker) fails with NG05105 as soon as it opens.'
-        ]);
+        // Wiring styles: prepend the documented style entries to every application project's
+        // `build` target (or just `project`, when given), then apply the theme class to each
+        // project's index.html.
+        const workspace = await readWorkspace(tree);
```
lskramarov: **Moving this read out of `if (project)` makes it a hard failure where it used to be skipped.**

`readWorkspace` only ever probes `/angular.json` (`@schematics/angular/utility/workspace.js:21` pins `DEFAULT_WORKSPACE_PATH`), and `HostTree.readText` throws `FileDoesNotExistException` when the file is absent — while the CLI itself accepts either name (`@angular/cli/src/utilities/config.js:89`: `['angular.json', '.angular.json']`).

Verified against the built `dist/ng-add/index.js`: a tree containing only `/.angular.json` passes the CLI's in-workspace scope check yet gives `FileDoesNotExistException | Path "/angular.json" does not exist.` Nothing catches it — `safeRule` wraps only the later `addRootProvider` rules — so the rule rejects, the tree is discarded, and the user gets **none** of the dependency writes at lines 116-132 that worked before this PR.

Rare config name, but a real regression from a previously guarded read; the same throw is what an Nx-shaped tree hits, which is a bit ironic given `@nx/angular:webpack-browser` was deliberately added to the builder allowlist.

While here: this is the second `readWorkspace(tree)` in the function — the one at line 98 is parsed, used for a single `.get()`, and thrown away. Nothing between the two touches `angular.json`, so it can be hoisted and reused.

_Generated with Claude Code_

packages/schematics/src/utils/workspace-styles.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export type KoobiqTheme = 'light' | 'dark' | 'auto';
+
+/**
+ * Builders that represent an application's browser bundle. `@angular/build:application` is the
+ * builder generated by the current `@schematics/angular` `application` generator; the
+ * `@angular-devkit/build-angular:*` names are kept for workspaces still on an older Angular CLI,
+ * which this schematic's peer range still supports.
+ */
+const APPLICATION_BUILDERS = new Set<string>([
```
lskramarov: **An unlisted builder gets no setup at all, not degraded setup.**

A project on a wrapper builder (`@angular-builders/custom-webpack:browser`, `ngx-build-plus:browser`, `@nx/angular:browser-esbuild`, `@nx/angular:application` — none listed) is dropped from `wiredProjects`, and since `providerRules` is built only from that array (`index.ts:154`) it gets no styles, no `index.html` class **and no providers** — only a generic "follow the installation guide" warning. `addRootProvider` would have succeeded: `getMainFilePath` falls back to `options.main` for any unrecognized builder.

Gating on the target *literally named* `build` (line 136) narrows it further.

This also forks a list Angular already maintains: `@angular/cdk/schematics` ships the same four official names as `PROJECT_BUILDERS` behind `getProjectBuildTargets(project)`, which scans **all** targets by builder rather than one by name. `@schematics/angular` itself gates on `project.extensions.projectType === 'application'` (service-worker, web-worker, server). Either of those keeps koobiq out of the business of chasing Angular CLI and Nx release cycles.

_Generated with Claude Code_

packages/schematics/src/ng-add/ng-add.spec.ts
```diff
@@ -76,4 +92,164 @@ describe(`ng add '@koobiq/components'`, () => {
             "Unable to find project 'test' in the workspace"
         );
     });
+
+    describe('styles, theme and animations', () => {
```
lskramarov, line 96: **16 green tests, and none of them can see the primary failure mode.**

`packages/schematics/src/utils/testing.ts:7` pins `newProjectRoot: 'projects'`, so every fixture project has `root: 'projects/<name>'`. No test creates a project with `root: ''` — the layout `ng new` actually produces — and none seeds `node_modules/**/*.d.ts`.

That is why the suite passes while `ng new demo && cd demo && ng add @koobiq/components` produces an `app.config.ts` with neither `kbqThemeProvider` nor `provideAnimations`. Adding the two `.d.ts` stubs plus `projectRoot: ''` to a fixture reproduces it in under a second and would have caught this before merge.

The idempotency tests are similarly blind to the substring-guard false positives, since no fixture contains any of the needles in a comment or a spec file. Worth a case for each: needle-in-a-comment, needle-in-a-spec, and `root: ''`.

_Generated with Claude Code_

packages/schematics/src/utils/html-config.ts
```diff
@@ -0,0 +1,48 @@
...
+ * `kbq-app-background` paints the page's own background/text color, `kbq-{theme}` picks the theme.
+ * `'auto'` only gets `kbq-app-background`: `KbqThemeService` applies `kbq-light`/`kbq-dark` to
+ * `document.body` itself at runtime, so a static class here would just go stale before the app
+ * bootstraps, or fight the service once it does.
+ *
+ * Re-running is idempotent: any class this function manages is stripped from the existing `class`
+ * value first, so switching `theme` between runs doesn't leave a stale `kbq-light` sitting next to
+ * a newly added `kbq-dark`. Any other existing classes, and their relative order, are preserved.
+ */
+export function setKoobiqThemeBodyClass(html: string, theme: KoobiqTheme): HtmlEditResult {
```
lskramarov, line 153: **`@angular/cdk/schematics` already ships this, parse5-backed — and it has none of the four bugs above.**

`addBodyClass(host, htmlFilePath, className)` (`node_modules/@angular/cdk/schematics/utils/html-manipulation.js`) parses with `parse5({ sourceCodeLocationInfo: true })`, finds `body.attrs.find(a => a.name === 'class')` regardless of quoting, and is already idempotent. No duplicate-attribute case, no `data-class` case, no `<body>`-in-a-comment case, no `$&` expansion. The package also re-exports `parse5` itself for the stale-class strip `addBodyClass` doesn't cover.

Same for `getIndexHtmlPath` at `workspace-styles.ts:73` vs `getProjectIndexFiles(project)` (`utils/project-index-file.js`): same three `index` shapes, but unioned across every build target and joined with posix `join` — koobiq's `${project.root}/src` concat yields `/src/index.html` for a root-level project with no `sourceRoot`. Angular tracks its own changes there on every upgrade (v20 already made `index` optional).

`@angular/cdk` is a declared peer of `@koobiq/components` and a root dependency, so it is strictly safer to require at ng-add time than `@schematics/angular/utility`, which is only a devDependency here.

_Generated with Claude Code_

packages/schematics/src/ng-add/index.ts
NikGurev: #1950

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Раньше проверка «а вдруг это уже добавлено» смотрела на все файлы проекта, включая node_modules — из-за этого в новых проектах она ошибочно решала, что провайдеры уже есть, и ничего не добавляла. Теперь эта проверка смотрит только туда, где реально описаны провайдеры приложения (нахожу через официальный помощник Angular CLI), поэтому чужие файлы и папка node_modules больше не мешают.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Тема с автоопределением светлая/тёмная дёргала системную функцию, которой нет на сервере — из-за этого сборка с серверным рендерингом (SSR) падала. Добавил проверку «мы точно в браузере» перед тем как включать эту функцию.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Раньше при повторном запуске поиск «где лежит настройка темы» мог случайно найти и изменить не тот файл (например, файл теста), если совпадение находилось раньше нужного места. Теперь ищем и меняем строго в файле с настройками приложения, который определяется через официальный API схематики, а не перебором всех файлов.

packages/schematics/src/utils/html-config.ts
NikGurev: Поправил. Старый способ вставки класса в index.html через регулярные выражения ломался на одинарных кавычках — атрибут class задваивался, и браузер тихо игнорировал нашу тему. Переписал на собственную обработку html, а не поиск текста, так что кавычки и порядок атрибутов больше не важны.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил тем же способом, что и в первом комментарии: проверка «уже настроено» теперь смотрит только в реальный список провайдеров приложения, а не ищет текст по всем файлам. Так упоминание «kbqThemeProvider(» в комментарии или тестовом файле больше не собьёт схематику с толку.

packages/schematics/src/utils/html-config.ts
NikGurev: Поправил вместе с переходом на parse5 (см. соседние комментарии про index.html) — там больше нет замены текста через .replace() со спецсимволами, поэтому строка вида a$&b в атрибуте больше ничего не ломает.

packages/schematics/src/utils/html-config.ts
NikGurev: Поправил тем же переходом на parse5: теперь ищем именно тег <body> и именно атрибут class, а не любой похожий текст. Атрибут data-class и упоминание <body> внутри комментария (например, из кода Google Tag Manager) больше не путают схематику.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Теперь перед тем как что-то менять, сначала проверяем — получится ли вообще безопасно найти нужное место в файле. Если да — меняем напрямую, без «поймать любую ошибку и тихо продолжить». Если нет (необычный способ запуска приложения) — тогда честно предупреждаем и ничего не трогаем, так что лишних неиспользуемых импортов больше не остаётся.

packages/schematics/src/utils/workspace-styles.ts
NikGurev: Поправил. Раньше при выборе конкретной темы (light или dark) сервис всё равно знал про обе темы и мог случайно переключиться на ту, чьи стили не подключили. Теперь при выборе light/dark явно говорим сервису, что доступна только эта одна тема.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Раньше подсказка про анимации могла вообще не появиться, если схематика не нашла ни одного подходящего проекта. Теперь в этом случае подсказка тоже выводится.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил вместе с переходом на разбор кода через официальный API (см. первый комментарий). Теперь замена настройки темы при повторном запуске ищет вызов kbqThemeProvider по структуре кода, а не по конкретному написанию (кавычки, пробелы, запятая на конце), поэтому больше не может «не узнать» уже добавленную настройку и молча оставить её устаревшей.

packages/schematics/src/ng-add/index.ts
NikGurev: Поправил. Если конфиг воркспейса не получилось прочитать (например, он называется .angular.json, а не angular.json), схематика раньше падала с непонятной технической ошибкой. Теперь в этом случае сразу показываем понятное сообщение, до того как что-то успели изменить в package.json.

packages/schematics/src/utils/workspace-styles.ts
NikGurev: Поправил. Раньше схематика решала «это приложение или нет» по короткому списку известных сборщиков (builder), и незнакомый сборщик она просто пропускала целиком — без стилей и без провайдеров. Теперь решаем по типу проекта (application/library), как это делает сам Angular CLI, поэтому незнакомый сборщик больше не оставляет проект совсем без настройки.

packages/schematics/src/ng-add/ng-add.spec.ts
NikGurev: Добавил тесты на все описанные тут случаи: пустой корень проекта с node_modules, упоминание провайдера в комментарии или тестовом файле, незнакомый сборщик, отсутствие angular.json, серверный рендеринг с автотемой и случай, когда подходящих проектов не нашлось.

packages/schematics/src/utils/html-config.ts
NikGurev: Спасибо за наводку. Готовый addBodyClass из @angular/cdk не подошёл один в один — нам ещё нужно уметь снимать старый класс темы при повторном запуске, а он только добавляет. Но идею забрал: переписал вставку класса на том же parse5, которым пользуется CDK, так что все перечисленные тут проблемы (кавычки, комментарии, левые атрибуты) закрыты тем же способом.

packages/schematics/src/utils/html-config.ts
NikGurev: Небольшое уточнение к предыдущему ответу: от parse5 в итоге отказался. При сборке пакета всплыло предупреждение — библиотеки Angular не должны тянуть за собой рантайм-зависимости в dependencies, а parse5 нужен был только ради этой одной правки в одном файле. Вместо библиотеки написал свой маленький разборщик тега <body> и его атрибутов (учитывает и одинарные, и двойные, и вообще без кавычек кавычки, игнорирует < body внутри комментариев, не путает > внутри значения атрибута с концом тега). Результат тот же — просто без лишней зависимости. Комментарии рядом про кавычки, data-class и комментарий с <body> актуальны так же, там тоже поменялся способ, а не суть исправления.

NikGurev: Проверил — раньше тест был общий (просто "не сломалось"), добавил три отдельных теста именно на те примеры, что в комментарии: `$&`, `$1.00` в чужом атрибуте и `$$` в самом class, с точным ожидаемым результатом строка-в-строку. Все проходят, PR обновлён.

packages/schematics/src/utils/html-config.ts
NikGurev: Вместо библиотеки написал свой маленький разборщик тега и его атрибутов (учитывает и одинарные, и двойные, и вообще без кавычек кавычки, игнорирует < body внутри комментариев, не путает > внутри значения атрибута с концом тега). Результат тот же — просто без лишней зависимости.

lskramarov (approved):

PR #1972, opened 2026-09-01
fix(e2e): eliminate flaky screenshot tests (#DS-5461)

lskramarov:
## Why

`playwright.config.ts` retries a failing test twice on CI, so a test that passes on a retry is reported as flaky and does not fail the run. The suite has therefore been almost always green while being measurably unstable.

Mining every `e2e.yml` run still within log retention (99 runs) found that **43 of them (43 %) contained at least one test that passed only on a retry** — 28 distinct tests across 8 spec files. **Every red e2e run in that window was caused by one of those files**; not one failure was attributable to a genuine regression. No test ever timed out: every failure was a screenshot comparison.

| | Before | After |
| --- | --- | --- |
| Full suite, `PLAYWRIGHT_RETRIES=0 --repeat-each=5`, Docker, 8 workers | **27 failures** in 2810 results | **0 failures** in 2810 results |

## What

Nearly every failure was a screenshot taken before the state it captures had settled, so the fix is one idea applied consistently: **assert the settled state, then take the shot**. No component behaviour changed, and the tests that assert transient states on purpose are untouched.

| Cause | Components | Verified |
| --- | --- | --- |
| Scrollbar reveal window | `sidepanel`, `popover`, `content-panel`, `notification-center` | 17/50 failing → **0/10** |
| Captured before highlight.js landed | `code-block` | 5/30 → **0/40** at 16 workers |
| Anti-aliased edges of the autofill tint | `form-field` (3 tests) | **800/800** at 16 workers |
| `mode: 'serial'` as an amplifier | `dropdown` (16 tests → 1) | **0/320** in parallel contexts |
| Fixture decorating after the shot | `file-upload` | **0/10** |

### The dominant cause

`KbqScrollbarTrack.revealed` shows the track for `hideDelay` (1000 ms) after a scroll or a `flashScrollIndicators()`, then hides it. Three properties made that a flake generator:

1. **Playwright cannot freeze it.** `animations: 'disabled'` fast-forwards CSS animations and transitions; this is a `timer()` in TypeScript driving a class binding.
2. **It is one-way.** Once the timer fires nothing reveals the track again, so the assertion waits for a state that is *already gone* rather than one that has yet to arrive — which is exactly why retries never rescued this class.
3. **Nothing decided where a baseline landed.** The committed baselines disagreed with each other: `content-panel`'s held the revealed track, `sidepanel`'s the faded one.

A new `e2eWaitForSettledScrollbars` in `packages/e2e/utils/scrollbar.ts` waits for the settled state, which is the only one that stays reachable.

## Notes for review

- **Three baselines are regenerated** — `content-panel/01-{light,dark}.png` and `popover/01-light.png`. They had been recorded *inside* the reveal window, with the thumb; `sidepanel`'s was recorded outside it and is unchanged.
- **`maxDiffPixels: 200` on two `form-field` screenshots** is the only tolerance in the suite. `animations: 'allow'` has to stay there — otherwise the 600000 s `background-color` transition that suppresses Chrome's autofill background is fast-forwarded and the controls paint raw blue — so Playwright never stabilizes those shots. Every occurrence ever observed measured 8, 29, 33 or 86 pixels; the other flakes here moved 932, 2958 and 10742.
- **`datepicker` is not fixed.** One occurrence in 99 CI runs, and it would not reproduce in 200 local repeats. Nothing was changed on a guess; it is now visible rather than absorbed.
- **`PLAYWRIGHT_RETRIES` is added but the default is unchanged** (`isCI ? 2 : 0`). Suggest keeping CI at 2 until this has ridden a few weeks of real runs, then running it at 0 nightly before moving the default.

## Found along the way, out of scope

- `KbqCodeBlock.scrollTo` defers the scroll until highlighting reports itself done, but the scroll range keeps growing afterwards — so it rests one pixel short of the bottom in roughly one run in thirty and nothing corrects it. A consumer scrolling a code block to its end right after load hits the same thing. The spec works around it with a comment pointing here.
- `sidepanel` and `scrollbar/deprecated` opt into WebKit via `test.use({ browserName: 'webkit' })`, but the snapshot path template carries neither `{projectName}` nor `{platform}`, so 8 baselines share a flat namespace with the Chrome ones. Fixing it properly means a separate WebKit project, which re-renders those baselines under a Safari descriptor — kept out so it cannot confound the verification here.
- 91 `expect(await …)` reads that take a computed style or geometry once, with no retry, right after an async action. None observed failing.

## Verifying

```bash
PLAYWRIGHT_RETRIES=0 node tools/e2e/run.js \
  yarn playwright test packages/components --repeat-each=5 --trace=retain-on-failure
```

Traces have to be requested explicitly — the config captures them `on-first-retry`, which never happens when there are none. `PLAYWRIGHT_WORKERS=16` is what reproduced `code-block` and `form-field`; both were invisible at the default 8.

Full write-up with the measurements behind each fix: `docs/e2e-flakiness.md`.


lskramarov (commented):
Self-review at `max` effort: 10 finder angles plus targeted experiments in Docker. The fixes work — the full suite goes 27 → 0 failures at `--repeat-each=5` — but several of the new gates prove less than their comments claim, and one is dead.

**The load-bearing check.** I neutered `e2eWaitForSettledScrollbars` to a no-op, kept the regenerated baselines, and ran ×20 at 16 workers: **95 failures, all in `content-panel` (20/20) and `sidepanel` (18-20/20)**. `popover` and `notification-center` stayed at 0. So two of the four call sites are real and two are not — see the inline comments.

**Highest-value items:** the notification-center wait is dead code with a false comment and its CI flake is still undiagnosed; the code-block poll satisfies its own condition rather than waiting for highlighting; `maxDiffPixels` is the wrong knob for anti-aliasing noise.

**Two findings that could not be anchored inline** (the lines are outside the diff):

1. `packages/components/dropdown/e2e.playwright-spec.ts:25` — removing `mode: 'serial'` is right (it turned 1 flake into 16), but it exposes an ungated sequence: `hover()` on the submenu trigger is followed immediately by three `ArrowDown`s and an `ArrowRight`, while nested panels open through `.pipe(delay(0, asapScheduler))` plus a CD pass and an overlay attach. Under `fullyParallel` with `workers: '100%'`, if the overlay has not attached the keys drive the root panel's key manager instead and the following `hover()` blocks to the 15s timeout. Wants an explicit `toBeVisible` on the second `.cdk-overlay-pane`.

2. `tools/e2e/Dockerfile:25` — "CI=true selects the same Playwright behavior as the runner: retries: 2" is no longer unconditional now that `PLAYWRIGHT_RETRIES` overrides it and `docker-compose.yml` sets that variable on this exact image. The parallel sentence in `06-testing.md` was updated; this one was not, so the two now disagree.

**Verified clean:** baseline integrity (decoded pixel diffs show the three regenerated PNGs differ from main only inside the thumb rectangles — no font drift, nothing smuggled in); the `PLAYWRIGHT_RETRIES` plumbing end to end including the docs config, the update-snapshots overlay and the empty-string path; `resolveRetries` input handling; the `progress-bar`/`progress-spinner`/`content-panel` `waitForTimeout` removals; the added `await` in `actions-panel`; and the revealed-track coverage, which survives in nine other spec files.

_Generated with Claude Code_

packages/components/notification-center/e2e.playwright-spec.ts
```diff
@@ -10,6 +10,9 @@ test.describe('KbqNotificationCenterModule', () => {
             await page.goto('/E2eNotificationCenterStates');
             const locator = getComponent(page);
 
+            // The centre scrolls itself to the bottom on init, which reveals the scrollbar track.
```
lskramarov: **This wait is dead code, and the comment is factually wrong.**

The track is never revealed on this route, so `e2eWaitForSettledScrollbars` returns immediately and the 3831px CI flake this PR claims to fix is untouched.

Measured three ways:
- A `MutationObserver` installed through `addInitScript` and left running 3s from first paint logs **zero** appearances of `kbq-scrollbar-track_revealed`.
- A direct probe reads `revealed=0, tracks=1` at this line, and the helper returns in **8ms** — against 1311ms at `content-panel`.
- Neutering the helper to a no-op and re-running ×20 leaves notification-center at **0 failures**.

`scrollToBottom()` is reachable only from `subscribeToRevealLoadMoreRow` on a `loadingMore` false→true transition, which the static fixture never triggers — and the component's own doc says "the panel always opens scrolled to the top".

So the real cause of that CI failure is still unknown. Either diagnose it or drop the claim; right now the code and `docs/e2e-flakiness.md` both assert it is handled.

packages/components/code-block/e2e.playwright-spec.ts
```diff
@@ -7,6 +7,29 @@ test.describe('KbqCodeBlockModule', () => {
...
+            const scrolledCodeArea = page.getByTestId('e2eCodeBlockWithTabs').locator('.kbq-code-block__main');
+
+            // Driven here rather than left to the fixture. `KbqCodeBlock.scrollTo` defers until
+            // highlighting reports itself done, but the scroll range keeps growing after that as the
+            // line-numbers plugin restructures the block — so the component's own scroll rests one
+            // pixel short of the bottom in roughly one run in thirty, and nothing corrects it. One CSS
+            // pixel shifts the whole code area, which at threshold: 0 is a failure.
+            await expect
+                .poll(async () => {
+                    await scrolledCodeArea.evaluate((el) => el.scrollTo({ top: el.scrollHeight }));
```
lskramarov: **The poll makes its own condition true, so it never waits for highlighting.**

`scrollTo` without `behavior: 'smooth'` updates `scrollTop` synchronously, so iteration 1 returns 0 and the poll exits at once. Nothing in the predicate observes highlight.js, the row count, or the scroll range — yet the comment above and `docs/e2e-flakiness.md` both claim resting at the end "implies highlighting has landed".

If the dynamic import resolves after the poll returns, the line-numbers plugin grows `scrollHeight` 150→180px, `scrollTop` stays where the test put it, and nothing re-scrolls before the screenshot — the exact 1616-vs-1770px mismatch this block exists to fix.

It held 40/40 at 16 workers, so the window is narrow. But the gate does not close it. Pinning the post-highlight invariant (row count, or `scrollHeight` stable across two samples) and *then* scrolling would.

packages/components/code-block/e2e.playwright-spec.ts
```diff
@@ -7,6 +7,29 @@ test.describe('KbqCodeBlockModule', () => {
...
+
+                    return scrolledCodeArea.evaluate((el) => el.scrollHeight - el.clientHeight - el.scrollTop);
+                })
+                .toBe(0);
```
lskramarov: **Exact `toBe(0)` on a residual this repo already documents as fractional.**

`packages/components/notification-center/notification-center.ts:67-80` describes this exact expression:

> `scrollHeight` and `clientHeight` are CSSOM `long`s (rounded to the nearest integer) while `scrollTop` is an `unrestricted double` […] at the true bottom `scrollHeight - scrollTop - clientHeight` lands anywhere in (-1, 1) rather than on 0 — an exact `<= 0` check then succeeds only about half the time.

…and settles on `SCROLLED_TO_BOTTOM_TOLERANCE = 2`. This is stricter than the check that doc calls unreliable.

On any fractional layout the residual is a *fixed* non-zero value, so re-scrolling each iteration cannot converge: the poll burns its full 5s inside a 15s test and fails with "expected 0, received 0.5". That trades one flake for another.

```suggestion
                .toBeLessThanOrEqual(1);
```

packages/components/file-upload/e2e.playwright-spec.ts
```diff
@@ -7,6 +7,18 @@ test.describe('KbqFileUploadModule', () => {
...
         const getMultipleFileUploadTable = (locator: Locator) => locator.getByTestId('e2eMultipleFileUploadTable');
 
+        /**
+         * The hover and focus states in these tables are painted on by the fixture after the first
+         * render, and the single-file one is re-applied in a macrotask after the component clears it.
+         * `goto` resolving says nothing about either, so a shot taken without this shows no focus ring.
+         */
+        const expectFixtureDecorated = async (table: Locator) => {
+            await expect(table.locator('.dev-hover .kbq-file-upload__item:not(.kbq-hovered)')).toHaveCount(0);
+            await expect(table.locator('.dev-focused .kbq-file-upload__action:not(.cdk-keyboard-focused)')).toHaveCount(
```
lskramarov: **This gate is satisfied before the component clears the class, so it cannot prove the re-application.**

`e2e.ts:231-250` runs add (`afterNextRender`) → component focus monitor clears → `setTimeout` re-adds. A `:not(.cdk-keyboard-focused)` count of 0 is true in window 1 *and* window 3, so the check can pass at window 1 and the screenshot then lands in the cleared window 2 — precisely the state the new fixture comment says the spec waits past.

A positive count that is only reachable after the re-application would be one-way:

```suggestion
            await expect(table.locator('.dev-focused .kbq-file-upload__action.cdk-keyboard-focused')).toHaveCount(
                await table.locator('.dev-focused .kbq-file-upload__action').count()
            );
```

packages/components/file-upload/e2e.playwright-spec.ts
```diff
@@ -7,6 +7,18 @@ test.describe('KbqFileUploadModule', () => {
         const getSingleFileUploadTable = (locator: Locator) => locator.getByTestId('e2eSingleFileUploadTable');
         const getMultipleFileUploadTable = (locator: Locator) => locator.getByTestId('e2eMultipleFileUploadTable');
 
+        /**
+         * The hover and focus states in these tables are painted on by the fixture after the first
+         * render, and the single-file one is re-applied in a macrotask after the component clears it.
+         * `goto` resolving says nothing about either, so a shot taken without this shows no focus ring.
+         */
+        const expectFixtureDecorated = async (table: Locator) => {
+            await expect(table.locator('.dev-hover .kbq-file-upload__item:not(.kbq-hovered)')).toHaveCount(0);
```
lskramarov: **Permanently vacuous for the single-file table.**

`singleFileUploadRows` (`e2e.ts:116-162`) has no `className: 'dev-hover'` cell — only `dev-focused` — and `.kbq-file-upload__item` exists only in `multiple-file-upload.component.html`. Either condition alone makes this selector match zero elements in `KbqSingleFileUploadComponent states`, so it is 0-of-0 and can never fail there.

The helper reads as covering both tables symmetrically while covering only one, and it would stay green if the hover decoration broke entirely.

packages/components/form-field/e2e.playwright-spec.ts
```diff
@@ -515,8 +515,13 @@ test.describe('KbqFormFieldModule', () => {
...
                 // Nothing in the autofill rules keys on `:hover`; the state does, and the state is
                 // what owns every channel except the tint.
-                expect(await paint(control)).toBe(before.control);
-                expect(await paint(container)).toBe(before.container);
+                //
+                // Polled rather than read once: `hover()` resolves as soon as the pointer has moved,
+                // and a single read right after it fails on whatever the frame happened to hold. A
+                // rule that really did key on `:hover` still fails here, because the poll never
+                // returns to the pre-hover value.
+                await expect.poll(() => paint(control)).toBe(before.control);
```
lskramarov: **`expect.poll` turns a must-always invariant into a may-once one.**

The property under test is "hovering changes nothing, ever". `expect.poll` invokes its callback immediately and exits on the **first** sample equal to `before.control` — which is exactly what is on screen the instant `hover()` resolves. A `:hover` rule that applies with any latency (a transition, a `transition-delay`, a class set on rAF) is sampled before it lands and passes.

It is worse for `backgroundColor` specifically: this control carries the 600000s autofill suppression transition asserted at lines 468 and 765 of this same file. A `:hover` background-color change restarts that transition from the current computed value, so the string stays equal to `before.control` for the whole 5s poll. That channel becomes unfalsifiable.

The comment's claim that such a rule "still fails here" holds only for an instantaneous, permanent change. The single read was the right shape; the fix for its flakiness is to establish that the hover actually landed *first*, then read once.

packages/components/form-field/e2e.playwright-spec.ts
```diff
@@ -736,8 +741,16 @@ test.describe('KbqFormFieldModule', () => {
...
+             * `maxDiffPixels` is the price of that: these are the only shots in the suite Playwright
+             * does not stabilize, so the anti-aliased edges of the autofill tint land a few units
+             * either side of a rounding boundary from run to run. Every occurrence measured — four on
+             * CI and one in 800 local repeats — differed by 8, 29, 33 or 86 pixels, each of them an
+             * edge pixel with a delta of single digits per channel. The cap sits above that and an
+             * order of magnitude below anything real: the other flakes this suite has produced moved
+             * 932, 2958 and 10742 pixels.
              */
-            const screenshot = { animations: 'allow' } as const;
+            const screenshot = { animations: 'allow', maxDiffPixels: 200 } as const;
```
lskramarov: **`maxDiffPixels` is the wrong knob for this cause — `threshold` is the right one.**

The measured noise is single-digit-per-channel on anti-aliased edges (`rgba(174,185,208)` vs `rgba(179,189,211)`, ≈2% RGB). That is a **magnitude** problem. `maxDiffPixels: 200` instead admits 200 pixels differing by *any* amount, because the project pins `threshold: 0` globally.

At `deviceScaleFactor: 2` that now hides:

| Regression on one cell | device px | caught? |
|---|---|---|
| missing 1px border | ~1856 | yes |
| lost 2px focus ring | ~3700 | yes |
| `border-radius` dropped, four corners | ~55 | **no** |
| recoloured thin icon glyph | ~115 | **no** |
| 1px seam shorter than ~50 CSS px | <200 | **no** |

That last row matters — this describe's own docstring names "the 1px seam the focus geometry leaves behind" as a reason these shots exist.

```suggestion
            const screenshot = { animations: 'allow', threshold: 0.05 } as const;
```

packages/components/popover/e2e.playwright-spec.ts
```diff
@@ -88,6 +89,9 @@ test.describe('KbqPopoverModule', () => {
             await page.goto('/E2ePopoverStates');
             const locator = getComponent(page);
 
+            // The popovers flash their scrollbar track on open.
```
lskramarov: **This call site is not load-bearing, and the comment's mechanism was not observable.**

Neutering `e2eWaitForSettledScrollbars` to a no-op and running ×20 at 16 workers produces 95 failures — **all** in `content-panel` (20/20) and `sidepanel` (18-20/20). Popover and notification-center stay at **0**. A `MutationObserver` over 3s from first paint also logs no reveal on this route.

So what fixed popover was the regenerated `01-light.png`, not this wait: with a settled baseline, `toHaveScreenshot`'s own 5s retry outlasts the 1000ms `hideDelay`. Worth noting that contradicts the helper's JSDoc ("Neither does a retry") — here the retry is exactly what does.

Two of the four call sites are carrying a wait that never fires. Keeping them is defensible as insurance, but the comments should not assert a mechanism that does not run.

packages/e2e/utils/scrollbar.ts
```diff
@@ -1,13 +1,35 @@
...
+ * machine happened to be on, so the thumb is present in some runs and absent in others.
+ *
+ * `animations: 'disabled'` does not help: the reveal is an RxJS timer driving a class binding, not
+ * a CSS animation. Neither does a retry, because the assertion would be waiting for a state that is
+ * already gone rather than one that has yet to arrive.
+ *
+ * The settled state is the only one that stays reachable, which is why this waits for the track to
+ * be hidden rather than for it to be shown.
+ */
+export const e2eWaitForSettledScrollbars = (root: Page | Locator): Promise<void> =>
```
lskramarov: The helper itself is sound where it fires — measured blocking 1311ms at `content-panel` and 1303ms at `sidepanel`, and removing it makes those two fail 20/20 and 18-20/20. Two smaller points:

**1. It is satisfied by "the reveal has not started yet" as well as by "settled".** `toHaveCount(0)` resolves on its first poll, and 0 is also the count before the fixture scrolls. It happens to land inside the window at the two sites that matter, but that is timing, not structure. The correct shape is six lines below in this same file — `e2eExpectNoScrollbarAfterFlash` waits for `toHaveClass(/kbq-scrollbar-track_revealed/)` *first*.

This bites hardest when **recording** a baseline: with no `expected`, `toHaveScreenshot` only waits for two stable frames, well inside the 1000ms window, so a revealed track gets written as the new baseline. That is plausibly how `content-panel/01-light.png` acquired the bad baseline this PR regenerates.

**2. No type-check gate was added for the bug class fixed just above.** `tsconfig.playwright-spec.json` is consumed only by Playwright's transpiler, and `linters.yml` runs cspell/prettier/stylelint/eslint but no `tsc`. `npx tsc -p tsconfig.playwright-spec.json --noEmit` exits 0 on this branch today, so adding it is a one-line change that goes green immediately and closes the class instead of fixing one instance by hand.

tools/e2e/run.js
```diff
@@ -163,12 +163,12 @@ if (isUpdatingSnapshots) {
 
 const env = { ...process.env, PLAYWRIGHT_VERSION: version };
 
-// docker-compose.yml reads all three, and in forwarded mode it is parsed by a process on the other
+// docker-compose.yml reads all four, and in forwarded mode it is parsed by a process on the other
 // side of the WSL boundary, which does not inherit the Win32 environment. WSLENV is what carries a
 // variable across; /u marks it as travelling in that direction only. Names of variables that are
 // not set are ignored, and an entry repeated from an existing WSLENV is harmless.
```
lskramarov: **This claim is false, and it is the comment this PR edited.**

Measured: with `WSLENV=ZZ_NEVER_SET/u:ZZ_EMPTY/u:ZZ_ZERO/u` and only two of the three set on Win32, the Linux side reports `ZZ_ZERO=0`, `ZZ_EMPTY=`, **`ZZ_NEVER_SET=`** — the unset name is materialised as an empty-valued variable. A control run with no `WSLENV` shows none of the three.

So on every Windows-forwarded run, `E2E_PLATFORM` and `PLAYWRIGHT_RETRIES` reach the compose process as `""` even when the developer set neither. Survivable only because all three interpolations use `:-` (colon-minus, where empty counts as unset).

The day someone writes `${E2E_PLATFORM-linux/amd64}`, the platform pin resolves to empty — the pin `docker-compose.yml` says is what stops "an arm64 machine from quietly producing renders that differ from CI" — and it breaks *only* on Windows-via-WSL, the one platform this forwarding exists for. This comment would send that author looking elsewhere.

docs/e2e-flakiness.md
```diff
@@ -0,0 +1,180 @@
...
+**Fix.** The spec waits for the fixture to have finished decorating the table it is about to shoot.
+**Verified:** 0 failures in 10 repeats.
+
+## Not fixed
+
+- **`datepicker › scrolls back to the part the caret returns to`** — 1 occurrence in 99 CI runs, and it
+  did not reproduce in 200 local repeats under 16 workers. No mechanism established, so nothing was
+  changed. Left for the next occurrence, which will now be visible rather than absorbed.
+- **`{projectName}` in the snapshot path template.** `sidepanel` and `scrollbar/deprecated` opt into
+  WebKit with `test.use({ browserName: 'webkit' })`, but `playwright.config.ts:84` carries neither
```
lskramarov: **Four `file:line` references in this document are broken by this same commit.**

- `playwright.config.ts:84` was the `pathTemplate` on `main`; the 28-line `resolveRetries` block moves it to **112**, and line 84 is now `testDir`.
- `code-block/e2e.ts:132` → **133** (the `data-testid` this PR adds).
- `form-field/e2e.playwright-spec.ts:740` → **753** (the comment this PR adds).
- `file-upload/e2e.ts:241` — the `setTimeout` is 240 on main, 243 at HEAD; 241 matches neither.

Correct as written: `scrollbar.ts:644`, `code-block-highlight.ts:21`, `actions-panel/e2e.playwright-spec.ts:30`.

Naming the symbol rather than the line survives the next edit.

docs/e2e-flakiness.md
```diff
@@ -0,0 +1,180 @@
...
+  `Promise<void>`. Nothing type-checks these files — `playwright.config.ts` only compiles them,
+  without checking — so
+  it compiled and failed `tsc --noEmit`. Fixed.
+
+## Retry policy
+
+`PLAYWRIGHT_RETRIES` now overrides the config; the default is unchanged at `isCI ? 2 : 0`. See
+[06-testing.md](guides/06-testing.md).
+
+Keep CI at 2 until this has ridden a few weeks of real runs. Then run `PLAYWRIGHT_RETRIES=0` nightly,
```
lskramarov, line 183: **The nightly this plan turns on cannot be run — not even by hand.**

`.github/workflows/e2e.yml` triggers only on `push: branches: [main]` and `pull_request:`. There is no `schedule:` and no `workflow_dispatch:`, though seven other workflows in this repo have one or both.

So the three-step plan (keep 2 → observe at 0 nightly → flip the default) has no step 2, and the evidence it says is needed before changing the default can never be collected on CI. A `workflow_dispatch` with a `retries` input on the existing `tests` job — which already passes `PLAYWRIGHT_WORKERS` via step `env:` — would be a few lines.

docs/e2e-flakiness.md
```diff
@@ -0,0 +1,180 @@
+# E2E flakiness: audit and fixes
```
lskramarov, line 1: **This document is unreachable.**

`grep -rn "e2e-flakiness"` across the repo returns only self-references. It links out to `guides/06-testing.md`; that file does not link back, and `docker-compose.yml`'s new comment also points only at 06-testing.md. Every other prose guide lives at `docs/guides/NN-*.md`.

It is pinned to a commit and a Playwright version, and it carries three forward-looking commitments — the retry-policy plan, the `{projectName}` snapshot-namespace collision, and the `KbqCodeBlock.scrollTo` defect (filed "separately" with no ticket id). Nobody will find them again once this PR scrolls out of view.

Either link it from 06-testing.md, or move the durable parts there and let the rest be PR-description material.

lskramarov:
Review findings addressed in b410d05b3. All 15 fixed; three of them changed what the tests actually prove.

**The two waits that did nothing are gone.** Measured with a `MutationObserver` running from first paint: neither `popover` nor `notification-center` ever reveals a scrollbar track, so both calls were no-ops with comments claiming otherwise. `popover` was fixed by its regenerated baseline; `notification-center`'s flake is therefore still undiagnosed, and is now recorded under **Not fixed** next to `datepicker` instead of looking like coverage.

**The other two are load-bearing, and I checked rather than assumed.** Stubbing `e2eWaitForSettledScrollbars` to a no-op while keeping the regenerated baselines fails 95 times in a ×20 run at 16 workers — 20/20 in `content-panel`, 18–20/20 across `sidepanel`. The helper now waits for the expected track count *before* waiting for none revealed, since "nothing is revealed" is equally true of a viewport whose track has not been created yet. Counts are passed per call site from measured values, so a track that never arrives fails loudly.

**Gates that were satisfied by their own side effect or by the wrong window:**

- `code-block` gates on the rebuilt line-number rows first, then re-scrolls. Scrolling first satisfied the check with its own side effect. The remainder now allows a pixel — same reason `notification-center` carries `SCROLLED_TO_BOTTOM_TOLERANCE`.
- `file-upload` waits on a marker the fixture sets last. The old count was zero before the classes were applied and again after the focus monitor cleared them, and half of it matched nothing at all in the single-file table.
- `form-field` waits until the control actually matches `:hover`, then reads once.

**`threshold: 0.05` replaces `maxDiffPixels: 200`.** The noise is a magnitude (~2 % of the YIQ range on edge pixels); a count-based cap large enough for it would also hide the one-pixel seam those shots exist to catch.

**Infrastructure:** `dropdown` now waits for the nested overlay before the keys that drive it; `yarn run check-e2e-types` is wired into the linters workflow so the `Promise<Disposable>` class of bug is caught rather than fixed by hand; `workflow_dispatch` takes a `retries` input so the suite can actually be run at zero retries on CI; the WSLENV and Dockerfile comments are corrected; the audit doc is linked from `06-testing.md` and its line references replaced with symbol names.

**Verified:** 2810 results / 0 failures at 8 workers, and 2040 / 0 at 16 workers across every touched component.

_Generated with Claude Code_

artembelik (approved):

NikGurev (approved):

PR #1974, opened 2026-09-02
fix(dropdown): search and keyboard navigation (#DS-5122)

artembelik:


lskramarov (commented):
Review of `82d8387` (max-effort pass: 10 finder angles, one verifier per candidate, gap sweep). 15 inline comments, ranked most severe first; every item was re-verified against the PR-head files before posting.

**What matters most**

- The `items.changes` re-highlight pipeline leaves a destroyed item as the key manager's active item after an empty-result query, so ENTER clicks a detached element (navigates for `<a href>` items) and the next ArrowDown skips the first item; the same gate re-reads `this.search()!` after `delay(0)` and can throw, which silently kills the pipeline for the rest of the dropdown's life.
- `contentChild(KbqDropdownSearch)` is not scoped to the owning panel: a search field in a nested dropdown declared inside the parent flips the parent into search mode and leaves nothing focused when it opens.
- Nested triggers in a search panel move DOM focus out of the field (ENTER replay and hover) and nothing brings it back, contradicting the documented "caret stays in the field" contract.
- `scrollActiveItemIntoView()` runs on every keystroke reaching the panel (snapping a wheel-scrolled list back on each typed character) while the auto-activation path never scrolls.
- Each key-manager swap leaks the previous manager (two per open/close cycle with `kbqDropdownContent`), since the forked `ListKeyManager` subscribes in its constructor and has no teardown.
- The new active-descendant keyboard model ships without any ARIA wiring; the highlight is a CSS class only.

**Before merging**: the branch is 5 commits behind `origin/main` and conflicts on rebase in `dropdown-item.component.ts` (#1972 rewrote `KbqDropdownItem.focus()` right where this PR inserts `setActiveStyles`/`setInactiveStyles`). Main now forces `preventScroll: true` on item and panel focus and reveals through `element.scrollIntoView?.({ block: 'nearest', inline: 'nearest' })` (`kbqFocusAndReveal` in core); the new `search.focus()` and `scrollActiveItemIntoView()` should follow the same policy after the rebase.

**Minor notes not posted inline**

- No Playwright/visual coverage for the new `_active` state (incl. the selected variant and the adjacent-radius merge) or the borderless in-overlay search field; `dropdown/e2e.ts` has no search fixture, so a theme regression would only be caught by jsdom class checks.
- `applyInitialFocus` / `focusFirstItem` JSDoc (dropdown.component.ts:611-628) still describe only the origin-based branches; the search branch now runs first for every origin.
- The spec's `open()` helper is hard-wired to one host, so six tests re-inline the open sequence, four of them without the trailing `detectChanges()`; a generic `openDropdown(host, providers)` would remove the drift.
- The new describe has no `jest-axe` assertion while seven sibling specs run `axe(overlayContainerElement)`; it likely passes today (no roles), so this is a convention nudge.
- The hover decision is split between the item (`inSearchMode?.()` veto) and the panel (`setActiveItem` on hover); moving it wholly into the panel would drop hover-focus for custom `KbqDropdownPanel` implementers, so the current shape is the conservative one and this is only a note.

**On the existing Copilot comments**: the first two look incorrect (`reset()` exists on `AbstractControlDirective`, which `NgControl` extends, and `value(): string` is explicitly typed; CI is green on both), the third (stale `_active` styles after a manager swap) is valid and is addressed together with the key-manager leak comment.

_Generated with Claude Code_

packages/components/dropdown/dropdown.component.ts
```diff
@@ -377,10 +401,29 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
...
+        // Driven by the item list rather than by the query, because consumers commonly debounce their
+        // own filtering. `delay(0)` keeps the write out of the pass that reported the new items.
+        this.directDescendantItems.changes
+            .pipe(
+                filter(() => !!this.search()),
+                delay(0),
+                filter(() => !!this.search()!.value() && this.isActiveItemStale()),
+                takeUntilDestroyed(this.destroyRef)
+            )
+            .subscribe(() => this.keyManager.setFirstItemActive());
```
lskramarov: **Stale active item survives an empty-result query: ENTER then clicks a destroyed element and ArrowDown skips the first item.**

The forked `ListKeyManager` never clears an active item that disappeared from the list (it only re-indexes one that is still present), and `setFirstItemActive()` returns without touching state when the only remaining item is disabled or the list is empty (`list-key-manager.ts:438-450`). On the shipped example: open → ArrowDown (`New tab` active) → type a query with no matches (`Nothing found` only) → ESC clears the query → all items re-render as new instances, but that pass is gated out by `!!this.search()!.value()`, so `activeItem` stays the destroyed instance at index 0 and nothing is highlighted. From there:

- ENTER → `activeItem.getHostElement().click()` on a detached element: `preventDefault()` swallows the key and nothing happens. For `<a href kbq-dropdown-item>` (the link example) a disconnected `<a>` still follows its hyperlink, so ENTER navigates to a filtered-out URL.
- ArrowDown → `setNextItemActive()` sees index 0 and moves to item 1, skipping the first item (or does nothing if the stale index was the last one).

A second problem in the same gate: `this.search()!` is re-read one macrotask after line 410 checked it. With `kbqDropdownContent` and noop animations, an item click that changes the list synchronously schedules the `delay(0)` timer, the leave player finishes in a microtask and `lazyContent.detach()` destroys the search field, and the timer then evaluates `undefined.value()`. RxJS turns that into an error notification, the subscription closes (no error handler, nothing resubscribes), and the auto-highlight is dead for the rest of the dropdown's life. With real animations it is a ~1 ms race; a consumer hiding the field with `@if` in a later tick reproduces it deterministically.

The pipeline also has no open-state gate: with non-lazy content the projected `@for` keeps being change-detected while the panel is closed, so a list re-render landing after a close with a retained query (a debounced filter result, async data) highlights item 0 while closed, and the next open shows a pre-highlighted item that ENTER activates without any arrowing (a normal search-mode open has no active item). `KbqSelect` runs the same pipeline gated on `panelOpen` (`select.component.ts:1203`); here `panelAnimationState === 'enter'` is the open flag.

Dropping the stale reference whenever the item is gone, re-highlighting only when a query is present, reading the query defensively and gating on the open state keeps the covered case (`'t'` → `Two`) and fixes all four paths:

```suggestion
                filter(() => this.panelAnimationState === 'enter' && this.isActiveItemStale()),
                takeUntilDestroyed(this.destroyRef)
            )
            .subscribe(() => {
                // Drop the stale reference first: with no enabled result `setFirstItemActive` is a no-op, and
                // ENTER would replay a click on the destroyed item while the arrows resume from its old index.
                this.resetActiveItem();

                if (this.search()?.value()) {
                    this.keyManager.setFirstItemActive();
                }
            });
```

A spec for it: ArrowDown → `filter('zzz')` → `filter('')` → flush → ArrowDown → expect `One` (currently `Two`), and ENTER after clearing leaves `selected` null. Side note: `isActiveItemStale()` itself is a verbatim copy of `KbqSelect.isActiveItemStale()` (`select.component.ts:1970-1975`); a `hasStaleActiveItem()` on the forked `ListKeyManager` would serve both callers, but that is a follow-up.

packages/components/dropdown/dropdown.component.ts
```diff
@@ -113,11 +119,17 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
     private ngZone = inject(NgZone);
     private document = inject(DOCUMENT);
     private defaultOptions = inject<KbqDropdownDefaultOptions>(KBQ_DROPDOWN_DEFAULT_OPTIONS);
+    private readonly destroyRef = inject(DestroyRef);
 
-    private readonly search = contentChild(KbqFormField);
+    private readonly panelFormField = contentChild(KbqFormField);
+
+    private readonly search = contentChild(KbqDropdownSearch);
```
lskramarov: **`contentChild(KbqDropdownSearch)` is not scoped to the owning panel, so a search field inside a nested dropdown flips the parent into search mode.**

Signal `contentChild()` defaults to `descendants: true`. With the supported pattern of declaring a nested `<kbq-dropdown>` inside the parent's content (recursive-template example, `NestedDropdownDeclaredInsideParentDropdown` spec host), a `<kbq-form-field kbqDropdownSearch>` that belongs to the nested panel is created with the consumer view and becomes the root's `search()` too. The root then builds an `ActiveDescendantKeyManager`, `applyInitialFocus` calls `focus()` on an input whose DOM node is not connected (a no-op), so nothing in the root panel is focused: its `(keydown)` never fires (arrows / ESC / TAB dead), hovering root items no longer focuses them (`dropdown-item.component.ts:193`), and `hasSearch()` pins the root's width. Before this PR the same composition focused the first item.

`items` already solves exactly this with `item.parentDropdownPanel === this` (`updateDirectDescendants`, line 769). The search query needs the same ownership filter, e.g. have the directive inject `KBQ_DROPDOWN_PANEL` and pick the match that belongs to this panel:

```ts
// dropdown-search.ts
readonly panel = inject(KBQ_DROPDOWN_PANEL, { optional: true });

// dropdown.component.ts
private readonly searches = contentChildren(KbqDropdownSearch, { descendants: true });
private readonly search = computed(() => this.searches().find((search) => search.panel === this));
```

`panelFormField` / `hasSearch` has the same reach (it only affects the width lock, and that part predates the PR). A spec host combining the nested-declared pattern with a search in the nested panel, asserting the root focuses its first item on keyboard open, fails today.

Related, and independent of scoping: `applyInitialFocus` has no fallback when `search.focus()` cannot take focus at all, e.g. the control is `disable()`d while options load or the field is `[hidden]`. The `else` branches are skipped, nothing inside the overlay is focused, and ESC/arrows are dead until an outside click. Main's version always reached `focusPanel()` / `setFirstItemActive()`. A one-line guard after `search.focus()` covers both this and the nested case above:

```ts
if (!this.panelFormField()?.control().focused) {
    this.focusPanel();
}
```

packages/components/dropdown/dropdown.component.ts
```diff
@@ -540,12 +583,22 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
                 }
 
                 break;
+            case ENTER:
+                // The highlighted item never sees the key event itself, so replay it as a click and let
+                // consumer handlers and nested triggers behave exactly as they do for the mouse.
+                if (this.inSearchMode() && this.keyManager.activeItem) {
+                    event.preventDefault();
+                    this.keyManager.activeItem.getHostElement().click();
```
lskramarov, line 624: **With a nested trigger in a search panel, the caret leaves the field and never comes back.**

This comment promises nested triggers behave like the mouse, and the docs say the caret stays in the field, but neither the keyboard nor the hover path restores it:

- **Keyboard:** ArrowDown to a nested trigger → ENTER → the replayed click opens the submenu with `openedBy` undefined → `focusFirstItem('program')` puts real DOM focus on the first submenu item. LEFT / ESC there → `destroy('keydown')` → `this.focus(undefined)` → `nativeElement.focus()` on the trigger **item** in this panel (`dropdown-trigger.directive.ts:461-462, 361`). From that state ArrowDown moves `.kbq-dropdown-item_active` while `.cdk-focused` stays on the item (two rows in hover state), typed letters go to a manager without typeahead, and ESC now closes the whole root instead of clearing the query (the directive's handler is on the form-field, no longer on the event path).
- **Mouse:** hovering the nested trigger opens it with `openedBy = 'mouse'` → the submenu's `applyInitialFocus('mouse')` → `focusPanel()` takes the caret. Hovering a sibling closes it via the hover closing action (no reason, no restore) and the focused panel is detached → `document.activeElement` is `body`. In normal mode the next `mouseenter` would re-focus an item; search mode returns early at `dropdown-item.component.ts:193`, so arrows, typing and ESC are dead until the user clicks the input.

`search.focus()` is only called at open time (line 637); `KbqSelect` re-focuses its field after every key manager move (`select.component.ts:2142-2146`). One way to close the loop is in `KbqDropdownTrigger.destroy()`: for a nested trigger whose `parent.inSearchMode()`, hand the caret back via `this.parent.focusFirstItem(reason === 'keydown' ? 'keyboard' : 'program')` regardless of `reason`/`openedBy`, and cover it with a spec host that combines `kbqDropdownSearch` with a nested `[kbqDropdownTriggerFor]` item (none of the new hosts does).

packages/components/dropdown/dropdown.component.ts
```diff
@@ -540,12 +583,22 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
...
             default:
                 if (keyCode === UP_ARROW || keyCode === DOWN_ARROW) {
-                    this.keyManager.setFocusOrigin('keyboard');
+                    this.setFocusOrigin('keyboard');
                 }
 
                 this.keyManager.onKeydown(event);
+                this.scrollActiveItemIntoView();
 
                 return;
```
lskramarov, line 639: **`scrollActiveItemIntoView()` runs for every key that reaches this branch, but only Up/Down move the highlight; the one non-key path that moves it never scrolls.**

In search mode every keydown the directive doesn't stop bubbles here: letters, Backspace, Delete, Shift, Ctrl, CapsLock. The search-mode manager has no typeahead, horizontal or Home/End handling, so for all of those `onKeydown` moves nothing and `scrollIntoView({ block: 'nearest' })` is called on an unchanged item. Repro: ArrowDown once, wheel the panel so the highlight is out of view, press Shift or type a letter → the panel snaps back to the highlighted item before the (debounced) filter even runs, and every printable character forces a synchronous layout.

Meanwhile the `items.changes → setFirstItemActive()` path (line 415) never scrolls: arrow deep into a long list, type a character that filters the highlighted item out, and the new first result is highlighted above `scrollTop`, so ENTER picks an item the user cannot see.

Scroll only when the highlight actually moved:

```suggestion
            default:
                if (keyCode === UP_ARROW || keyCode === DOWN_ARROW) {
                    this.setFocusOrigin('keyboard');
                    this.keyManager.onKeydown(event);
                    this.scrollActiveItemIntoView();
                } else {
                    this.keyManager.onKeydown(event);
                }

                return;
```

and call `this.scrollActiveItemIntoView()` after `setFirstItemActive()` at line 415 as well. (Subscribing to `keyManager.change` like the select does would also scroll on `mouseenter`, shifting the list under the pointer, and would need re-wiring on every manager swap.) The `scrollIntoView` mock in the spec is already in place and never asserted on: one test that a letter keydown does not scroll and ArrowDown / re-filter scroll once would pin this. After rebasing on main, the call itself should match the reveal policy #1972 introduced (`kbqFocusAndReveal`: `scrollIntoView?.({ block: 'nearest', inline: 'nearest' })`).

packages/components/dropdown/dropdown-search.ts
```diff
@@ -0,0 +1,105 @@
...
+                    return formField.control();
+                },
+                get keydownTarget() {
+                    return formField.elementRef.nativeElement;
+                },
+                clearByEscape: false
+            };
+        })
+    ],
+    host: {
```
lskramarov, line 29: **The search-mode highlight is invisible to assistive technology.**

DOM focus stays in the input while ArrowUp/Down move a highlight that exists only as `[class.kbq-dropdown-item_active]`; the input gets no `role`/`aria-controls`/`aria-activedescendant`, items have no `id`/`role`, the panel has no `role`/`id`. A screen-reader user hears nothing per arrow press and ENTER activates an item they were never told about. Before this PR the example's input stopped every keydown, so Tab reached the items with real focus; now Tab goes through `tabOut` and closes the panel, so the silent highlight is the only route from the field to the items. AGENTS.md: "It MUST follow all WCAG AA minimums, including focus management, color contrast, and ARIA attributes."

For context, select/autocomplete use `ActiveDescendantKeyManager` without `aria-activedescendant` too, so this extends a library-wide gap rather than deviating from it, but this PR is the one documenting the highlight as the keyboard model. A minimal wiring would be: a stable `id` and a role on `KbqDropdownItem`, a role/`id` on the panel, and `role="combobox"` + `aria-expanded` + `aria-controls` + `[attr.aria-activedescendant]` written onto the `kbqInput` host by this directive from the panel's active item (cleared in `resetActiveItem()`), plus an `expect(await axe(overlayContainerElement)).toHaveNoViolations()` case in the new describe, as `app-switcher.spec.ts:1795-1807` does for its search. Note the trigger deliberately carries no `aria-haspopup` because the panel has no menu semantics, so the role choice is a design decision.

packages/components/dropdown/dropdown.component.ts
```diff
@@ -658,9 +715,47 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
...
+    }
+
+    /**
+     * With a search field the caret has to stay in the input, so the active item is only highlighted and
+     * typeahead is left off — it would race the query the user is typing.
+     */
+    private initKeyManager(): void {
+        this.keyManagerInSearchMode = this.inSearchMode();
+
+        this.keyManager = this.keyManagerInSearchMode
```
lskramarov: **Every swap leaks the previous key manager; with `kbqDropdownContent` that is two per open/close cycle.**

The fork's `ListKeyManager` subscribes to `_items.changes` in its constructor and has no `destroy()` (`core/a11y/key-manager/list-key-manager.ts:63-76`); only `tabSubscription` is replaced here. With lazy content holding the search field (the `LazySearchDropdown` host), the directive is created on every open and destroyed after every close, so the effect re-runs `initKeyManager()` twice per cycle: after N cycles, 2N orphaned managers (each with its `tabOut`/`change`/`letterKeyStream` Subjects, the Focus ones also with the debounced typeahead pipeline) stay reachable from `directDescendantItems.changes` until `ngOnDestroy`. The `FocusKeyManager` built at close is a throwaway: the panel DOM is already gone and the next open replaces it. An `@if`-toggled search while the panel is open additionally strands a manager that still pins a destroyed item, and leaves that item's `.kbq-dropdown-item_active` on (as Copilot noted below).

Cheapest fix in this file: one instance per mode and only switch the reference and the `tabOut` subscription, clearing the outgoing manager's highlight on the way:

```ts
private activeDescendantKeyManager?: ActiveDescendantKeyManager<KbqDropdownItem>;
private focusKeyManager?: FocusKeyManager<KbqDropdownItem>;

private initKeyManager(): void {
    this.keyManagerInSearchMode = this.inSearchMode();

    this.keyManager?.setActiveItem(-1);
    this.keyManager = this.keyManagerInSearchMode
        ? (this.activeDescendantKeyManager ??= new ActiveDescendantKeyManager<KbqDropdownItem>(this.directDescendantItems))
        : (this.focusKeyManager ??= new FocusKeyManager<KbqDropdownItem>(this.directDescendantItems).withTypeAhead());
    ...
}
```

The general alternative is a `destroy()` on the forked `ListKeyManager` (store the `_items.changes` subscription, unsubscribe it and the typeahead one), called here and in `ngOnDestroy`; that touches core and its API snapshot.

packages/components/dropdown/dropdown-search.ts
```diff
@@ -0,0 +1,105 @@
...
+        '(click)': 'handleClick($event)',
+        '(keydown)': 'handleKeydown($event)'
+    },
+    exportAs: 'kbqDropdownSearch'
+})
+export class KbqDropdownSearch implements AfterContentInit {
```
lskramarov, line 35: **This is `KbqSelectSearch` (`core/select/common.ts:63-141`, exported from `@koobiq/components/core`) member for member.**

`ngControl`, `focus()`, `reset()`, `value()`, both `ngAfterContentInit` guards and the ESC-clears-then-`stopPropagation` + `[SPACE, HOME, END]` keydown contract are the same bodies; what is new here is LEFT/RIGHT in the stop-list, the `(click)` stop, the two providers, and two exported throw helpers (`dropdown-errors.ts:37-56`) that share a byte-identical example block and become permanent public API where the select copy uses inline `throw Error(...)`. The two copies already drift (stopped keys), and the next ESC/IME/caret fix has to be discovered and applied twice.

Sharing is not free, which is why this is a question rather than a suggestion: the core copy is pinned to the deprecated `KBQ_FORM_FIELD_REF` because `core` cannot import `form-field`, and `extends KbqSelectSearch` would drag in `changes`, `isSearchChanged`, `setPlaceholder`/`hasPlaceholder`, the `valueChanges` subscription, `value(): unknown` and a public `handleKeydown` (re-declaring `(keydown)` in the subclass host would also register a second listener). Would a small base in core (accessor + guards + a `protected stoppedKeys` list) that both directives extend be acceptable, or is the intent to let the dropdown variant diverge? Either way the two throw helpers could collapse to one before they ship as public API.

packages/components/dropdown/dropdown-search.ts
```diff
@@ -0,0 +1,105 @@
...
+    private readonly formField = inject(KBQ_FORM_FIELD);
+
+    /** The form control the query is bound to. */
+    get ngControl() {
+        return this.formField.control().ngControl;
+    }
+
+    ngAfterContentInit(): void {
+        if (this.formField.control().controlType !== 'input') {
```
lskramarov, line 51: **Neither of these two throw paths is exercised by a spec.**

All five search hosts bind `[formControl]` and use a plain `kbqInput`, so `grep` finds no test for either message, while both helpers are exported public API (`dropdown.api.md:391,394`) and this is the directive's only runtime validation. Two tests in the existing `toThrow` style (see lines 634-646 of the spec) would do, with two extra hosts copied from `SearchWithoutCleanerDropdown`:

```ts
it('should throw when the search input is not bound to a form control', () => {
    const fixture = createComponent(SearchWithoutControlDropdown); // <input kbqInput />

    expect(() => fixture.detectChanges()).toThrow(/must be bound to a form control/);
});

it('should throw when the search field does not host a plain kbqInput', () => {
    const fixture = createComponent(SearchWithNumberInputDropdown); // <input kbqInput kbqNumberInput [formControl]="control" />

    expect(() => fixture.detectChanges()).toThrow(/containing an input\[kbqInput\]/);
});
```

packages/components/dropdown/dropdown.component.ts
```diff
@@ -540,12 +583,22 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
                 }
 
                 break;
+            case ENTER:
+                // The highlighted item never sees the key event itself, so replay it as a click and let
+                // consumer handlers and nested triggers behave exactly as they do for the mouse.
+                if (this.inSearchMode() && this.keyManager.activeItem) {
+                    event.preventDefault();
+                    this.keyManager.activeItem.getHostElement().click();
+                }
```
lskramarov, line 625: **The replayed click can close the dropdown while this keydown is still bubbling, and the branch returns before `stopPropagation()`.**

`click()` → panel `(click)="close()"` → `closed.emit` → trigger `destroy()` → `overlayRef.detach()`, all synchronous, which also removes this overlay from the CDK `OverlayKeyboardDispatcher`. The ENTER keydown then reaches `body` and the dispatcher hands it to the next overlay with `keydownEvents()` observers (a modal, sidepanel, actions panel, or a consumer overlay), which is exactly what the comment at line 606 guards against for ESC. Every in-repo subscriber filters for ESCAPE, so nothing breaks today; a consumer reacting to ENTER (e.g. submit-on-enter in a dialog hosting the dropdown, `KbqActionsPanelRef.keydownEvents` is public) would fire on the selection keystroke. Pre-PR the activation was the browser's default action after dispatch, so the overlay was still registered when the key hit `body`.

Stopping the keydown inside this branch keeps non-search ENTER untouched:

```suggestion
                if (this.inSearchMode() && this.keyManager.activeItem) {
                    event.preventDefault();
                    // The replayed click can close the dropdown synchronously, unregistering the overlay
                    // from the CDK keyboard dispatcher before this keydown reaches `body`.
                    event.stopPropagation();
                    this.keyManager.activeItem.getHostElement().click();
                }
```

packages/components/dropdown/_dropdown-theme.scss
```diff
@@ -44,6 +47,7 @@
         &.kbq-selected {
             @include _kbq-dropdown-item-state(states-selected);
 
+            &.kbq-dropdown-item_active,
```
lskramarov: **`_active` is missing the `:not(.kbq-disabled)` guard its `.cdk-focused` twin carries.**

The key manager only refuses to *move onto* disabled items; an item that becomes disabled while it is the active one keeps `.kbq-dropdown-item_active` (`disabled` is a plain setter, nothing evicts it). Compiled, `.kbq-dropdown-item.kbq-selected.kbq-dropdown-item_active` is (0,3,0) and outranks `.kbq-dropdown-item.kbq-disabled` (0,2,0), so a selected item disabled while highlighted paints `states-selected-hover` instead of `states-disabled`. The plain rule at line 28 ties at (0,2,0) and only loses to `.kbq-disabled` by source order, so it deserves the same guard for symmetry.

```suggestion
            &.kbq-dropdown-item_active:not(.kbq-disabled),
```

(and `&.kbq-dropdown-item_active:not(.kbq-disabled),` at line 28). Unrelated but worth knowing while touching this file: the safe-area rule at lines 33-39 nests `.kbq-dropdown__panel_safe-area-active` *inside* `.kbq-dropdown-item`, so it compiles to a descendant selector that can never match the panel ancestor; it is inert today, and if it is ever fixed, `_active` needs to be listed there next to `.cdk-focused`.

packages/components/dropdown/dropdown-search.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (event.keyCode === ESCAPE && this.value()) {
+            this.reset();
+            event.stopPropagation();
+
+            return;
+        }
+
+        // Keys the panel would otherwise act on while the caret is in the field. Vertical arrows,
+        // ENTER and TAB are deliberately left to the panel.
+        if ([SPACE, HOME, END, LEFT_ARROW, RIGHT_ARROW].includes(event.keyCode)) {
```
lskramarov: **LEFT is swallowed unconditionally, so a search field inside a nested panel loses its close key.**

The panel closes a submenu on LEFT (`handleKeydown`, lines 574-578), but that branch is unreachable from a search field because the keydown is stopped here at the form-field host; ESC still works, though it takes two presses with a non-empty query. Nothing forbids `kbqDropdownSearch` in a nested panel and nothing tests it. `KbqNavbar` handles the same tension by letting the horizontal arrow through only when the caret cannot move any further (`navbar.component.ts:352-355, 365-369`, `input.selectionStart === 0`); the top-level panel ignores LEFT/RIGHT anyway (`this.parent` guard), so a caret-at-start check here costs nothing there. The RTL mirror (RIGHT at caret end) can follow separately. Note the existing test dispatches LEFT on an empty input, so it would need a value first.

While here: SPACE, HOME and END in this list protect nothing for this panel. `handleKeydown` has no case for them and the fork's manager returns without `preventDefault()` for HOME/END unless `withHomeAndEnd()` (never called), while SPACE only feeds a `letterKeyStream` nobody subscribes to in search mode. The select stops them because it really acts on them; here they are dead code, and the test at spec:2648 only asserts that a panel spy is not called.

packages/components/dropdown/dropdown-search.ts
```diff
@@ -0,0 +1,105 @@
...
+            // Merged, not replaced: the field still inherits whatever defaults the app configures.
+            provide: KBQ_FORM_FIELD_DEFAULT_OPTIONS,
+            useFactory: () => ({
+                ...inject(KBQ_FORM_FIELD_DEFAULT_OPTIONS, { optional: true, skipSelf: true }),
+                noBorders: true,
+                inOverlay: true
+            })
+        },
+        // The panel owns ESCAPE: the first press clears the query, the second closes the dropdown.
+        kbqCleanerFactoryProvider(() => {
```
lskramarov: **This provider is a copy of `KbqFormField`'s own (`form-field.ts:118-130`) with only `clearByEscape` flipped, and the flip changes nothing observable.**

On ESC the coalesced host `(keydown)` of this directive runs before the cleaner's `fromEvent` listener on the same element, and `ngControl.reset()` writes `''` into the input synchronously, so by the time `KbqCleaner.onKeyDown` runs, `canShow` is already false and `clearByEscape: true` would not clear anything either; on the second press both are no-ops and the panel closes. (`KbqSelectSearch` ships without such an override; its `canCleanerClearByEsc = false` targets a deprecated, unused property.)

The cost is fourteen lines that must track `KbqCleanerContext` by hand (the copy already lacks the optional `clear?()` hook), the `kbqCleanerFactoryProvider` import, and the DI-only test at spec:2830-2836 that pins the implementation. Deleting lines 30-44 (and that test plus its `KBQ_CLEANER_CONTEXT` import) leaves the two behavioural ESC tests (2663-2682 with a cleaner, 2796-2810 without) passing.

packages/components/dropdown/dropdown.component.ts
```diff
@@ -352,20 +369,27 @@ export class KbqDropdown implements AfterContentInit, KbqDropdownPanel, OnInit,
...
+        // The search field can show up long after content init — rendered by `kbqDropdownContent` when
+        // the panel first opens, or revealed by a structural directive — so the key manager follows the
+        // mode instead of being decided once.
+        effect(() => {
+            const inSearchMode = this.inSearchMode();
+
+            if (this.keyManager && this.keyManagerInSearchMode !== inSearchMode) {
+                this.initKeyManager();
+            }
+        });
```
lskramarov, line 401: **Three loose ends in this effect: a flip while the panel is open never re-applies focus, `navigationWithWrap` leaks in as a dependency, and the mode is tracked twice.**

The comment above promises support for a search field "revealed by a structural directive", but the effect only rebuilds the manager; `applyInitialFocus` is reachable only from the trigger's `init()` at open time. Keyboard-open a normal dropdown (first item DOM-focused), then reveal the search with `@if`: the new `ActiveDescendantKeyManager` starts at -1 and the field is never focused, so ArrowDown twice lights item 2 through `_active` while item 1 keeps `.cdk-focused` (two hover-styled rows), and typed letters go to the focused button instead of the query. Remove the search while open and the focused input is destroyed, focus falls to `body`, and the panel's `(keydown)` never fires (ESC and arrows dead until TAB or a click). The `ConditionalSearchDropdown` spec only flips `showSearch` before `open()`. When `panelAnimationState === 'enter'`, calling `resetActiveItem()` on the outgoing manager before the swap and `applyInitialFocus(this.focusOrigin)` after it would hand focus over the same way opening does (deferred via `ngZone.onStable` if focusing inside the effect trips NG0100 in dev mode).

`initKeyManager()` also reads `this.navigationWithWrap()` (line 732) inside the effect without `untracked`, so after the first flip the input becomes an accidental dependency: a runtime change re-runs the effect, which compares modes and does nothing, while the live manager keeps its old wrap setting until the next rebuild (every open/close for lazy content, never for static content, which matches pre-PR). `untracked(() => this.initKeyManager())` keeps the dependency to exactly `inSearchMode`, as the comment states; a separate `effect(() => this.keyManager?.withWrap(this.navigationWithWrap()))` if the input is meant to be live.

Smaller: `keyManagerInSearchMode` is only ever written in `initKeyManager()` in lock-step with the class chosen at line 728, and the file already discriminates with `instanceof FocusKeyManager` in `setFocusOrigin`. The two managers are sibling subclasses, so the check is exact and the field, its JSDoc and lines 726-727 can go:

```suggestion
        effect(() => {
            const inSearchMode = this.inSearchMode();
            const keyManagerInSearchMode = this.keyManager instanceof ActiveDescendantKeyManager;

            if (this.keyManager && keyManagerInSearchMode !== inSearchMode) {
                this.initKeyManager();
            }
        });
```

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -2515,6 +2525,316 @@ describe('KbqDropdown', () => {
...
+
+        beforeEach(() => {
+            originalScrollIntoView = Element.prototype.scrollIntoView;
+            Element.prototype.scrollIntoView = scrollIntoView;
+        });
+
+        afterEach(() => {
+            Element.prototype.scrollIntoView = originalScrollIntoView;
+        });
+
```
lskramarov: **This stub duplicates the global polyfill and is never asserted on.**

`tools/jest/setup.ts:69-71` already installs `Element.prototype.scrollIntoView = jest.fn()` for every spec (via `setupFilesAfterEnv`, with `clearMocks: true`), so the comment does not hold here and the swap replaces one no-op mock with another; `expect(scrollIntoView` appears nowhere in the file, so `scrollActiveItemIntoView()` could be deleted from the component without failing a test. Since the property is already a function, `jest.spyOn(Element.prototype, 'scrollIntoView')` works (the pattern `inline-edit.spec.ts:892` uses); one assertion in the arrow-key test, `expect(scrollIntoView).toHaveBeenCalledWith({ block: 'nearest' })`, plus one that a typed letter does not scroll, would cover the behaviour this PR adds.

```suggestion
```

packages/components/dropdown/dropdown.spec.ts
```diff
@@ -2515,6 +2525,316 @@ describe('KbqDropdown', () => {
...
+            withDisabled.detectChanges();
+
+            expect(items[1].classList).not.toContain('kbq-dropdown-item_active');
+            expect(getActiveItem()).toBe('Enabled');
+
+            dispatchKeyboardEvent(getInput(), 'keydown', ENTER);
+            withDisabled.detectChanges();
+            flush();
+
+            expect(withDisabled.componentInstance.clicked).toBe(false);
```
lskramarov, line 2778: **This assertion cannot fail on its own.**

`clicked` is only set by the disabled `<a>`, and line 2757 already proved the active item is `Enabled`, so ENTER can only click `Enabled` (no handler); an implementation where ENTER does nothing at all passes unchanged. The observable effect, `Enabled` clicked and its click bubbling to `(click)="close()"`, is what the sibling test at 2606-2611 asserts:

```suggestion
            expect(withDisabled.componentInstance.clicked).toBe(false);
            expect(overlayContainerElement.querySelector(PANEL_SELECTOR)).toBeFalsy();
```

Small related nit: the title at line 2648 says "SPACE and the horizontal arrows" while the body also dispatches HOME and END.

packages/components/dropdown/dropdown.component.ts
artembelik: В `dbe5acc`: панель отдаёт каретку через `KbqDropdown.restoreFocus()`, вызывается из `KbqDropdownTrigger.focus()`; подменю панели с поиском больше не забирает фокус при открытии мышью.

packages/components/dropdown/dropdown-search.ts
artembelik: Зеркало для RIGHT в конце строки сделал сразу, SPACE/HOME/END из стоп-листа убрал.

packages/components/dropdown/dropdown-search.ts
artembelik: не буду добавлять новые сущности в рамках бага, создам отдельную задачу на это - DS-5488

packages/components/dropdown/dropdown-search.ts
artembelik: реализуем отдельно DS-5489

packages/components/dropdown/dropdown-search.ts
artembelik: Провайдер удалён в `1f33411`: проверил на примере с `kbq-cleaner` в браузере и обоими тестами на ESCAPE - без него поведение то же, обработчик директивы успевает очистить поле раньше слушателя клинера.

NikGurev (approved):

lskramarov (approved):

PR #1977, opened 2026-09-02
fix(tags): move validation to kbq-tag-list (#DS-2943)

artembelik:


packages/components/tags/tag-list.component.ts
```diff
@@ -565,13 +565,6 @@ export class KbqTagList
     registerInput(inputElement: KbqTagTextControl): void {
         this.tagInput = inputElement;
         this.registeredInput.set(inputElement);
```
lskramarov, line 574: **The `validation-tag-list` docs example is silently gutted by this change.**

It relies on both mechanisms removed here — the `ngControl.invalid` gate in `emitTagEnd()` and this `statusChanges` → `setErrors` bridge:

```ts
// packages/docs-examples/components/validation/validation-tag-list/validation-tag-list-example.ts:75
tagInputFormControl: this.formBuilder.control('', [Validators.pattern('[a-zA-Z]*')])
```

bound via `formControlName="tagInputFormControl"` on the `<input kbqTagInputFor>` (line 40), next to `<kbq-hint>Provide only latin letters</kbq-hint>` (line 52).

That example is rendered on the live Validation page (`packages/components/core/validation/examples.validation.en.md:45`, `.ru.md:50`) — the same page the four new doc sections here link to as `[ErrorStateMatcher](/en/other/validation)`.

**After this PR:** typing `тег` + Enter creates the tag, no error ever surfaces, and because the handler clears the DOM value with `input.value = ''` (which never reaches the `DefaultValueAccessor`), `tagInputFormControl` stays `pattern`-invalid for the rest of the session — so `reactiveForm.invalid` is stuck `true` with nothing shown anywhere.

The example needs to be migrated onto the tag-list-level pattern this PR documents. The same stale wiring is also left in `packages/components-dev/validation/module.ts:157`.

packages/components/tags/tag-input.ts
```diff
@@ -128,6 +127,9 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
     private trimDirective = inject(KbqTrim, { optional: true, self: true });
     /**
      * The form control instance bound to the input, if any.
+     *
+     * @deprecated Unused. Bind `[formControl]`/`[ngModel]` to `<kbq-tag-list>` instead — it is the
```
lskramarov: **This deprecation (and the matching prose in all four `.md` files) is unconditional, but the binding it tells people to remove is load-bearing for the autocomplete pattern.**

`KbqAutocompleteTrigger` registers itself as `NG_VALUE_ACCESSOR` for `input[kbqAutocomplete]` (`autocomplete-trigger.directive.ts:99-103`, `:119`) and pushes picked options through it (`this.onChange(event.source.value)`, `:595`). Six shipped examples depend on that:

- `tag-autocomplete-{overview,search,removable,draggable,editable}-example.ts` — `[(ngModel)]="tagInputModel"`
- `tag-autocomplete-option-operations-example.ts:56` — `[formControl]="control"`

A consumer who follows `Bind [formControl]/[ngModel] to <kbq-tag-list> instead` loses `valueChanges` on the input, so `onControlValueChanges` stops running and panel filtering, `canCreate` and `hasDuplicates` all break.

Only the **validators** belong on `<kbq-tag-list>`. Worth qualifying both this JSDoc and `tag-autocomplete.{en,ru}.md:88` ("Bind `[formControl]`, `[formControlName]` or `[ngModel]` to `<kbq-tag-list>`") accordingly — `tools/public_api_guard/components/tags.api.md:191` now publishes the member as `// @deprecated`, which reads as "this whole thing is going away".

packages/components/tags/tag-input.ts
```diff
@@ -414,12 +386,25 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
 
-    private trimValue(value) {
-        return this.trimDirective ? this.trimDirective.trim(value) : value;
+    /** Single path for adding a tag, shared by typed input, blur and paste. */
+    private addTag(value: string): void {
+        if (this.distinct() && this.isDuplicate(value)) return;
+
+        // An empty value adds nothing, so latching the flag here would make the next
+        // programmatic tags update look UI-initiated.
+        if (value) {
```
lskramarov, line 410: **The `pendingUIChange` latch still leaks — for the exact pattern this PR's new docs recommend.**

The flag is set here, *before* the emit, and is cleared in exactly one place — `tag-list.component.ts:517-519`, inside the `tags.changes` subscription. There is no other reset path.

The new prose says (`tag-input.en.md:97`): *"To keep an invalid value out of the list, filter it in the `(kbqTagInputTokenEnd)` handler."* A handler that filters produces no `tags.changes`, so the latch stays `true` indefinitely.

**Scenario:** user types `Вирус` + Enter → flag latched → handler rejects → nothing added, nothing clears the flag. Later, a purely programmatic `formControl.setValue(['x', 'y'])` re-renders the `@for`, `tags.changes` fires, the stale latch is consumed, and `propagateTagsChanges()` (`tag-list.component.ts:848-856`) runs `change.emit()`, `valueChange.emit()` and `onChange(valueToEmit)` — which routes through Angular's `setUpViewChangePipeline` → `markAsDirty()`. The control is now dirty from an edit the user never made, so `ShowOnControlDirtyErrorStateMatcher` consumers start showing errors.

`tag-list.component.spec.ts:943` (`should not set the control to dirty when rendered tags changed programmatically`) passes only because no tag was ever rejected first — adding one rejected `emitTagEnd()` before the mutation flips it red.

The comment above shows the empty-value branch was reasoned about; the rejected-value branch is the larger one. A correlated token that expires within the same microtask (or comparing `tags.map(t => t.value)` against the last-propagated `_value` and dropping the flag entirely) would close both.

packages/docs-examples/components/tags/tag-autocomplete-with-form-control-validators/tag-autocomplete-with-form-control-validators-example.ts
```diff
@@ -0,0 +1,165 @@
...
+                    </kbq-option>
+                }
+            </kbq-autocomplete>
+
+            <kbq-hint>Only latin letters, up to {{ maxTagCount }} tags</kbq-hint>
+
+            <kbq-error>
+                @if (formControl.hasError('required')) {
+                    Field is required
+                } @else if (formControl.getError('maxTagCount'); as error) {
```
lskramarov: **`@else if` hides one of two errors that can be set at the same time — in the example that exists to teach multi-validator usage.**

`maxTagCountValidator` and `latinValidator` are independent. Using only this example's own UI: pick `Вирус` from `OPTIONS` (line 19) so `latin` fires, then pick `Trojan`, `Botnet`, `Rootkit` → 7 tags → **both** errors are set, but only "No more than 5 tags" renders. Remove two tags and "Invalid tags: Вирус" appears for the first time — whack-a-mole.

(`required` genuinely cannot co-occur, since it only fires on `[]` where both custom validators return `null`, so the first `@else` link is fine.)

The pre-PR template used two independent `@if` blocks; the chaining was introduced at the same time as the second independent validator. No existing example in the repo chains `@else if` across errors that can co-occur — `validation-basic-multiple-validators` chains only mutually exclusive `required`/`minlength`/`maxlength`. The house shape for this case is one `<kbq-error>` per condition, as in `timepicker-field-validation`.

Same block at `tag-input-with-form-control-validators-example.ts:68`.

packages/components/tags/tag-input.ts
```diff
@@ -414,12 +386,25 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
+            this._tagList?.notifyPendingTagChange();
+        }
+
+        this.tagEnd.emit({ input: this.inputElement, value });
     }
 
-    private hasControl(): boolean {
-        return !!this.ngControl;
+    private isDuplicate(value: string): boolean {
```
lskramarov: **`[distinct]="true"` still cannot dedupe *within* a single paste, and the refactor now makes it look as if it can.**

`tags` is a `@ContentChildren` `QueryList` that only refreshes during change detection. `onPaste` runs `dividedString.forEach((item) => this.addTag(...))` (line 361) entirely inside one event handler; `output().emit()` invokes listeners synchronously and `markForCheck` only schedules — so no CD runs between iterations and `_results` is literally the same array every time.

**Scenario:** empty list, `[distinct]="true"`, `COMMA` separator, paste `"a,a"` → both `isDuplicate('a')` calls return `false` → two identical tags.

`main` had the same hole (a single hoisted snapshot filter), so this is **not a regression** — but moving to a per-item live-list read reads like the guarantee is now stronger, and nothing tests it. `should filter duplicate values on paste when distinct=true` (`tag-input.spec.ts:286`) can't catch it: that host's `add` is a bare `jest.fn()` that never appends, so the list is constant by construction.

Deduping against the authoritative value (or a `Set` built once in `onPaste` and updated as items are accepted) would close it — and would also fix the efficiency point below.

lskramarov (commented):
## Code review — `fix(tags): stop bypassing validation on paste` (#DS-2943)

The core direction is right: the input-level `NgControl` gate and the `registerInput()` → `setErrors` bridge were genuinely broken (that bridge also wiped the tag list's own errors whenever the input control happened to be valid), and consolidating typed/blur/paste into one `addTag()` is a real improvement. The `onPaste` empty-data guard fix is correct — `(data && data.length === 0)` was dead code. I also ran the full gate set on this branch: jest (tags/form-field/autocomplete, 485 passed), eslint, stylelint, prettier, cspell, `ngc --noEmit` on docs-examples, and the example-module generator all pass, and the regenerated `example-module.ts` is byte-identical to what the generator produces.

15 findings below, most severe first.

**Blocking-ish (consumer-visible):**

1. `validation-tag-list` — a **shipped docs example** relies on both removed mechanisms and is now a no-op demo whose control is permanently invalid.
2. The `ngControl` deprecation and the new prose tell people to move `[formControl]`/`[ngModel]` off the input **unconditionally** — but that binding is the autocomplete's `NG_VALUE_ACCESSOR` in six examples.
3. The `pendingUIChange` latch still leaks for rejected non-empty values — the exact pattern the new docs recommend — spuriously marking the control dirty on a later programmatic `setValue`.
4. `@else if` in both validator examples hides co-occurring errors, reachable with the example's own options.

**Correctness (pre-existing, but in code this PR rewrites):** `[distinct]` still can't dedupe within a paste; `blur()` has no `disabled` guard.

**Process:** no `BREAKING CHANGE:` footer, no `docs/guides/migration.{en,ru}.md` entry, no schematic — `fix(popover)` (`5e7db141a`, same series) shipped all three for a smaller change. And the commit subject describes the opposite of what changed, which is what lands in the changelog.

**Tests:** 8 of the 12 new tests pass unchanged on `main`.

**Docs:** the stated default `ErrorStateMatcher` behaviour omits `form.submitted`; "regardless of its source" is not true for in-place tag editing (`preventEditSubmit` is the real hook and goes unmentioned).

Plus efficiency, reuse and duplicate-coverage notes inline.

packages/components/tags/tag-input.ts
```diff
@@ -311,38 +311,26 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
-
-        (this.ngControl.statusChanges as EventEmitter<string | null>).emit(this.ngControl.status);
-    }
+    triggerValidation(): void {}
 
     /**
      * Checks to see if the (tagEnd) event needs to be emitted.
      * @docs-private
      */
     emitTagEnd(): void {
```
lskramarov, line 328: **`blur()` reaches `emitTagEnd()` with no `disabled` guard, so disabling a focused tag input appends whatever was typed.**

Verified in a real browser: Chromium/Edge fire `blur` synchronously when an element gets `disabled` (WebKit does not — so this is Chromium-only).

**Scenario:** user types `abc` without pressing Enter; the app calls `formControl.disable()` → `setDisabledState` → `KbqTagList.disabled` → the `KbqTagInput.disabled` getter (line 239) → the `[attr.disabled]` host binding (line 113) lands on the `<input>` → browser fires `blur` → `addOnBlur` defaults to `true`, and the autocomplete's default `onInputBlur` predicate returns `true` for a `null` `relatedTarget` → `emitTagEnd()` → a tag is appended to a control that was just disabled, and `notifyPendingTagChange()` latches.

`KbqTagList.blur()` already guards its own side effects with `if (!this.disabled)` (`tag-list.component.ts:738`); `KbqTagInput.blur()` does not.

Pre-existing (a disabled control has `invalid === false`, so `main`'s gate did not stop it either), but `blur()` is a method this PR edits, so it seems like the moment to add the guard.

packages/components/tags/tag-input.spec.ts
```diff
@@ -698,6 +725,172 @@ describe(KbqTagInput.name, () => {
...
+                expect(componentInstance.inputControl.invalid).toBe(true);
+
+                componentInstance.tagInput().onKeydown(createKeyboardEvent('keydown', ENTER, inputElement, 'Enter'));
+
+                expect(componentInstance.add).toHaveBeenCalledWith(
+                    expect.objectContaining({ value: 'NOT-LATIN-LOWERCASE' })
+                );
+            });
+
+            it('should not block (tagEnd) on paste when the input control is invalid', () => {
```
lskramarov, line 829: **This test passes unchanged on `main` — it pins nothing.**

`git show main:packages/components/tags/tag-input.ts` shows `onPaste` emitting `tagEnd` directly; the `!this.hasControl() || !this.ngControl.invalid` gate existed **only** in `emitTagEnd()`. So paste never consulted `ngControl` in the first place, and this test — which sits in `describe('with a deprecated form control bound to the input')` and reads as the regression guard for the headline fix — is green before and after.

Same for all seven tests in `describe('with validators on <kbq-tag-list>')` (line 730): that host binds no `[formControl]` to the `<input>`, so `hasControl()` was already `false` and `registerInput()`'s `if (inputElement.ngControl?.statusChanges)` never subscribed.

**Net: 8 of the 12 new tests are green on `main`.** Only three actually fail there — the ENTER gate (`:829`), the `setErrors` overwrite (`:861`) and `triggerValidation()` (`:882`). Worth either dropping the vacuous ones or reshaping them so they fail against the pre-fix code.

packages/components/tags/tag-list.component.ts
```diff
@@ -565,13 +565,6 @@ export class KbqTagList
     registerInput(inputElement: KbqTagTextControl): void {
```
lskramarov, line 572: **Two consumer-visible behaviours are removed with no `BREAKING CHANGE:` footer, no migration entry and no schematic.**

```
$ git diff --name-only main...HEAD | grep -Ei 'migration|schematic'
(nothing)
```

and neither commit body carries a footer.

The convention here looks well established — `5e7db141a` (`fix(popover)`, same review series) touched `docs/guides/migration.en.md +20`, `migration.ru.md +20`, `collection.json +5`, `migrations.json +5`, and its migration prose is written for exactly this shape: *"That is the one change here with no compile error behind it."* `migration.en.md:1038` lists the components reviewed in 20.3.0 and would need `tags` added.

**Who breaks:** a consumer with `<input kbqTagInputFor [formControl]="ctrl">` plus any validator, upgrading on a patch bump. Values their validator used to reject now become tags, and their `<kbq-error>` stops rendering — no compile error, no `ng update` output, nothing. `[formControl]` on `input[kbqTagInputFor]` is a template binding a warn-only schematic can match exactly, the way `popover-leave-delay` does.

packages/components/tags/tag-input.ts
```diff
@@ -311,38 +311,26 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
-        if (!this.hasControl() || (this.hasControl() && !this.ngControl.invalid)) {
-            if (this.distinct() && this.hasDuplicates) return;
-
-            this._tagList?.notifyPendingTagChange();
-            this.tagEnd.emit({ input: this.inputElement, value: this.trimValue(this.inputElement.value) });
-        }
+        this.addTag(this.trimValue(this.inputElement.value));
```
lskramarov: **The commit subject describes the opposite of what changed, and it is what lands in the changelog.**

`main` is squash-merged, so `fix(tags): stop bypassing validation on paste` becomes the release-note line under Bug Fixes (`packages/cli/src/release/changelog-writer-options.ts`).

But paste already ignored the input-level gate before this PR — which is precisely why `should not block (tagEnd) on paste when the input control is invalid` is green on `main`. What actually changes is that **ENTER and blur stop validating**, with validation relocated to the `<kbq-tag-list>` control.

A consumer reading "stop bypassing validation on paste" upgrades expecting pasted values to start being rejected; instead their previously-blocked *typed* values start creating tags. Something like `fix(tags): move validation to <kbq-tag-list>` would describe it. Separate from the missing footer above — this one is a retitle.

packages/components/tags/tag-input.en.md
```diff
@@ -88,6 +88,18 @@ import { kbqTagsDefaultOptionsProvider } from '@koobiq/components/tags';
...
 
+### Validation
+
+Validators belong on `<kbq-tag-list>`, not on `<input kbqTagInputFor>`: the tag list is the form control for the whole set of tags, while the input is only a text entry helper. Bind `[formControl]`, `[formControlName]` or `[ngModel]` to `<kbq-tag-list>`.
+
+The control value is the array of tags, so a single set of validators covers both the number of tags and the content of each one.
+
+Validators do not prevent a tag from being added: the tag is added and the field goes into the error state — the same way for typing and for pasting from the clipboard. To keep an invalid value out of the list, filter it in the `(kbqTagInputTokenEnd)` handler, which is called for every tag being created regardless of its source.
+
+When the error becomes visible is controlled by [ErrorStateMatcher](/en/other/validation), and by default happens once the control is invalid and touched.
```
lskramarov: **The stated default drops half of the shipped matcher.**

```ts
// packages/components/core/error/error-state-matcher.ts:60
return !!(control?.invalid && (control.touched || form?.submitted));
```

A reader with a `<form [formGroup]>` who submits without ever focusing the tag list sees `<kbq-error>` appear and concludes the matcher is misbehaving — this sentence told them submit-triggered display requires `ShowOnFormSubmitErrorStateMatcher`, which the same line links to.

Same sentence in all four new sections: `tag-input.en.md:99`, `tag-input.ru.md:99`, `tag-autocomplete.en.md:94`, `tag-autocomplete.ru.md:94`.

packages/components/tags/tag-input.en.md
```diff
@@ -88,6 +88,18 @@ import { kbqTagsDefaultOptionsProvider } from '@koobiq/components/tags';
 })
 ```
 
+### Validation
+
+Validators belong on `<kbq-tag-list>`, not on `<input kbqTagInputFor>`: the tag list is the form control for the whole set of tags, while the input is only a text entry helper. Bind `[formControl]`, `[formControlName]` or `[ngModel]` to `<kbq-tag-list>`.
+
+The control value is the array of tags, so a single set of validators covers both the number of tags and the content of each one.
+
+Validators do not prevent a tag from being added: the tag is added and the field goes into the error state — the same way for typing and for pasting from the clipboard. To keep an invalid value out of the list, filter it in the `(kbqTagInputTokenEnd)` handler, which is called for every tag being created regardless of its source.
```
lskramarov: **"regardless of its source" is not accurate — in-place tag editing never routes through `(kbqTagInputTokenEnd)`.**

`tagEnd` has exactly one emission site (`tag-input.ts:399`), reachable only from `emitTagEnd()` and `onPaste`. Edits commit through `KbqTagList.listenToTagsEdit()` (`tag-list.component.ts:903-911`) straight into `propagateTagsChanges()`.

This same page documents editing 33 lines above (`### Editing`, `<!-- example(tag-input-editable) -->` at line 58). So: double-click a tag, retype it as `Вирус`, press Enter — the invalid value lands in the control value with no `tagEnd` in sight, and the absolute clause has told the reader to stop looking.

The library ships the right hook — `preventEditSubmit` (`tag.component.ts:289`, demonstrated in `tag-editable-with-validation-example.ts:24`) — and neither doc mentions it. `tag-autocomplete.en.md:90` drops the absolute clause entirely and is the wording to mirror here.

packages/components/tags/tag-input.spec.ts
```diff
@@ -971,3 +1164,74 @@ class TestTagInput {
...
+    template: `
+        <kbq-form-field>
+            <kbq-tag-list #tagList [formControl]="listControl">
+                @for (tag of listControl.value; track $index) {
+                    <kbq-tag [value]="tag">{{ tag }}</kbq-tag>
+                }
+                <input
+                    [formControl]="inputControl"
+                    [kbqTagInputFor]="tagList"
+                    [kbqTagInputAddOnBlur]="false"
```
lskramarov: **`[kbqTagInputAddOnBlur]="false"` means the blur path is never exercised with an input-level control.**

`addTag()`'s own comment claims it is *"shared by typed input, blur and paste"*, but this is the only host with a control on the `<input>`, and it pins `addOnBlur` off — so `componentInstance.tagInput().blur({} as FocusEvent)` at line 875 short-circuits at `if (this.addOnBlur && ...)` and never reaches `emitTagEnd()`.

Pre-PR, blur was gated by the same `!this.ngControl.invalid` check as ENTER. Restoring that gate inside `emitTagEnd()`, or adding any early return to `blur()`'s `addOnBlur` branch, would leave all twelve new tests green while blur-created tags are silently blocked again for anyone with a control on the input.

packages/components/tags/tag-input.ts
```diff
@@ -370,23 +358,7 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
-
-            items = items.filter((item) => !tagValues.includes(item));
-        }
-
-        if (items.length) {
-            this._tagList?.notifyPendingTagChange();
-        }
-
-        items.forEach((item) => this.tagEnd.emit({ input: this.inputElement, value: item }));
+        dividedString.forEach((item) => this.addTag(this.trimValue(item)));
```
lskramarov: **Dropping the hoisted snapshot turns paste dedupe into O(N×M) DOM reads.**

`main` computed the comparison set once per paste:

```ts
const tagValues: string[] = this._tagList.tags.map(({ value }) => value);
items = items.filter((item) => !tagValues.includes(item));
```

Now every pasted item re-scans the `QueryList`, and `KbqTag.value` is:

```ts
// tag.component.ts:350 — the documented default for a <kbq-tag> without [value]
get value(): any { return this._value ?? this.elementRef.nativeElement.textContent?.trim(); }
```

`QueryList.some` only short-circuits on a *match*, so the normal miss case walks all M. Pasting a 500-entry list against 500 unbound tags with `[distinct]="true"` is ~250k `textContent` serializations plus ~500k short-lived strings, synchronously, before `preventDefault()` runs — against 500 reads pre-PR.

Gated behind opt-in `distinct` and `[value]`-less tags, so it is tens of milliseconds rather than a freeze. The one-line fix — build a `Set` of existing values before the loop — also closes the within-paste dedupe gap noted on `isDuplicate`.

packages/docs-examples/components/tags/tag-input-with-form-control-validators/tag-input-with-form-control-validators-example.ts
```diff
@@ -11,17 +11,26 @@ import { KbqIconModule } from '@koobiq/components/icon';
...
-const customMaxLengthValidator = (max: number): ValidatorFn => {
-    return ({ value }: AbstractControl): ValidationErrors | null => {
-        if (!value) {
-            return null;
-        }
+const MAX_TAG_COUNT = 5;
+const LATIN_PATTERN = /^[a-zA-Z]+$/;
+
+/** Validates the number of tags — the control value is the whole tag array. */
+const maxTagCountValidator = (max: number): ValidatorFn => {
```
lskramarov: **This is `Validators.maxLength`, hand-rolled — in the document that teaches tag-list validation.**

```js
// node_modules/@angular/forms/fesm2022/forms.mjs:748
function maxLengthValidator(maxLength) {
    return (control) => {
        const length = control.value?.length ?? lengthOrSize(control.value);
        ...
```

Its own JSDoc: *"requires the number of items in the control's value to be less than or equal to the provided maximum length."* Behaviour is identical for `null`, `[]` and over-length arrays; only the error key differs (`maxlength: {requiredLength, actualLength}` vs `maxTagCount: {max}`). Both example files already import `Validators` for `Validators.required` on the same line, so the swap costs nothing.

Because this is documentation, the cost is not the ~15 duplicated lines but the implied lesson that the built-in does not work on tag arrays — and the repo does use built-ins on array controls elsewhere (`file-upload-multiple-required-reactive-validation-example.ts:53` uses `Validators.minLength(2)` on a `FormControl<FileList | KbqFileItem[]>`).

The three copies have already drifted: both examples return `{ maxTagCount: { max } }`, while `tag-input.spec.ts:1167` returns `{ maxTagCount: true }` with an untyped `AbstractControl` (implicit `any` on `value.length`).

`latinValidator` is genuinely custom — it reports `invalidTags` per item, which no built-in does — and should stay.

packages/components/tags/tag-input.spec.ts
```diff
@@ -698,6 +725,172 @@ describe(KbqTagInput.name, () => {
...
+
+                componentInstance.tagInput().onPaste(createPasteEvent('b'));
+                fixture.detectChanges();
+
+                expect(componentInstance.form.controls.tags.value).toEqual(['a', 'b']);
+                expect(componentInstance.form.controls.tags.valid).toBe(true);
+                expect(componentInstance.tagList().errorState).toBe(false);
+            });
+
+            describe('ErrorStateMatcher', () => {
```
lskramarov: **These three tests are already covered in `tag-list.component.spec.ts`, against a near-clone of the same host.**

`tag-list.component.spec.ts:2010-2093` has `describe('ErrorStateMatcher')` with:

- `describe(ErrorStateMatcher.name)` — "should not be in error state initially when invalid but untouched" (`:2012`), "should be in error state when invalid and touched" (`:2018`)
- `describe(ShowOnFormSubmitErrorStateMatcher.name)` — `:2053`, `:2063`

driven by `TagListWithErrorStateMatcher` (`:252`) — the same `<form [formGroup]>` + `formControlName` + `[errorStateMatcher]` + submit-button shape as the new `TestTagListValidation`, whose submit click even re-inlines that file's `getSubmitButton` helper.

The error-state pipeline lives entirely in `KbqTagList`/`KbqFormFieldControl` and is untouched here, and `ErrorStateMatcher` never observes *how* the value became invalid — so reaching it via `onPaste` adds no coverage.

Minor, in the same block: `expect(componentInstance.tagList().errorState).toBe(false)` at `:783` is entailed by the `valid === true` assertion on the line above and can never fail independently.

artembelik: поправил

lskramarov (approved):

NikGurev (approved):

PR #1978, opened 2026-09-02
fix(badge)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `badge`, in the same shape as the 20.3.0 component reviews.

### `badgeColor` never reported what was written

The input's setter took a color, its getter returned a CSS class:

```ts
badge.badgeColor = KbqBadgeColors.Error;
badge.badgeColor; // 'kbq-badge_error'
```

It is an `input()` now and reports the raw color. The `kbq-badge_<color>` class still lands on the host, from an internal `computed`, so **styles and screenshots are unchanged** — only a programmatic read sees the difference.

### `compact` and `outline` gained `booleanAttribute`

`<kbq-badge compact>` passed the empty string, which is falsy, so the attribute did nothing. It works now — and `[compact]="'false'"`, previously truthy, now means `false`.

### Dead code the review found

- **`iconItem`** was a `contentChild(KbqIconItem)` that nothing read — not the template, not the host bindings, not anywhere in the package. Removed, so every badge in a list or table stops paying for a discarded query.
- **`isIconButton`** was write-only. The block was copied from `KbqButtonCssStyler`, where it *is* host-bound (`[class.kbq-button]`); the badge binds no class to it, so the `cdr.detectChanges()` it guarded forced a re-entrant change-detection pass over the ancestor view from inside an effect, rendering nothing new. Removed along with the `ChangeDetectorRef` injection.

### Closed internals

`KbqBadgeCssStyler` is `@docs-private` and its remaining members (`icons`, `nativeElement`, `updateClassModifierForIcons`) are private. Nothing outside the component ever drove the icon spacing it applies; the classes it puts on the host are the contract.

## Migration

`badge-signals` runs from `ng update @koobiq/components@20`. It rewrites `compact` / `outline` reads to calls — on receivers typed `KbqBadge` and through template reference variables on `<kbq-badge>`, in external and inline templates — and reports the rest. `badgeColor` is deliberately **not** rewritten: appending `()` would compile and hand back a different string.

## Testing

- `badge.component.spec.ts`: 3 → 12 tests, covering the color fallback, the class swap on change, the raw-color read, the valueless-attribute case, the `'false'` string coercion the guide promises, and the leading-icon spacing branch — which had no coverage at all before.
- `badge-signals/index.spec.ts`: 17 tests — auto-fix, idempotence, receiver scoping, template refs, every warning, the template-only consumer, and the `--fix=false` path.
- Full `packages/components` (5002 tests) and full `packages/schematics` (451 tests) suites pass.
- `check-api` is in sync.

No e2e screenshots were regenerated: every e2e case binds its inputs, so nothing it renders changed.

## Review

Reviewed at `xhigh` effort; seven findings, six applied. The one left alone: the icon-spacing algorithm is duplicated across `badge`, `link` and `button`, and consolidating it would mean touching `button`, which is outside this diff and this owner.

BREAKING CHANGE


tools/public_api_guard/components/badge.api.md
```diff
@@ -9,38 +9,34 @@ import * as i1 from '@angular/cdk/a11y';
...
-// @public (undocumented)
+// @public
 export const badgeRightIconClassName = "kbq-badge-icon_right";
 
-// @public (undocumented)
+// @public
 export class KbqBadge {
+    readonly badgeColor: i0.InputSignal<string>;
+    protected readonly badgeColorClass: i0.Signal<string>;
+    readonly compact: i0.InputSignalWithTransform<boolean, unknown>;
```
lskramarov, line 21: **`check-public-api-any` will fail on this head — the baseline was not re-approved.**

`booleanAttribute` turned `compact`/`outline` into `InputSignalWithTransform<boolean, unknown>`, adding two `unknown`s, while the diff removes only one `any` (`readonly icons: i0.Signal<readonly any[]>`). Running the tool's own counting logic over this golden gives **badge = 2** against `tools/check-public-api-any/baseline.json` → `"badge": 1`.

`.github/workflows/api.yml` runs `yarn run check-public-api-any` on every `pull_request`, so the **API** job exits 1 with:

```
❌ The published type surface gained `any` / `unknown`:
  - badge: 1 → 2
```

Fix: `yarn run approve-public-api-any`.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+ * A `.ts` file is a badge consumer if it names any of the exported symbols, imports the package, or renders
+ * the element in an inline template — a component that only imports `KbqBadgeModule` names none of the types.
+ */
+function referencesBadge(content: string): boolean {
+    return /\bKbqBadge\w*\b/.test(content) || content.includes(BADGE_PACKAGE) || content.includes(`<${BADGE_ELEMENT}`);
+}
+
+export default function badgeSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **`fix` is `undefined` under `ng update`, so the migration never writes anything.**

`migrations.json` declares no `"schema"` for any of its entries, so `schema.json`'s `"fix": { "default": true }` never reaches the factory — `ng update` invokes it with `{}`. `fix` is `undefined`, `commit()` takes the else branch, and every consumer file is only logged as `[badge-signals] would update <path> (run with --fix to apply)`; `tree.overwrite` is never called.

So the advertised `ng update @koobiq/components@20` rewrites **zero** `compact`/`outline` reads. Only the manual `ng generate` path (which does resolve `collection.json`'s schema) auto-fixes — the two invocation modes behave oppositely, and `index.spec.ts` only ever exercises the `ng generate` one because it always passes `fix` explicitly.

Twelve siblings guard this, with a comment describing exactly this problem — `app-switcher-signals/index.ts:413`, `navbar-signals-and-aria/index.ts:596`:

```suggestion
        const { project } = options;
        // `ng update` invokes migrations with no options at all, and migrations.json declares no schema, so the
        // schema default never reaches us — applying the fix is the intended behaviour there.
        const fix = options.fix ?? true;
```

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+            edits.push({ start: node.getEnd(), end: rhs.getStart(sourceFile), text: '.set(' });
+            edits.push({ start: rhs.getEnd(), end: rhs.getEnd(), text: ')' });
+        }
+
+        return;
+    }
+
+    // Read (incl. optional chain `x?.compact`): append `()`.
+    edits.push({ start: node.getEnd(), end: node.getEnd(), text: '()' });
```
lskramarov, line 422: **Compound assignments are rewritten into unparseable TypeScript.**

The write branch above matches only `ts.SyntaxKind.EqualsToken`, so `||=`, `&&=`, `??=` and `+=` fall through to this read branch and get `()` appended to the assignment *target*.

Verified by running this file's own `collectReceivers`/`classifyAccess`:

```ts
class D { w(badge: KbqBadge) { badge.compact ||= true; } }
// →           badge.compact() ||= true;
```

`The left-hand side of an assignment expression must be a variable or a property access.` `tree.overwrite` commits it and the run reports `updated 1 file(s)`, so `ng update` hands the consumer a file that no longer parses. `--fix=false` won't reveal it either.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
```
lskramarov: **The ref pattern is unanchored and has no assignment guard, so it rewrites expressions that are not the badge.**

`\b` matches after a `.`, and the regex runs over the whole template text with no awareness of Angular scoping, of bindings vs. text, or of assignment. Executed against this exact function:

| input | output |
| --- | --- |
| `<kbq-badge #badge/>{{ item.badge.compact }}` | `{{ item.badge.compact() }}` |
| `@for (badge of rows; …) { {{ badge.outline }} }` | `{{ badge.outline() }}` |
| `<p>Set badge.compact to shrink it.</p>` | `…badge.compact()…` (page text) |
| `<!-- badge.outline was the old flag -->` | `badge.outline()` (comment) |
| `(click)="badge.compact = true"` | `(click)="badge.compact() = true"` |
| `@if (x) { <kbq-badge #badge/> }<o>{{ badge.compact }}</o>` | `{{ badge.compact() }}` |

The first row throws `item.badge.compact is not a function` on first render; the fifth is rejected by the template parser.

`navbar-signals-and-aria/index.ts:413` (`memberAccessPattern`) already guards the assignment case with `(?!\s*=(?!=))` and reports those separately via `collectRefWriteWarnings`, with a comment explaining why.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
+    const start = node.getStart(sourceFile);
+    const end = node.getEnd();
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
```
lskramarov: **A local that shadows a badge receiver is rewritten as if it were the badge.**

`inReceiverScope` matches on receiver *text* + source range only, so any inner declaration reusing the name is inside the outer receiver's scope. Verified:

```ts
class D { r(badge: KbqBadge, rows: any[]) { return rows.filter((badge) => badge.compact); } }
// → rows.filter((badge) => badge.compact())
```

The arrow parameter is `any`, so this compiles clean and throws `TypeError: badge.compact is not a function` on the consumer's own data at runtime. `for (const badge of rows) { console.log(badge.compact); }` inside the same method breaks identically.

`navbar-signals-and-aria/index.ts:176` (`receiverOf`, innermost wins) and `button-toggle-signals-and-aria/index.ts:242` (`resolveDeclaration`) both collect *untyped* inner declarations for exactly this reason.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    let result = content;
+
+    for (const { start, end, text } of sorted) {
+        result = result.slice(0, start) + text + result.slice(end);
+    }
+
+    return result;
+}
+
+const isFunctionLike = (node: ts.Node): boolean =>
```
lskramarov: **`isFunctionLike` omits `MethodSignature`, so a `KbqBadge` parameter in a type position widens the receiver scope to the whole file.**

When `findAncestor(node, isFunctionLike)` finds nothing it falls back to `sourceFile` (lines 103 and 124), and every same-named receiver in the file is then in scope. Verified:

```ts
interface R { read(badge: KbqBadge): void; }
class P { compact = false; c(badge: P) { return badge.compact; } }
// → return badge.compact();   ← P, not KbqBadge
```

One receiver is collected, `{ text: 'badge', scope: SourceFile }`, and the unrelated `P` receiver becomes a call — a compile error the consumer did not have before. `type Handler = (badge: KbqBadge) => void;` (a `FunctionTypeNode`) reproduces it identically.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+            commit(filePath, original, content);
+        }
+
+        for (const filePath of htmlPaths) {
+            const original = tree.read(filePath)?.toString();
+
+            if (!original) continue;
+
+            const { content, changed } = await migrateTemplate(original);
```
lskramarov: **The HTML pass never reports `badgeColor` — the one member whose *value* changed.**

This loop only calls `migrateTemplate`, which handles `SIGNAL_MEMBERS` (`compact`/`outline`). `warnReceiverMembers` is called only from the `.ts` loop and walks the TypeScript AST, so it can see neither `.html` files nor inline template literals.

```html
<kbq-badge #b [badgeColor]="'error'">x</kbq-badge><span [class]="b.badgeColor">
```

rendered the class string `kbq-badge_error` before; after the migration `b.badgeColor` is the `InputSignal` function object, so the binding silently breaks — no rewrite, no warning, no compile error.

`app-switcher-signals/index.ts:314` adds `collectRefManualMembers(template, refs)` for precisely this case and threads it into the report.

packages/schematics/src/migrations/badge-signals/data.ts
```diff
@@ -0,0 +1,95 @@
...
+ * File-scoped patterns that can't be auto-fixed reliably — surfaced with file locations (in both `fix` and
+ * dry-run mode). Only evaluated for files that reference the badge, so the patterns stay scoped.
+ */
+export const warnPatterns: WarnPattern[] = [
+    {
+        anchor: '\\bKbqBadge\\b',
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqBadge\\b',
+        message:
+            'A KbqBadge view/content query returns the component instance, whose `compact`/`outline` are now ' +
+            'signals — reading them is a double call, e.g. `this.badge().compact()`. Verify query reads manually.'
```
lskramarov: **This "double call" advice contradicts the fix the same run applies.**

The pattern above also matches the *decorator* form. Verified:

```ts
@ViewChild(KbqBadge) badge: KbqBadge;   // pattern matches → warning fires
```

That is `index.spec.ts:73`'s own fixture. For it the schematic correctly emits `this.badge.compact()` — a **single** call — while logging "reading them is a double call, e.g. `this.badge().compact()`". A user who follows the printed guidance writes `this.badge().compact()` and gets `TypeError: this.badge is not a function`, since `this.badge` is a `KbqBadge` instance, not a signal. `README.md:43` carries the same conflation in one row.

Separately, `[^\n;]*` on line 71 is line-bound, so the prettier-wrapped form gets **no** warning at all (also verified):

```ts
readonly badge = viewChild(
    KbqBadge
);
```

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+            if (changed) {
+                consumers++;
+                commit(filePath, original, content);
+            }
+        }
+
+        // Nothing here uses the badge, so the summary would only be noise.
+        if (consumers === 0) return;
```
lskramarov, line 933: **This early return suppresses the report for exactly the projects the `booleanAttribute` note targets.**

`consumers` is incremented for a `.ts` file only when `referencesBadge` matches, and for an `.html` file only when it was actually *rewritten*. A project whose badge usage is markup-only hits neither:

```html
<!-- app.html -->
<kbq-badge compact>5</kbq-badge>
```
```ts
// app.ts — pulls the badge in through a shared NgModule, names no badge symbol
```

Result: zero log output. Both badges silently flip from default size to compact — the one change in this PR that alters rendering — and the guides' "the rest is reported" is false for that project. `alert-signals` prints its footer unconditionally.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    return { content, changed };
+}
+
+/** Pass B (core) — parse a template, discover badge refs, rewrite their value-safe signal reads. */
+async function migrateTemplate(template: string): Promise<{ content: string; changed: boolean }> {
+    if (!template.includes(BADGE_ELEMENT)) return { content: template, changed: false };
+
+    const parsed = await parseTemplate(template);
+
+    if (!parsed.tree) return { content: template, changed: false };
```
lskramarov: **A template the parser rejects is skipped in total silence.**

`parseTemplate` runs with `tokenizeExpansionForms: true`, so a bare `{` in text — a JSON/code sample in a `<pre>`, an unbalanced brace — yields `{ tree: undefined }`. For

```html
<kbq-badge #b>5</kbq-badge><pre>{ "a": 1 }</pre><span>{{ b.compact }}</span>
```

this returns `changed: false`, `consumers` is not incremented, nothing is logged, and the stale `b.compact` read ships.

`app-switcher-signals/index.ts:351`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria` and `navbar-signals-and-aria` all set `unparseable: true` on this branch and log `UNPARSEABLE_TEMPLATE_MESSAGE`.

packages/components/badge/badge.component.ts
```diff
@@ -122,25 +126,24 @@ export class KbqBadgeCssStyler {
...
 
-    set badgeColor(value: string | KbqBadgeColors) {
-        this._badgeColor = value || KbqBadgeColors.FadeContrast;
-    }
+    /** Color of the badge. */
+    readonly badgeColor = input<string | KbqBadgeColors>(KbqBadgeColors.FadeContrast);
 
-    private _badgeColor: string | KbqBadgeColors = KbqBadgeColors.FadeContrast;
+    /** @docs-private */
+    protected readonly badgeColorClass = computed(
```
lskramarov: **Putting the fallback in the derived `computed` rather than in the input's `transform` keeps a smaller version of the bug this PR fixes.**

The input stores the falsy value verbatim, so with `[badgeColor]="color()"` and `color()` returning `''`:

- the host renders `kbq-badge_fade-contrast`
- `badgeRef().badgeColor()` returns `''`, behind a published type of `InputSignal<string>`

That is still "a read that does not match what you see" — the defect the PR description opens with.

The sibling review put it on the input (`packages/components/alert/alert.component.ts:126`):

```ts
readonly alertColor = input(KbqAlertColors.Info, {
    transform: (value: KbqEnumValues<KbqAlertColors> | null | undefined): KbqEnumValues<KbqAlertColors> =>
        value || KbqAlertColors.Info
});
```

so `alertColor()` can never report a falsy value and the class expression is a plain interpolation. The new spec covers only the class half (`badge.component.spec.ts:67`); nothing asserts what `badgeColor()` reports for `''`.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    ts.SyntaxKind.ProtectedKeyword,
+    ts.SyntaxKind.ReadonlyKeyword
+]);
+
+/**
+ * Collects the receivers annotated with `typeName`, by explicit annotation only (no cross-package type
+ * resolution): method/function params, class fields (incl. `@ViewChild(KbqBadge) x: KbqBadge` and constructor
+ * parameter-properties) and typed locals.
+ */
+function collectReceivers(sourceFile: ts.SourceFile, typeName: string): Receiver[] {
```
lskramarov: **Receiver discovery is annotation-only and there is no "could not resolve this receiver" note, so a long tail of ordinary shapes is silently skipped.**

All verified to produce zero receivers (or zero edits) and zero log lines:

- `this.badge!.compact` — receiver text is `this.badge!` (a `NonNullExpression`)
- `(b as KbqBadge).compact`
- a field typed `KbqBadge | undefined` — a `UnionTypeNode` is not a `TypeReferenceNode`
- `const { compact } = badge;`
- `badge['compact']`
- `badges: KbqBadge[]` / `QueryList<KbqBadge>` with `badges.forEach(b => b.compact)`
- `import { KbqBadge as Badge }`

The dangerous ones are those that still compile: `if (this.badge!.compact)` now tests an `InputSignalWithTransform` function object, which is always truthy, so the branch inverts with no error and nothing printed.

`button-toggle-signals-and-aria/index.ts:331` emits `UNRESOLVED_SIGNAL_READ_MESSAGE` for exactly this blind spot.

packages/schematics/src/migrations/badge-signals/index.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+            'class Demo {\n' +
+            '    read(badge: KbqBadge) {\n' +
+            '        return badge.badgeColor;\n' +
+            '    }\n' +
+            '}\n';
+
+        appTree.overwrite(ts, source);
+
+        expect((await run()).readText(ts)).toContain('return badge.badgeColor;');
+        expect(messages.join('\n')).toContain('kbq-badge_<color>');
```
lskramarov: **This assertion is satisfied by a different code path, so the `badgeColor` warning is untested.**

`kbq-badge_<color>` also appears in `data.ts:93` (`SUMMARY[1]`), which is printed for every project with `consumers > 0` — and this fixture has one. The companion assertion `toContain('return badge.badgeColor;')` cannot fail either, since `badgeColor` is not in `SIGNAL_MEMBERS`. Delete the entire `valueChanged` branch of `collectReceiverWarnings` and this test stays green.

Same shape at line 221: `isIconButton` is interpolated into `warnPatterns[1].message` via `STYLER_PRIVATE_MEMBERS.join(', ')`, which `logWarnings` prints for any file merely *naming* `KbqBadgeCssStyler` — so the `stylerAccess` branch is untested too.

Assert on a string unique to each path instead, e.g. `'Refactor to avoid reading them'` and `'there is nothing left to drive from outside'`.

packages/schematics/src/migrations/badge-signals/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+    visitAll(collector, (parsed.tree as { rootNodes: unknown[] }).rootNodes);
+
+    if (collector.refs.size === 0) return { content: template, changed: false };
+
+    return rewriteRefReads(template, [...collector.refs]);
+}
+
+/** Interior `[start, end]` ranges of inline `@Component({ template: '…' })` string literals. */
+function collectInlineTemplateRanges(sourceFile: ts.SourceFile): Array<{ start: number; end: number }> {
```
lskramarov: **Byte-identical copy of a helper already exported from the module this file imports on line 7.**

`packages/schematics/src/utils/typescript.ts:74` exports `collectInlineTemplateRanges` with the same body, brace for brace, down to the `+1` / `-1` comment. Line 7 here is `import { forEachClass, parseTemplate } from '../../utils/typescript'` — and `forEachClass` is imported *only* to feed this copy.

Eight call sites already use the shared one: `app-switcher-signals`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `list-tree-multiple-input`, `navbar-signals-and-aria`, `utils/icon-migration`.

The shared version also documents a rule this copy drops ("A template literal with a `${…}` substitution is skipped"), so a future fix there — e.g. handling `ts.isTemplateExpression` — will silently not reach this migration.

Delete lines 351-385 and add `collectInlineTemplateRanges` to the line-7 import, dropping `forEachClass`.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **This sentence contradicts itself: three of the four subsections below are components it just listed in the first wave.**

Section 18's subsections are, in order:

| line | subsection | named in the first-wave list on this line? |
| --- | --- | --- |
| 1052 | `#### Badge` | no — the one this PR adds |
| 1075 | `#### Search expandable` | **yes** |
| 1097 | `#### Split button` | **yes** |
| 1119 | `#### Title` | **yes** |

So "the second is the one each subsection below belongs to" tells a reader that search-expandable, split-button and title were reviewed in both waves. Attributing the wave per subsection (or just naming badge in the list) avoids it.

`docs/guides/migration.ru.md:1042` carries the identical claim and needs the same fix.

packages/schematics/src/migrations/badge-signals/index.ts
artembelik: Вместо якорей в регулярке переписываю только внутри диапазонов выражений, собранных из разобранного шаблона, плюс lookbehind против `item.badge` и guard присваивания. Ссылка вдобавок ограничена своим embedded view, поэтому шестая строка таблицы тоже закрыта: все шесть случаев зафиксированы тестами.

packages/schematics/src/migrations/badge-signals/index.ts
artembelik: Печатаю ноту с номерами строк, а заодно научил проход видеть `!`, скобки и алиас импорта, так что `this.badge!.compact` теперь переписывается. `badge['compact']` и деструктуризацию репортю, разрешать их не стал.

docs/guides/migration.en.md
artembelik: Убрал деление на волны целиком: список первой волны устаревал с каждым PR. Во вступлении осталось только то, что подраздела нет у компонента, чьё ревью ничего видимого снаружи не изменило.

packages/schematics/src/migrations/badge-signals/index.ts
artembelik: Считаю потребителем любой файл, где встречается `<kbq-badge`, а не только переписанный, поэтому проект с разметкой без символов бейджа сводку получает. Безусловный футер печатал бы её и в проектах, где бейджа нет вовсе.

lskramarov (approved):

PR #1979, opened 2026-09-02
fix(progress-spinner)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `progress-spinner`, in the same shape as the 20.3.0 component reviews.

### `size` was the last accessor input — and it did two things

```ts
set size(value: ProgressSpinnerSize | string) {
    this._size = value;
    this.svgCircleRadius = value === 'big' ? '47%' : '42.5%';
}
```

That side effect is exactly why the automated signal migration skipped it. The radius is a `computed` now, `size` is a plain `input()`, and it is typed `ProgressSpinnerSize` instead of an arbitrary string — resolving the `//@TODO use Exclude<KbqDefaultSizes, 'normal'>` that predates this review.

### `value` gained `numberAttribute`

`value="40"` used to pass the string `"40"`; `Math.min(100, '40')` coerced it by accident. It is a number now.

### Closed internals

`percentage`, `dashOffsetPercent` and `svgCircleRadius` are `protected`. The first two are derived from the `value` the consumer already binds; the last is SVG geometry, not a contract.

### `_IdGenerator`

The generated `id` comes from the CDK `_IdGenerator` instead of a module-level `let id = 0`, so it no longer collides across lazily loaded bundles. The `kbq-progress-spinner-<n>` shape is unchanged.

### One line outside the component

`KbqLoaderOverlay.spinnerSize` reports `ProgressSpinnerSize` rather than `string`, to keep feeding the narrowed input. It is a narrowing of a getter's return type, so nothing that reads it breaks.

## Migration

`progress-spinner-signals` runs from `ng update @koobiq/components@20`. It rewrites `size` reads to calls — on receivers typed `KbqProgressSpinner` and through template reference variables on `<kbq-progress-spinner>`, in external and inline templates — and reports the rest.

## Testing

- `progress-spinner.component.spec.ts` rewritten: 8 → 10 tests. The old ones read `percentage` off the instance; they assert the rendered `stroke-dashoffset`, circle radius and host classes now, which is what the value actually drives.
- `progress-spinner-signals/index.spec.ts`: 13 tests — auto-fix, idempotence, receiver scoping, template refs, warnings, and the `--fix=false` path.
- Full `packages/components` (4994 tests) and `packages/schematics` (447 tests) suites pass.
- `check-api` is in sync.

No e2e screenshots were regenerated: the e2e cases pass `size="compact"` / `size="big"` as before, and the rendered geometry is unchanged.

BREAKING CHANGE


packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return (
+        /\bKbqProgressSpinner\w*\b/.test(content) ||
+        content.includes(SPINNER_PACKAGE) ||
+        content.includes(`<${SPINNER_ELEMENT}`)
+    );
+}
+
+export default function progressSpinnerSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **`fix` is `undefined` under `ng update`, so the migration never writes anything.**

`migrations.json` declares no `schema` for this entry (correctly — that matches every sibling), so the
`"default": true` in `schema.json` never reaches the rule when the CLI runs the migration. `options.fix` is
`undefined`, `commit()` takes the else branch for every file, and the consumer gets
`[progress-spinner-signals] would update <file> (run with --fix to apply)` with zero writes — followed by a
build failure on the very `size` reads that were supposed to be rewritten.

Eleven siblings guard this. `list-tree-multiple-input/index.ts:219-220` even carries the explanation:

> `ng update` invokes migrations with no options at all, and migrations.json declares no schema,
> so the schema default never reaches us — applying the fix is the intended behaviour there.

`index.spec.ts:37` always passes `fix` explicitly (`run(fix: boolean = true)`), so no test can catch it, and
`Schema.fix` being typed non-optional hides it from the compiler.

```suggestion
        const { project } = options;
        // `ng update` invokes migrations with no options at all, and migrations.json declares no schema,
        // so the schema default never reaches us — applying the fix is the intended behaviour there.
        const fix = options.fix ?? true;
```

packages/components/progress-spinner/progress-spinner.component.ts
```diff
@@ -37,40 +51,36 @@ const MAX_DASH_ARRAY = 295;
...
-    //  Accessor inputs cannot be migrated as they are too complex.
-    @Input()
-    get size(): ProgressSpinnerSize | string {
-        return this._size;
-    }
-
-    set size(value: ProgressSpinnerSize | string) {
-        this._size = value;
+    /** Progress of the operation, in percent. Clamped to `[0, 100]` and only rendered in `determinate` mode. */
+    readonly value = input(0, { transform: numberAttribute });
```
lskramarov: **`numberAttribute` with no fallback turns a nullish binding into a fully-drawn ring.**

`numberAttribute(v, fallbackValue = NaN)` returns `NaN` for `null`, `undefined`, `''`, `'abc'`, `'40px'`,
`true`, `[]` and `{}`. `NaN` survives the clamp — `Math.max(0, Math.min(100, NaN))` is `NaN` — so
`dashOffsetPercent()` produces the literal string `"NaN%"`.

CSSOM rejects an unparseable declaration value, so `style.setProperty('stroke-dashoffset', 'NaN%')` is a
**no-op**:

- **first render** — nothing is set. `.kbq-progress-spinner__circle` declares `stroke-dasharray: 295%` and no
  `stroke-dashoffset` outside the indeterminate branch, so the initial `0` applies and the ring paints
  **100% complete**.
- **later transition from a valid value** — the previous inline offset stays, so the ring **freezes** at the
  old percentage.

So `<kbq-progress-spinner [value]="progress$ | async" />` renders "done" while the data is still loading. The
base version clamped `null` to `0` and painted an empty ring.

The type-level guard moved the other way at the same time: the guard file now records
`InputSignalWithTransform<number, unknown>`, so under `strictTemplates` `[value]="progress$ | async"`,
`[value]="'40%'"` and `[value]="{}"` all type-check where they used to be errors. The PR narrows `size` for
type safety and drops the only static check on `value` in the same commit.

Minor, same subject: the migration note says "A binding that passes `null` or `undefined` used to clamp to
`0`". Only `null` did — `Math.min(100, undefined)` was already `NaN` on the base version.

```suggestion
    readonly value = input(0, { transform: (value: unknown) => numberAttribute(value, 0) });
```

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
```
lskramarov: **The template-ref rewrite has no assignment guard, so it emits a template that will not parse.**

The lookahead only skips an immediate `(`, so an assignment passes straight through:

```html
<kbq-progress-spinner #spinner />
<button (click)="spinner.size = 'big'">grow</button>
```

becomes `(click)="spinner.size() = 'big'"` — an Angular expression parse error. The migration turns a
compiling app into a build failure.

This also contradicts the TS pass in this same file: the comment at lines 149-150 says a write is
*deliberately* left untouched because there is no writable member. The invariant holds in TypeScript and is
broken in every external and inline template.

`navbar-signals-and-aria/index.ts:405-420` already has exactly this guard, with the reasoning spelled out
("rewriting `ref.expanded = value` blindly into `ref.expanded() = value` would replace a valid assignment
with invalid syntax"), and reports such writes separately via `collectRefWriteWarnings`.

```suggestion
        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()(?!\\s*=(?!=))`, 'g');
```

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+        const next = content.replace(pattern, '$1.$2()');
```
lskramarov: **The rewrite is a raw regex over the whole template text, so it corrupts unrelated identifiers.**

The AST is used only to *discover* ref names; the replacement then ignores it entirely and matches every
textual `<ref>.size` in the file. Verified outputs for a template that declares
`<kbq-progress-spinner #spinner />`:

| input | output |
| --- | --- |
| `@for (spinner of rows; track spinner.id) { {{ spinner.size }} }` | `{{ spinner.size() }}` on a plain row object |
| `[cfg]="state.spinner.size"` | `state.spinner.size()` — `\b` matches after a `.` |
| `<!-- spinner.size is derived -->` | comment rewritten |
| `<p>Read spinner.size to get the size.</p>` | visible body text rewritten |
| `spinner.size` | attribute string rewritten |

The `@for` case is the dangerous one: `spinner` is a very common loop-variable name, and the result is
`TypeError: spinner.size is not a function` on the first change-detection pass, in a file the migration
reported as successfully fixed.

Two smaller issues in the same expression:

- `members` (`SIGNAL_MEMBERS.join('|')`) is interpolated **unescaped** while `ref` goes through
  `escapeRegExp` — safe only because the array is `['size']` today.
- `SpinnerRefCollector` ignores the ref's `exportAs` value, so `#spinner="cdkOverlayOrigin"` on a spinner
  element is collected as a spinner ref even though `spinner` is then the directive instance.

lskramarov (commented):

15 findings, ranked most severe first. Thirteen are correctness, one is test coverage, one is the
migration guide.

The two that block: the schematic never writes anything under `ng update`
(`index.ts:394`), and `numberAttribute` without a fallback makes a nullish `[value]` paint a **full** ring
instead of an empty one (`progress-spinner.component.ts:64`).

Most of the rest are guards that exist in the newer sibling migrations (`navbar-signals-and-aria`,
`button-toggle-signals-and-aria`) but not in this clone of the older `alert-signals`: the assignment
lookahead in the ref regex, `variableScope` for block-scoped locals, `initializerTypeOf` for
`inject()`/`viewChild()` receivers, and the template-side manual-member report.

*Posted by Claude Code.*

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    // Already migrated: `x.size()` (call) or `x.size.set(...)` — leave alone (idempotent).
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.size = RHS`. Every KbqProgressSpinner signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
```
lskramarov: **Only `=` is recognised as a write, so every other assignment form is rewritten into invalid syntax.**

The early `return` lives *inside* the `EqualsToken` branch, so any other parent falls through to the "read"
branch at line 167 and gets `()` appended. Verified outputs on a receiver typed `KbqProgressSpinner`:

| in | out |
| --- | --- |
| `spinner.size += 'x';` | `spinner.size() += 'x';` |
| `spinner.size++;` | `spinner.size()++;` |
| `spinner.size ||= 'big';` | `spinner.size() ||= 'big';` |
| `[spinner.size] = ['big'];` | `[spinner.size()] = ['big'];` |
| `({ x: spinner.size } = o);` | `({ x: spinner.size() } = o);` |

All five are `TS2364: The left-hand side of an assignment expression must be a variable or a property
access` — the consumer gets a syntax error on already-overwritten source, rather than the clean compile
error the comment above promises.

Conversely `const { size } = spinner`, `spinner['size']` and `delete spinner.size` are silently skipped with
no edit and no warning. Guarding the read branch on "not an assignment target of any kind" covers both
directions better than enumerating one token.

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+ * parameter-properties) and typed locals.
+ */
+function collectReceivers(sourceFile: ts.SourceFile, typeName: string): Receiver[] {
+    const receivers: Receiver[] = [];
+    const add = (text: string, scope: ts.Node) =>
+        receivers.push({ text, start: scope.getStart(sourceFile), end: scope.getEnd() });
+
+    const visit = (node: ts.Node): void => {
+        if (ts.isParameter(node) && ts.isIdentifier(node.name) && isTypeReference(node.type, typeName)) {
+            add(node.name.text, findAncestor(node, isFunctionLike) ?? sourceFile);
```
lskramarov: **A parameter in a type-only position escalates its receiver scope to the whole file.**

`isFunctionLike` (lines 55-62) lists only declarations — it omits `FunctionTypeNode`, `MethodSignature`,
`ConstructorTypeNode` and the call signatures. A parameter declared in a type therefore has no function-like
ancestor, `?? sourceFile` kicks in, and the receiver claims the entire source file.

```ts
export type SpinnerReady = (spinner: KbqProgressSpinner) => void;

export class Grid {
    layout(spinner: { size: number }) {
        return spinner.size * 2;   // → spinner.size() * 2
    }
}
```

Verified: `collectReceivers` returns a single receiver `{ text: 'spinner', start: 0, end: <EOF> }`, and the
unrelated `spinner.size` is rewritten → `TypeError: spinner.size is not a function`. Identical result for
`interface Host { onSpinner(spinner: KbqProgressSpinner): void }`, and for a module-level
`let spinner: KbqProgressSpinner;` at line 121.

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
+    const start = node.getStart(sourceFile);
+    const end = node.getEnd();
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
```
lskramarov: **Receiver identity is source text plus a span, so a shadowing binding of the same name is rewritten.**

Nothing re-binds the identifier, so any inner declaration reusing the name inside the receiver's span is
treated as the spinner. Verified:

```ts
read(spinner: KbqProgressSpinner) {
    boxes.forEach((spinner: Box) => console.log(spinner.size)); // → spinner.size() on a Box
    return spinner.size;
}
```

Same for a nested class re-binding `this.spinner`, and for *sibling* block-scoped declarations, because line
121 gives a typed `const` the whole enclosing function:

```ts
function render(kind: string) {
    if (kind === 'a') { const spinner: KbqProgressSpinner = this.a; use(spinner.size); }
    else { const spinner = this.otherWidget; use(spinner.size); } // rewritten too
}
```

`navbar-signals-and-aria/index.ts:106-126` added `isBlockScopeContainer` + `variableScope` for exactly this,
with the comment "Giving a block-scoped local the whole function would let an access elsewhere in that
function match it even where the name refers to a different, shadowing declaration." The clone dropped it.

Mirror-image gap: `this . spinner.size` and `this /* x */.spinner.size` are silently *missed*, because
`getText()` returns the raw spelling.

`index.spec.ts` has a "leaves reads on a receiver of an unrelated type alone" case, but its unrelated
receiver has a *different name* (`other`), so shadowing is untested.

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+        if (predicate(current)) return current;
+        current = current.parent;
+    }
+
+    return undefined;
+}
+
+/** Whether a type annotation refers to `typeName`. */
+function isTypeReference(type: ts.TypeNode | undefined, typeName: string): boolean {
+    return !!type && ts.isTypeReferenceNode(type) && ts.isIdentifier(type.typeName) && type.typeName.text === typeName;
```
lskramarov: **A union- or array-typed receiver is a silent no-op — rewrite *and* warnings.**

`ts.isTypeReferenceNode` is false for a `UnionTypeNode`, so this returns `false` and `collectReceivers`
yields `[]`. That short-circuits both `migrateTsExpressions` (line 219) and `warnReceiverMembers`
(line 232).

```ts
class Demo {
    @ViewChild(KbqProgressSpinner) spinner: KbqProgressSpinner | undefined;
    read() { return this.spinner!.size + this.spinner!.percentage; }
}
```

Verified: the file is left completely untouched. `size` stays an un-called `InputSignal` read **and** the
"`percentage` is now protected" warning never fires — the user gets a clean migration report and a broken
build.

`KbqProgressSpinner | undefined` is the idiomatic `@ViewChild` annotation under `strictNullChecks`, so this
is the common shape, not an edge case. Same for `KbqProgressSpinner[]` and for an aliased import
(`import { KbqProgressSpinner as Spinner }`).

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    ts.SyntaxKind.ProtectedKeyword,
+    ts.SyntaxKind.ReadonlyKeyword
+]);
+
+/**
+ * Collects the receivers annotated with `typeName`, by explicit annotation only (no cross-package type
+ * resolution): method/function params, class fields (incl. `@ViewChild(KbqProgressSpinner) x: KbqProgressSpinner` and constructor
+ * parameter-properties) and typed locals.
+ */
+function collectReceivers(sourceFile: ts.SourceFile, typeName: string): Receiver[] {
```
lskramarov: **`inject()` / `viewChild()` receivers are invisible — the sibling support for them was dropped.**

Receivers are collected from explicit type annotations only, so the initializer-typed shapes a modern
Angular consumer actually writes are never recognised:

```ts
private readonly spinner = inject(KbqProgressSpinner);
readonly spinner = viewChild(KbqProgressSpinner);
readonly spinner = viewChild.required(KbqProgressSpinner);
```

For the `inject` case the file is still *counted* as a consumer (`referencesSpinner` matches the import, so
`consumers++` fires and the summary prints), but `this.spinner.size` and `this.spinner.svgCircleRadius` get
no rewrite and no warning of any kind — `warnPatterns` (data.ts:45) never matches `inject` either.

`button-toggle-signals-and-aria/index.ts:108-175` added `collectTypeAliases` (for `import { X as Y }`) and
`initializerTypeOf` (for `viewChild`/`viewChild.required`/`contentChild`/`inject`), with the comment "which
is how a modern Angular consumer reaches a group". None of it survived the clone from the older
`alert-signals`.

packages/schematics/src/migrations/progress-spinner-signals/data.ts
```diff
@@ -0,0 +1,60 @@
...
+ * `input()`.
+ *
+ * - `spinner.size` → `spinner.size()` (value unchanged — auto-fixed)
+ * - `spinner.svgCircleRadius` / `percentage` / `dashOffsetPercent` → `protected` (warn)
+ *
+ * `id`, `value` and `mode` were already signals in 20.2.0 and are not touched here.
+ */
+
+/** Members of `KbqProgressSpinner` whose value is unchanged; a read must become a call. Auto-fixed. */
+export const SIGNAL_MEMBERS: readonly string[] = ['size'];
```
lskramarov: **`id`, `value` and `mode` became signals in 20.0.0, not 20.2.0 — and no schematic has ever covered them.**

The doc comment four lines up (repeated in `README.md:19`, `migration.en.md` and `migration.ru.md`) says
they "were already signals in 20.2.0 and are not touched here". Checking:

- `git show fec9db9a9:packages/components/progress-spinner/progress-spinner.component.ts` still has
  `@Input() id`, `@Input() value`, `@Input() mode`;
- they became `input()` in `17bc91dd3 feat: angular 20 (#DS-5079)`, i.e. **v20.0.0**;
- `git grep -n "ProgressSpinner" packages/schematics/src` matches nothing outside this new folder, so no
  migration ever rewrote a consumer read of them.

This schematic is registered for `20.3.0-0`, so it *does* run for a v19 → v20.3 upgrade — and it is the only
place those three would ever be handled. Left as-is, a consumer keeps `spinner.value`, `spinner.mode`,
`spinner.id` unrewritten and unreported: `spinner.value > 50` is now always `false`, `{{ spinner.value }}`
prints the function source, and a surviving `spinner.value = 40` overwrites the `InputSignal` property so
the next change detection throws `this.value is not a function` from `percentage`.

Note the ordering constraint: adding them to `SIGNAL_MEMBERS` is only safe once the template pass is scoped
to binding expressions (see the comment on `index.ts:290`) — `value` is far too common an identifier to feed
to a raw-text regex.

packages/schematics/src/migrations/progress-spinner-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+            commit(filePath, original, content);
+        }
+
+        for (const filePath of htmlPaths) {
+            const original = tree.read(filePath)?.toString();
+
+            if (!original) continue;
+
+            const { content, changed } = await migrateTemplate(original);
```
lskramarov: **Protected-member reads in templates are never reported, though three docs promise they are.**

`PROTECTED_MEMBERS` is reachable only through `collectProtectedAccess` → `warnReceiverMembers`, which runs
exclusively in the `.ts` loop and keys off TypeScript type annotations. This HTML loop calls only
`migrateTemplate`, which rewrites `size` and returns nothing else; `migrateInlineTemplates` likewise
produces no findings.

So:

```html
<kbq-progress-spinner #s [value]="v" />
<span>{{ s.percentage }}</span>
```

migrates with a clean report, then fails the next AOT build with NG9 *"Property 'percentage' is protected
and only accessible within class 'KbqProgressSpinner'"* — and no file pointer. `collection.json:148`,
`README.md:37` and the migration guide all state these members are reported.

Both newer siblings have a template-side reporting path for exactly this (`collectRefManualMembers` in
`navbar-signals-and-aria`, `Findings.manual` in `button-toggle-signals-and-aria`).

Two smaller gaps in the same loop worth folding in: an unparseable template is swallowed at `index.ts:307`
with no diagnostic at all (the siblings thread an `unparseable` flag and log
`UNPARSEABLE_TEMPLATE_MESSAGE`), and `consumers` is incremented on "mentions the spinner" for `.ts` but on
"was actually rewritten" for `.html`, so a project whose only usage is a plain `[value]` binding in an
external template can end with `consumers === 0` and lose the whole `SUMMARY`.

packages/schematics/src/migrations/progress-spinner-signals/data.ts
```diff
@@ -0,0 +1,60 @@
...
+    /** Owner of the member. The pattern is only evaluated for files that also name it. */
+    anchor: string;
+    pattern: string;
+    message: string;
+}
+
+export const warnPatterns: WarnPattern[] = [
+    {
+        anchor: '\\bKbqProgressSpinner\\b',
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqProgressSpinner\\b',
```
lskramarov: **The warn pattern matches the decorator queries too, and then tells the user to write code that throws.**

The alternation includes `ViewChild`/`ContentChild`, but a decorator query returns the **instance**, not a
signal. The migration's own spec case (`index.spec.ts:73-89`) matches:

```ts
@ViewChild(KbqProgressSpinner) spinner: KbqProgressSpinner;
read() { return this.spinner.size; }
```

The schematic correctly rewrites it to `this.spinner.size()` **and**, in the same run, logs "reading it is a
double call, e.g. `this.spinner().size()`". A consumer who follows the warning edits the correct output into
`this.spinner().size()` → `TypeError: this.spinner is not a function`.

The spec asserts the rewrite but never asserts the warning is absent, so the contradiction is invisible.

```suggestion
        pattern: '(?:viewChild|contentChild)[^\\n;]*\\bKbqProgressSpinner\\b',
```

packages/components/progress-spinner/progress-spinner.component.ts
```diff
@@ -1,16 +1,30 @@
...
-//@TODO use Exclude<KbqDefaultSizes, 'normal'> from '@koobiq/components/core' instead
-export type ProgressSpinnerSize = 'compact' | 'big';
-
-let id = 0;
+/** Sizes supported by the progress spinner. */
+export type ProgressSpinnerSize = Exclude<KbqDefaultSizes, 'normal'>;
 
 const MIN_PERCENT = 0;
 const MAX_PERCENT = 100;
+const MAX_DASH_ARRAY = 295;
```
lskramarov, line 22: **The dash array is calibrated for the *big* radius only, so the default compact size over-draws.**

The review made the radius vary with `size` (`47%` / `42.5%`) but left `MAX_DASH_ARRAY` size-independent.
With `viewBox="0 0 100 100"`, SVG percentages resolve against a normalized diagonal of 100, so:

| size | `r` | path length | dash array |
| --- | --- | --- | --- |
| big | `47%` | 2π·47 = **295.3** | 295 ✔ |
| compact (default) | `42.5%` | 2π·42.5 = **267.0** | 295 ✘ |

Drawn fraction on compact is `295·v / 267 = 1.105·v`:

- `[value]="44"` — what `e2e.ts` renders — paints 48.6% on compact vs 44.0% on big. Same input, two
  different pictures.
- `[value]="50"` paints 55%.
- everything from **90.5 upward paints a complete ring**, so a compact determinate spinner has no last 10%
  of travel.

Pre-existing, but this PR rewrote all three functions involved (`svgCircleRadius`, `percentage`,
`dashOffsetPercent`) and replaced the spec's size-independent `percentage` assertions with raw offset
strings (`dashOffsetPairs`), which turns the miscalibration into an asserted expectation rather than a
detectable defect.

`pathLength="100"` on the `<circle>` makes the arithmetic size-independent by construction. Note
`progress-spinner.scss:22` hardcodes the same `295` a second time, and the indeterminate arc
(`stroke-dashoffset: 80%`, under an untouched `// TODO: rework this place`) is off by the same ratio: 80.5%
of the ring on compact vs 72.8% on big.

packages/components/progress-spinner/progress-spinner.component.spec.ts
```diff
@@ -1,132 +1,133 @@
...
     });
 
-    it('should auto generate id', () => {
-        const fixture = TestBed.createComponent(TestApp);
-        const progressSpinnerDebugElement = fixture.debugElement.query(By.css('.default'));
-
-        expect(progressSpinnerDebugElement.nativeElement.getAttribute('id')).toBeDefined();
+    it('should auto generate a unique id', () => {
+        expect(defaultHost.getAttribute('id')).toMatch(/^kbq-progress-spinner-/);
+        expect(defaultHost.getAttribute('id')).not.toBe(host.getAttribute('id'));
```
lskramarov: **"should auto generate a unique id" cannot fail — it compares against a hardcoded id.**

`host` is the `.first` spinner, whose id is bound to `readonly id = signal('test-spinner')` (line 132). So
this line compares a generated id against the literal `'test-spinner'`, which the `toMatch(/^kbq-progress-
spinner-/)` on the line above already implies.

Replace `inject(_IdGenerator).getId('kbq-progress-spinner-')` with a constant string and both assertions
still pass — the collision the PR description says was fixed ships untested. Uniqueness needs a **second
unbound** `<kbq-progress-spinner />` in `TestApp` so two generated ids can be compared.

Related gaps in this file:

- The `numberAttribute` transform — the PR's headline breaking change — has zero coverage in either
  direction. `TestApp.value` is `signal(0)` typed `number` and no test binds `value="40"`, `null` or `''`.
  Dropping the transform entirely would leave all 10 tests green.
- If you do add such a test, assert the *number*, not the style: cssstyle 2.3.0 keeps `stroke-dashoffset` in
  `allExtraProperties` with no value validation, so under jsdom `style.strokeDashoffset` reads back
  `'NaN%'` where a real browser rejects it and keeps the previous declaration — i.e. the assertion would
  encode the opposite of production behaviour.
- The two negative tests in `progress-spinner-signals/index.spec.ts` (lines 91 and 161) both bail at an
  early guard before reaching the check they are named for: deleting
  `element.name === SPINNER_ELEMENT` from `SpinnerRefCollector.visitElement` keeps that suite green.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **The new "two waves" sentence contradicts the subsections it introduces.**

It assigns *search-expandable*, *split-button* and *title* to the first wave, then says "the second is the
one each subsection below belongs to" — but the `####` headings that follow are **Progress spinner**,
**Search expandable**, **Split button** and **Title**. Three of the four are in the list they were just
excluded from.

The pre-PR sentence ("Ten components went through a full review in 20.3.0: …") was coherent with the same
subsections. As written, a reader cannot tell which sections are new in this release.

`docs/guides/migration.ru.md:1042` carries the identical defect and needs the same edit.

```suggestion
Components went through a full review in 20.3.0. The first wave covered notification-center, popover, select, toast, tooltip, tree and tree-select; search-expandable, split-button, title and progress-spinner followed, and each has a subsection below. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```

packages/schematics/src/migrations/progress-spinner-signals/index.ts
artembelik: Union и прочие неразрешимые формы не резолвлю, а печатаю списком строк с пометкой посмотреть руками; чтения через `!` и скобки при этом переписываются.

packages/schematics/src/migrations/progress-spinner-signals/index.ts
artembelik: `inject()` разбираю полностью, а `viewChild()` оставляю предупреждению: там правильная форма `spinner().size()` - два вызова, и дописать один было бы хуже, чем не трогать.

packages/schematics/src/migrations/progress-spinner-signals/data.ts
artembelik: Взял более узкий паттерн, чем в suggestion: он ещё и не привязан к строке, поэтому ловит перенесённый prettier вызов.

docs/guides/migration.en.md
artembelik: Взял формулировку, которая уже лежит в main без деления на волны: твой вариант дал бы конфликт и откат к ней на ближайшем мерже.

packages/components/progress-spinner/progress-spinner.component.ts
artembelik: Фолбэк был в `6936c42d8`, а в мерж-коммите сузил параметр трансформа до `number | string | null | undefined` - это возвращает статическую проверку, про потерю которой ты писал.

packages/components/progress-spinner/progress-spinner.component.ts
artembelik: Дефект доревьюшный: радиус зависел от размера, а `MAX_DASH_ARRAY` и `stroke-dasharray` нет, ещё с init-коммита `bebf3bbef` - этот PR ни одно из четырёх чисел не трогал. Оставляю как есть и выношу отдельно: `pathLength="100"` меняет отрисовку compact в обоих режимах и требует перегенерации скриншотов, чему в PR про сигналы не место.

lskramarov (approved):

packages/components/loader-overlay/loader-overlay.component.ts
```diff
@@ -116,7 +116,7 @@ export class KbqLoaderOverlay implements OnInit, OnDestroy {
         return !(!!this.text || this.isExternalText || !!this.caption || this.isExternalCaption);
     }
 
-    get spinnerSize(): string {
+    get spinnerSize(): ProgressSpinnerSize {
```
NikGurev, line 119:
```suggestion
    get spinnerSize(): Exclude<KbqDefaultSizes, 'normal'> {
```

переиспользуй размеры из core вместо соседнего компонента, чтобы не увеличивать количество зависимостей между ними

NikGurev (approved):

packages/components/loader-overlay/loader-overlay.component.ts
artembelik: поправим отдельно, не в рамках миграции на сигналы

PR #1981, opened 2026-09-02
fix(markdown)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `markdown`, in the same shape as the 20.3.0 component reviews.

### `markdownText` rendered from its setter

```ts
set markdownText(value: string | null) {
    if (value && this.markdownText !== value) {
        this.resultHtml.set(this.getResultHTML(value));
    }
    this._markdownText = value;
}
```

That side effect is why the automated signal migration skipped it. The rendered HTML is a `computed` now and the input is a plain `input()`.

### Two behavior fixes, both from the `if (value && …)` guard

- **Clearing the text now clears the output.** The setter only re-rendered for a truthy value, so setting `markdownText` back to `null` or `''` left the previous HTML on screen indefinitely.
- **The projected content is a live fallback.** A `<kbq-markdown>` that both projects content and binds `[markdownText]` returns to the projected content once the input goes empty. Before, the projected content was captured once in `afterNextRender`, and only if the input happened to be empty at that moment.

Both fall out of modelling the output as `markdownText() || projectedText()` instead of an imperative write.

### `resultHtml`

Now a read-only `computed` rather than a `WritableSignal`. It was already `protected`, so this only reaches a subclass that pushed HTML into it.

## Migration

`markdown-signals` runs from `ng update @koobiq/components@20`. It rewrites `markdownText` reads to calls — on receivers typed `KbqMarkdown` and through template reference variables on `<kbq-markdown>`, in external and inline templates — and reports the rest.

## Testing

- `markdown.component.spec.ts`: 6 → 8 tests. The two new ones pin the behavior fixes — the output clears, and the projected content takes over. The existing snapshot tests, which cover the whole markdown → HTML surface, are unchanged and still pass, which is the evidence that rendering itself did not move.
- `markdown-signals/index.spec.ts`: 13 tests — auto-fix, idempotence, receiver scoping, template refs, the subclass warning, and the `--fix=false` path.
- Full `packages/components` (4995 tests) and `packages/schematics` (447 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return (
+        /\bKbqMarkdown\w*\b/.test(content) ||
+        content.includes(MARKDOWN_PACKAGE) ||
+        content.includes(`<${MARKDOWN_ELEMENT}`)
+    );
+}
+
+export default function markdownSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **`fix` is `undefined` under `ng update`, so the migration writes nothing.**

`migrations.json` declares no `"schema"` for `markdown-signals`, and the CLI invokes update migrations with no options — so `schema.json`'s `"default": true` never applies. `fix === undefined` makes `commit()` take the else branch for every file, logging

```
[markdown-signals] would update src/app/foo.ts (run with --fix to apply)
```

a flag `ng update` does not accept, and never calling `tree.overwrite`. Every `markdown.markdownText` read stays un-migrated, while the README, `migration.en.md` ("the `markdownText` reads are rewritten") and `collection.json` all promise it runs automatically.

`navbar-signals-and-aria/index.ts:594-596` documents this exact trap, and 11 of the 17 sibling migrations already use `options.fix ?? true`. `index.spec.ts` cannot catch it because `run(fix = true)` always passes the flag explicitly, and `schema.ts:5` declares `fix: boolean` non-optional, hiding the `undefined` from the type checker — worth making it `fix?: boolean` too, as navbar does.

```suggestion
        const { project } = options;
        // `ng update` invokes migrations with no options at all, and migrations.json declares no schema,
        // so the schema default never reaches us — applying the fix is the intended behaviour there.
        const fix = options.fix ?? true;
```

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+    if (!parsed.tree) return { content: template, changed: false };
+
+    const collector = new MarkdownRefCollector();
+
+    visitAll(collector, (parsed.tree as { rootNodes: unknown[] }).rootNodes);
+
+    if (collector.refs.size === 0) return { content: template, changed: false };
+
+    return rewriteRefReads(template, [...collector.refs]);
```
lskramarov: **The template rewrite is scope-blind: the AST is parsed, then thrown away.**

`migrateTemplate` uses the parse only to harvest bare ref names — `MarkdownRefCollector.refs` is a `Set<string>` carrying no span — and then hands the whole raw string to a regex. Two consequences, both reproduced by running the exact pattern:

1. **Any identically named `@for` / `let-` variable is rewritten.**
```html
<kbq-markdown #note [markdownText]="src" />
@for (note of notes; track note.id) {
  <span>{{ note.markdownText }}</span>
}
```
becomes `{{ note.markdownText() }}` over `notes: { markdownText: string }[]` → runtime `note.markdownText is not a function`. Same for an `<ng-template let-md>` shadowing a `#md` ref.

2. **Projected markdown prose and HTML comments are rewritten.** Unique to this component: `.kbq-markdown__input` is `display: none` and its `textContent` is fed to `marked`, so
```html
<kbq-markdown #md ngPreserveWhitespaces>
Set `md.markdownText` to render text.
</kbq-markdown>
```
silently edits the prose the user reads.

The spans are already on hand — `utils/typescript.ts:281-285` says the `HtmlParser` was chosen precisely so text-only ranges are available without reconstructing the source. Confining edits to interpolation/binding spans, and to the ref's declaring view, fixes all of it.

lskramarov (commented):
## `/code-review max` — 15 findings

Automated review of the full diff at `a46cfef7e`: 10 finder angles + a gap sweep, each finding verified individually. Correctness first, then test coverage and cleanup.

The headline is the first one: **as shipped, this migration writes nothing when run the way the PR advertises it** (`ng update`). `navbar-signals-and-aria` already fixed that exact trap and left a comment explaining it.

Several template-rewrite findings are inherited from the `alert-signals` clone rather than introduced here, but they ship again in this file and one of them (`(?!\s*=(?!=))`) was already solved in `navbar-signals-and-aria`.

Where a suggestion block appears, it is a complete fix. The rest describe the change rather than guess at it.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
```
lskramarov: **A template write is rewritten into an invalid assignment target.**

The pattern has no assignment guard, so

```html
<kbq-markdown #md /><button (click)="md.markdownText = null">clear</button>
```

becomes `(click)="md.markdownText() = null"`. Angular's expression parser rejects an assignment whose LHS is a call, so the consumer's app stops compiling — strictly worse than leaving it alone, and it contradicts this file's own TS-side policy at lines 150-151 ("leave the write untouched (it becomes a compile error the consumer fixes by hand)").

`navbar-signals-and-aria/index.ts:409-417` hit this and documents it verbatim; its `memberAccessPattern` carries a trailing `(?!\s*=(?!=))`. Verified side by side: with that guard the same input is left untouched.

```suggestion
        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()(?!\\s*=(?!=))`, 'g');
```

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+            edits.push({ start: node.getEnd(), end: rhs.getStart(sourceFile), text: '.set(' });
+            edits.push({ start: rhs.getEnd(), end: rhs.getEnd(), text: ')' });
+        }
+
+        return;
+    }
+
+    // Read (incl. optional chain `x?.compact`): append `()`.
+    edits.push({ start: node.getEnd(), end: node.getEnd(), text: '()' });
```
lskramarov, line 389: **Compound assignments fall through to the read branch and produce invalid TypeScript.**

The write guard above matches only `EqualsToken`, so `PlusEqualsToken`, `QuestionQuestionEqualsToken` and `PostfixUnaryExpression` reach this line. Verified by running this file's own `collectReceivers` / `classifyAccess` / `applyEdits`:

| input | output |
| --- | --- |
| `markdown.markdownText += '\n' + extra;` | `markdown.markdownText() += '\n' + extra;` |
| `markdown.markdownText ??= 'default';` | `markdown.markdownText() ??= 'default';` |

Both are TS2364 ("The left-hand side of an assignment expression must be a variable or a property access") — mangled source rather than the clean "compile error the consumer fixes by hand" that `data.ts:18-20` and the README promise. This is realistic here precisely because `markdownText` is `string | null`, unlike the booleans the sibling migrations were written for. `ts.isAssignmentExpression(parent)` (or checking the whole compound-operator range, plus `PrefixUnaryExpression`/`PostfixUnaryExpression`) covers it; `index.spec.ts:124-138` only tests `=`.

packages/components/markdown/markdown.component.ts
```diff
@@ -44,43 +45,31 @@ export const kbqMarkdownMarkedOptionsProvider = (options: MarkedOptions): Provid
...
 
     constructor() {
-        afterNextRender(() => {
-            const contentWrapper = this.contentWrapper();
-
-            if (!this.markdownText && contentWrapper?.nativeElement.textContent) {
-                this.resultHtml.set(this.getResultHTML(contentWrapper?.nativeElement.textContent));
-            }
-        });
+        afterNextRender(() => this.projectedText.set(this.contentWrapper().nativeElement.textContent));
```
lskramarov: **The projected fallback is frozen at first render, so the advertised "live fallback" is not live.**

`projectedText` is written on this line and nowhere else. Before this PR that was fine — the projected content was consulted only during that single `afterNextRender`. Now it is a permanent fallback source, so the snapshot becomes user-visible at arbitrary later times:

```html
<kbq-markdown [markdownText]="text()">{{ doc() }}</kbq-markdown>
```
First render snapshots the initial `doc()`. `doc.set('# new')` updates the hidden `<pre>` but not `projectedText`; a later `text.set(null)` then renders the **stale** original. Worse with `<kbq-markdown>@if (ready()) { {{ text() }} }</kbq-markdown>`: `projectedText` freezes at `''` before `ready()` flips, so clearing the input renders nothing, permanently.

This is the dominant shape — 10 of the 11 `<kbq-markdown>` call sites under `packages/docs-examples` project content. The "live fallback" claim appears in `README.md:47`, `data.ts` `SUMMARY`, and both migration guides, and the new spec only covers the one direction that works (input goes empty → first-render projection reappears).

`afterRenderEffect` is exported by the pinned `@angular/core` 20.3.30 and already used in `packages/components/skeleton/skeleton.ts:54`; a `MutationObserver` on `contentWrapper` is the other option. Either makes the claim true — or the docs should say "captured once at first render".

packages/schematics/src/migrations/markdown-signals/data.ts
```diff
@@ -0,0 +1,77 @@
...
+        anchor: SUBCLASS_ANCHOR,
+        pattern: '\\bresultHtml\\s*\\.\\s*set\\b',
+        message:
+            'KbqMarkdown.resultHtml is a read-only `computed` instead of a `WritableSignal`. It derives from ' +
+            '`markdownText` and the projected content, so a subclass that used to push HTML into it has to feed ' +
+            'the input instead.'
+    },
+    {
+        anchor: '\\bKbqMarkdown\\b',
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqMarkdown\\b',
```
lskramarov: **This warning fires on the decorator form the migration already fixes correctly, and its advice is wrong for that form.**

Verified: the pattern matches `@ViewChild(KbqMarkdown) markdown: KbqMarkdown;`. `collectReceivers` picks that property up from its explicit annotation and Pass A correctly rewrites `this.markdown.markdownText` → `this.markdown.markdownText()` — the exact fixture asserted at `index.spec.ts:73-89`. That same run then prints "reading it is a double call, e.g. `this.markdown().markdownText()`", which does not compile for a decorator query, since the field holds the instance rather than a signal. The user is told to hand-verify a case the tool got right.

The alternation also matches `viewChildren` / `ContentChildren` by prefix, where the query returns an array rather than "the component instance" the message describes.

Restricting it to the signal spellings — `(?:viewChild|contentChild)(?:\.required)?\s*\(` — covers the real case. Worth adding a test that asserts the warning is *absent* for the decorator fixture.

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return { content, changed };
+}
+
+/** Pass B (core) — parse a template, discover markdown refs, rewrite their value-safe signal reads. */
+async function migrateTemplate(template: string): Promise<{ content: string; changed: boolean }> {
+    if (!template.includes(MARKDOWN_ELEMENT)) return { content: template, changed: false };
+
+    const parsed = await parseTemplate(template);
+
+    if (!parsed.tree) return { content: template, changed: false };
```
lskramarov: **An unparseable template is skipped silently.**

If a consumer's `app.component.html` has `<kbq-markdown #md/>{{ md.markdownText }}` plus an unbalanced tag elsewhere, `parseTemplate` returns `{ tree: undefined, errors: [...] }` (`utils/typescript.ts:296-300`), this line returns unchanged, `consumers` is not incremented, and nothing is logged. Their build then fails on `md.markdownText` with no hint that the schematic skipped the file.

Seven siblings — `app-switcher-signals`, `navbar-signals-and-aria`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `list-tree-multiple-input` — define and emit `UNPARSEABLE_TEMPLATE_MESSAGE` on exactly this branch. `markdown-signals` defines no such message.

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+        node.forEachChild(visit);
+    };
+
+    visit(sourceFile);
+
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
```
lskramarov: **Receivers are matched by source text over the whole enclosing function, so a shadowing binding of another type is rewritten too.**

Verified by running this file's own logic:

```ts
interface Other { markdownText: string }

read(markdown: KbqMarkdown, items: Other[]) {
    return items.map((markdown: Other) => markdown.markdownText);
}
```
→ `markdown.markdownText()` inside the arrow. The inner `markdown` is an `Other`, so the consumer's build breaks with "This expression is not callable". The receiver's scope is the whole outer method (`findAncestor(node, isFunctionLike)`) and identity is `node.expression.getText(sourceFile) === r.text`, so any shadowing callback parameter, catch binding, or later `const markdown` of a different type inside that method is caught.

The same text-only design misses every inferred-type receiver (`const m = this.markdownQuery(); m.markdownText`), which is why `data.ts:62` has to fall back to a "verify manually" warning. A `ts.Program` pass is reachable in this package — `autocomplete-panel-width-auto/index.ts:28`, `empty-state-size-attr/index.ts:32` and `loader-overlay-size-attr/index.ts:32` all build one, and `utils/typescript.ts:110` already exports `canMigrateFile(sourceFile, program)`.

At minimum, narrowing a receiver's scope to start at its declaration and dropping it when an inner declaration shadows the name would remove the false positive.

packages/components/markdown/markdown.component.ts
```diff
@@ -44,43 +45,31 @@ export const kbqMarkdownMarkedOptionsProvider = (options: MarkedOptions): Provid
...
 
         effect(() => {
             if (this.resultHtml()) {
```
lskramarov: **The deferred `startMonitoringLinks()` can run after `ngOnDestroy`, leaking FocusMonitor registrations.**

The effect body queues `Promise.resolve().then(() => this.startMonitoringLinks())` (line 76) with nothing to cancel it. If the component is destroyed in the same task — an overlay/dialog closing after `appRef.tick()`, a synchronous router redirect, or `fixture.detectChanges(); fixture.destroy();` — `ngOnDestroy` runs `stopMonitoringLinks()` against a still-empty `links`, and the microtask then drains against the dead instance: it reads `this.outputWrapper()` (a `viewChild.required` — a stale `ElementRef`, or NG0951 inside a swallowed rejection), pushes the detached anchors into `this.links` and calls `focusMonitor.monitor(link, true)`. `FocusMonitor` is root-provided and its `_elementInfo` map is cleared only by `stopMonitoring`, so the detached subtree is retained for the life of the app.

The effect body predates this PR, but the constructor is rewritten here and the change widens the window: `resultHtml` can now go truthy → `null` → truthy, where before it was written at most once. An `onCleanup` callback on the effect, or a `DestroyRef.destroyed` guard inside `startMonitoringLinks`, closes it.

packages/schematics/src/collection.json
```diff
@@ -143,6 +143,11 @@
...
             "factory": "./migrations/split-button-optional-disabled/index",
             "schema": "./migrations/split-button-optional-disabled/schema.json"
+        },
+        "markdown-signals": {
+            "description": "Migrates KbqMarkdown consumers to its finished signal-based API. markdownText was the last accessor input in the package - its setter did the rendering, which is why the automated signal migration skipped it. Rewrites programmatic reads of markdownText to calls on receivers typed KbqMarkdown, including reads via template reference variables on <kbq-markdown> (external and inline templates). Warns on resultHtml, which became protected, and on view/content queries that return the instance. Reports the two behavior fixes the review made: clearing markdownText now clears the output instead of leaving the last render on screen, and a component that both projects content and binds the input falls back to the projected content once the input goes empty.",
```
lskramarov: **The description promises a `resultHtml` warning that cannot fire.**

> Warns on resultHtml, which became protected, and on view/content queries that return the instance.

`resultHtml` did not become protected — `git show 4bb1c9927:packages/components/markdown/markdown.component.ts` already has `protected resultHtml = signal<SafeHtml | null>(null);`, and the api.md diff only changes `WritableSignal` → `Signal`. `data.ts:33` says so itself ("`resultHtml` was already `protected`, so nothing a consumer can name changed visibility") and sets `PROTECTED_MEMBERS = []`, which makes `collectProtectedAccess` always return an empty set and `warnReceiverMembers` never log.

`ng update` prints these descriptions, so a consumer with a subclass that *reads* `this.resultHtml` is told the tool will flag it, sees nothing, and concludes they are unaffected. The only `resultHtml` warning that exists requires both `extends KbqMarkdown` **and** a literal `.set` call. The identical sentence is duplicated at `packages/schematics/src/migrations.json:111` and needs the same edit.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **The new "two waves" sentence contradicts the subsections it introduces.**

Section 18 has exactly four subsections at this commit: `#### Markdown` (1052), `#### Search expandable` (1070), `#### Split button` (1092), `#### Title` (1114). But the sentence says the first wave covered "…search-expandable, select, split-button, title…" and then "the second is the one each subsection below belongs to" — so three of the four subsections are claimed for both waves. Only Markdown is new here.

The pre-PR wording ("Ten components went through a full review in 20.3.0: …") was consistent; the rewrite broke it while adding one subsection. Same defect at `docs/guides/migration.ru.md:1042` ("ко второй относится каждый из подразделов ниже").

packages/schematics/src/migrations/markdown-signals/index.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+                '    read() {\n' +
+                '        return this.markdown.markdownText;\n' +
+                '    }\n' +
+                '}\n'
+        );
+
+        expect((await run()).readText(ts)).toContain('return this.markdown.markdownText();');
+    });
+
+    it('leaves reads on a receiver of an unrelated type alone', async () => {
```
lskramarov, line 93: **This test never reaches the code it names.**

The fixture declares no `KbqMarkdown`-typed receiver, so `collectReceivers` returns `[]` (verified by running it over this exact source) and `migrateTsExpressions` returns at `if (receivers.length === 0) return content;` — before `collectAccessEdits` / `inReceiverScope` are ever entered. The test would still pass if `inReceiverScope` returned `true` unconditionally, so the receiver-discrimination path is untested.

The `alert-signals` test this was cloned from (`index.spec.ts:135`) deliberately puts a real `alert: KbqAlert` parameter in scope alongside the unrelated read, which is what makes it exercise the discrimination. Adding a `markdown: KbqMarkdown` parameter to `Demo.read` next to `other: Other` restores that.

packages/schematics/src/migrations/markdown-signals/index.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+            ts,
+            "import { Component } from '@angular/core';\n" +
+                "@Component({ template: '<kbq-markdown #markdown></kbq-markdown>{{ markdown.markdownText }}' })\n" +
+                'class Demo {}\n'
+        );
+
+        expect((await run()).readText(ts)).toContain('{{ markdown.markdownText() }}');
+    });
+
+    it('leaves a template reference on an unrelated element alone', async () => {
```
lskramarov, line 166: **This test also short-circuits before the code it names.**

The fixture `'<other-thing #markdown></other-thing>\n<span>{{ markdown.markdownText }}</span>\n'` contains no `kbq-markdown` substring, so `migrateTemplate` returns on its first line (`if (!template.includes(MARKDOWN_ELEMENT))`) and neither `parseTemplate` nor `MarkdownRefCollector` runs. Deleting `if (element.name === MARKDOWN_ELEMENT)` from `visitElement` leaves this test green.

Putting a real `<kbq-markdown #other />` in the same template so the parser is reached, then asserting the `#markdown` ref on `<other-thing>` is still not rewritten, would actually cover the element-name check.

packages/schematics/src/migrations/markdown-signals/data.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/** Import specifier that marks a file as a markdown consumer. */
+export const MARKDOWN_PACKAGE = '@koobiq/components/markdown';
+
+/**
+ * Members that moved from `public` to `protected`. `resultHtml` was already `protected`, so nothing a
+ * consumer can name changed visibility — the list is empty and the machinery stays for symmetry with the
+ * sibling migrations.
+ */
+export const PROTECTED_MEMBERS: readonly string[] = [];
```
lskramarov: **Both of these tables are empty by construction, leaving ~60 lines that no input can execute.**

`PROTECTED_MEMBERS.includes(...)` on `[]` is always false, so `index.ts:236` (`if (protectedAccess.size > 0)`) is unreachable; `WRITABLE_MEMBERS.has(...)` on `new Set()` is always false, so `index.ts:157-162` is unreachable. Dead: `data.ts:17-21`, `32-40`; `index.ts:157-162`, `194-213`, `228-243`.

It is not merely inert — `index.ts:432` calls `warnReceiverMembers` *before* `migrateTsExpressions` on the same string, so every consumer `.ts` file pays a second `ts.createSourceFile(..., setParentNodes = true)`, a duplicate `collectReceivers` walk, and a `collectProtectedAccess` walk that looks names up in an empty array.

The comment above says the machinery "stays for symmetry with the sibling migrations", but the symmetry points the other way: `app-switcher-signals` and `filter-bar-signals` carry no `PROTECTED_*` symbol at all, and the siblings that do write carry non-empty `WRITABLE_MEMBERS` (`new Set(['selectedApp'])`, `new Set(['filter'])`). Deleting both tables and their consumers *is* the symmetric choice.

packages/schematics/src/migrations/markdown-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+    visitAll(collector, (parsed.tree as { rootNodes: unknown[] }).rootNodes);
+
+    if (collector.refs.size === 0) return { content: template, changed: false };
+
+    return rewriteRefReads(template, [...collector.refs]);
+}
+
+/** Interior `[start, end]` ranges of inline `@Component({ template: '…' })` string literals. */
+function collectInlineTemplateRanges(sourceFile: ts.SourceFile): Array<{ start: number; end: number }> {
```
lskramarov: **This duplicates the shared helper the file already imports from.**

`diff <(sed -n '319,353p' index.ts) <(sed -n '74,108p' ../../utils/typescript.ts)` reports exactly one differing line — the `export` keyword. Line 7 already reads `import { forEachClass, parseTemplate } from '../../utils/typescript';`, and `forEachClass` is imported *only* to feed this copy.

Seven siblings already import the shared one (`app-switcher-signals`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `list-tree-multiple-input`, `navbar-signals-and-aria`); this raises the private-fork count from 5 to 6. A future fix there — e.g. handling a `template` given as a template literal with substitutions, which both copies currently skip — silently misses this migration, and the local copy drops the shared version's JSDoc contract.

Swapping line 7 to `import { collectInlineTemplateRanges, parseTemplate } from '../../utils/typescript';` and deleting lines 318-353 is the whole change.

packages/schematics/src/migrations/markdown-signals/index.ts
artembelik: Сделал минимальный вариант из твоего же комментария - разрешение имени через объявления вместо прохода по `ts.Program`. Ресиверы из `inject()` и `viewChild()` теперь тоже распознаются, но `const m = query(); m.markdownText` по-прежнему мимо.

packages/components/markdown/markdown.component.ts
artembelik: `afterRenderEffect` тут не помогает: реактивных зависимостей у него нет, `textContent` не сигнал, и он отработал бы один раз. Взял второй вариант - `ContentObserver` из CDK, он же идиома репозитория.

packages/schematics/src/migrations/markdown-signals/data.ts
artembelik: Таблицу и ветку `.set(…)` убрал, а детекцию записи оставил - без неё вернулось бы замечание про составные присваивания. Вместо таблицы теперь `SIGNAL_API_METHODS`, который реально используется.

docs/guides/migration.en.md
artembelik: Взял формулировку, которая уже лежит в main без деления на волны: свой вариант дал бы конфликт и откат к ней на ближайшем мерже.

lskramarov (approved):

PR #1982, opened 2026-09-02
fix(loader-overlay)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `loader-overlay`, in the same shape as the 20.3.0 component reviews.

### `text` and `caption` lied about being required

```ts
@Input() text: string; // no initializer
@Input() caption: string;
```

The automated signal migration skipped both — it saw them read inside `@if` blocks and would not risk the narrowing. They are `input<string>()` now and report `string | undefined`, which is what they always held. Same pathology as `KbqSplitButton.disabled` in the first review wave: the call sites that were already wrong now fail to compile.

### `transparent` was backwards as a valueless attribute

`<kbq-loader-overlay transparent>` passed the empty string, which is falsy, so the attribute rendered the **filled** background — the opposite of how it reads. With `booleanAttribute` it means `true`, and `[transparent]="'false'"` means `false`.

### Closed internals

Everything the template uses to choose between a projected slot and an input is `protected` or `private` now: `isEmpty`, `isExternalIndicator`, `isExternalText`, `isExternalCaption`, `spinnerSize`, and the three content queries (which are signal queries as well). What the overlay renders is the contract, not how it decides.

The three getters became `computed`, so `isEmpty` now tracks a `[text]` that arrives later instead of relying on whoever happened to run change detection.

## Migration

`loader-overlay-signals` runs from `ng update @koobiq/components@20`. It rewrites `text` and `caption` reads to calls — on receivers typed `KbqLoaderOverlay` and through template reference variables on `<kbq-loader-overlay>`, in external and inline templates — and reports the rest. `size`, `transparent` and `card` were already signals in 20.2.0 and are not touched.

## Testing

- `loader-overlay.component.spec.ts`: 5 → 8 tests. The new ones pin the valueless `transparent` attribute, `isEmpty` reacting to a late `[text]`, and the unbound inputs reporting `undefined`. The existing class snapshot is unchanged.
- `loader-overlay-signals/index.spec.ts`: 13 tests — auto-fix, idempotence, receiver scoping, template refs, warnings, and the `--fix=false` path.
- Full `packages/components` (4996 tests) and `packages/schematics` (447 tests) suites pass.
- `check-api` is in sync.

No e2e screenshots were regenerated: every e2e case binds `[transparent]="false"` explicitly, so nothing it renders changed.

BREAKING CHANGE


packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return (
+        /\bKbqLoaderOverlay\w*\b/.test(content) ||
+        content.includes(OVERLAY_PACKAGE) ||
+        content.includes(`<${OVERLAY_ELEMENT}`)
+    );
+}
+
+export default function loaderOverlaySignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **The auto-fix never runs under `ng update` — `fix` arrives `undefined`.**

`migrations.json` declares no `"schema"` for any entry (0 occurrences in the file), and the CLI invokes each migration with no options at all — `@angular/cli/src/commands/update/cli.js:329` calls `this.executeSchematic(workflow, migration.collection.name, migration.name)` against the signature `executeSchematic(workflow, collection, schematic, options = {})`. So `schema.json`'s `"fix": { "default": true }` never reaches the rule.

A consumer runs `ng update @koobiq/components@20`, every `overlay.text` / `overlay.caption` read stays unmigrated, and the log fills with `[loader-overlay-signals] would update <file> (run with --fix to apply)`. Meanwhile `README.md:3` ("invoked automatically by `ng update`") and `docs/guides/migration.en.md` ("the `text` and `caption` reads are rewritten") promise the opposite.

12 of the 18 sibling migrations already guard this, and `app-switcher-signals/index.ts:412` spells out why:

```ts
```

`index.spec.ts:37` is `async function run(fix: boolean = true)` and always passes `fix` explicitly, so no test covers the `ng update` call shape.

```suggestion
        const { project } = options;
        const fix = options.fix ?? true;
```

tools/public_api_guard/components/loader-overlay.api.md
```diff
@@ -4,84 +4,70 @@
...
+    protected readonly isExternalText: _angular_core.Signal<boolean>;
+    protected readonly loaderSizeClass: _angular_core.Signal<string>;
     // (undocumented)
     ngOnDestroy(): void;
     // (undocumented)
     ngOnInit(): void;
+    readonly size: _angular_core.InputSignal<KbqDefaultSizes>;
+    protected readonly spinnerSize: _angular_core.Signal<ProgressSpinnerSize>;
+    readonly text: _angular_core.InputSignal<string | undefined>;
+    readonly transparent: _angular_core.InputSignalWithTransform<boolean, unknown>;
```
lskramarov, line 36: **`yarn run check-public-api-any` fails on this line — the baseline was not bumped.**

The `booleanAttribute` transform turns `transparent` into `InputSignalWithTransform<boolean, unknown>`, a second `unknown` in the hand-written surface. `tools/check-public-api-any/baseline.json` still records `"loader-overlay": 1`.

I ran the tool's own filter (`tools/check-public-api-any/index.ts:31-35` — `\b(any|unknown)\b` on non-generated, non-comment lines) over both revisions:

| revision | count |
| --- | --- |
| base `4bb1c9927` | 1 (`card`) |
| this branch | 2 (`card` line 18, `transparent` line 32) |

`.github/workflows/api.yml:20` runs `check-public-api-any` on every `pull_request`, so the API job exits 1 with `The published type surface gained \`any\` / \`unknown\`: loader-overlay: 1 → 2`.

```bash
yarn run approve-public-api-any
```

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
```
lskramarov: **The template rewrite is a raw `String.replace` over the whole file, so it corrupts unrelated `.text` / `.caption` reads.**

The parse is used only to *discover* ref names; the replacement then ignores the AST entirely. And `\b` matches after a `.`, so the comment above ("scoped to those exact identifiers") is not true — any path whose **last** segment equals the ref name matches.

I ran this exact regex in node against a template declaring `<kbq-loader-overlay #overlay />`. Every one of these was mutated and written to disk under `--fix`:

| input | output |
| --- | --- |
| `@for (overlay of overlays; track overlay.id) { {{ overlay.text }} }` | `{{ overlay.text() }}` — `TypeError: overlay.text is not a function` |
| `{{ vm.overlay.text }}` | `{{ vm.overlay.text() }}` — unrelated view-model |
| `<p>Set overlay.text to change the label.</p>` | `overlay.text()` in the rendered copy |
| `<!-- overlay.text is the input -->` | `<!-- overlay.text() is the input -->` |
| `<a href="/api#overlay.text">` | broken link |
| `<span data-doc="overlay.text">` | corrupted static attribute |

`SIGNAL_MEMBERS` here is `['text', 'caption']` — two of the most common property names in any codebase — which makes this far likelier than for `alert-signals`' `compact` / `alertStyle`. Anchoring with `(?<![.\w$])` before the ref would kill the `vm.overlay.text` case; the loop-variable and text-node cases need the rewrite to walk binding expressions rather than the raw string.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    // Already migrated: a call, or a `.set(...)` write — leave alone (idempotent).
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.member = RHS`. Every KbqLoaderOverlay signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
```
lskramarov: **Compound assignments are rewritten into code that does not parse.**

Only `EqualsToken` is treated as a write; every other assignment form falls through to the read branch on line 168 and gets `()` appended to the left-hand side:

| source | after migration |
| --- | --- |
| `overlay.text += ' (retrying)'` | `overlay.text() += ' (retrying)'` |
| `overlay.caption ??= 'Loading'` | `overlay.caption() ??= 'Loading'` |
| `overlay.text++` (`PostfixUnaryExpression` parent — not checked at all) | `overlay.text()++` |

All three are `TS2364` / `Invalid left-hand side in assignment`. The plain `=` case is deliberately left alone so the consumer gets a clean read-only error; the compound case turns a type error into a file the compiler cannot parse, at a line the schematic itself wrote. `index.spec.ts:124` covers only the plain `=` form.

Widening the guard to `ts.isAssignmentExpression(parent)` (or just `parent.left === node` for any binary assignment operator) plus a `PrefixUnaryExpression`/`PostfixUnaryExpression` check would return them all to "leave alone".

Separately, the comment three lines down still reads ``// Read (incl. optional chain `x?.compact`): append `()`.`` — `compact` is a `KbqAlert` member carried over from `alert-signals`; every other sibling adapted the example to its own member.

packages/schematics/src/migrations/loader-overlay-signals/data.ts
```diff
@@ -0,0 +1,77 @@
...
+    /** Owner of the member. The pattern is only evaluated for files that also name it. */
+    anchor: string;
+    pattern: string;
+    message: string;
+}
+
+export const warnPatterns: WarnPattern[] = [
+    {
+        anchor: '\\bKbqLoaderOverlay\\b',
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqLoaderOverlay\\b',
```
lskramarov: **This warning fires on the form the migration already fixed, and misses the form it is actually about.**

I ran the pattern in node:

| input | matches |
| --- | --- |
| `@ViewChild(KbqLoaderOverlay) loaderOverlay: KbqLoaderOverlay;` | **yes** |
| `readonly overlay = viewChild(KbqLoaderOverlay);` | yes |
| `readonly overlay = viewChild(\n    KbqLoaderOverlay\n);` | **no** |
| `@ViewChild(\n    KbqLoaderOverlay\n)\nreadonly overlay!: KbqLoaderOverlay;` | **no** |

The decorator form returns the **instance**, so the correct read is `this.overlay.text()` — which is exactly what `collectReceivers` + `classifyAccess` already produced (the declaration carries an explicit `: KbqLoaderOverlay` annotation, so it is a recognised receiver). The user reads "reading one is a double call, e.g. `this.overlay().text()`", edits the auto-fixed line, and gets `TS2349: This expression is not callable`. `index.spec.ts:73-89` feeds precisely that source and asserts only the rewrite, never that the warning stays silent — so the false warning ships untested.

Meanwhile `[^\n;]*` cannot cross a newline, so with `printWidth: 120` a real `viewChild(KbqLoaderOverlay, { read: ElementRef, descendants: true })` wraps and goes unwarned — the one case the message is right about.

Restricting to the lowercase signal-query forms and allowing newlines fixes both: `'(?:viewChild|contentChild)\\s*(?:\\.required)?\\s*\\([\\s\\S]{0,200}?\\bKbqLoaderOverlay\\b'`.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+        for (const filePath of htmlPaths) {
+            const original = tree.read(filePath)?.toString();
+
+            if (!original) continue;
+
+            const { content, changed } = await migrateTemplate(original);
+
+            if (changed) {
+                consumers++;
```
lskramarov: **The one silent breaking change gets no call-site warning — and this gate can drop it entirely.**

`transparent` moving to `booleanAttribute` is the only change in this PR that produces no compile error: `<kbq-loader-overlay transparent>` used to pass `''` (falsy → `_filled`) and now means `true` (→ `_transparent`). Everything else surfaces at `tsc`; this one changes pixels only.

Nothing scans templates for the attribute. `warnPatterns` has a single entry and `logWarnings` is called only in the `.ts` loop (line 431), so the change appears only in the once-per-project `SUMMARY`, naming zero files.

Worse, this `consumers++` sits behind `if (changed)`: an `.html` file counts as a consumer only when a `#ref.text` / `#ref.caption` rewrite actually landed. A project whose overlay usage is entirely `<kbq-loader-overlay transparent>` in external templates, with the module reaching the component through a shared barrel re-export (so no `.ts` matches `referencesLoaderOverlay`), hits `if (consumers === 0) return;` on line 454 and the `transparent` note is **not printed at all**.

The machinery is one `if` away — `LoaderOverlayRefCollector.visitElement` (lines 249-253) already iterates `element.attrs` on exactly this element, and `list-tree-multiple-input/index.ts:54-93` (same release) does per-file attribute findings for the same class of change.

Two gaps in the `SUMMARY` text itself (`data.ts:74-76`): `[transparent]="0"` and `NaN` also flip false → true, and `[transparent]="'false'"` used to mean **true** — the note says what it means now, not what it changed from.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.member = RHS`. Every KbqLoaderOverlay signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
+    ) {
+        if (WRITABLE_MEMBERS.has(node.name.text)) {
```
lskramarov: **A programmatic write produces no diagnostic at all, though three documents say it is reported.**

`WRITABLE_MEMBERS` is `new Set<string>()` (`data.ts:23`), so this `has()` is unconditionally false and the two `edits.push` calls below are unreachable. The `return` is load-bearing (it stops `x.text = 'y'` becoming `x.text() = 'y'`), but nothing is logged on the way out — `text`/`caption` are not in `PROTECTED_MEMBERS`, so `warnReceiverMembers` says nothing, and `warnPatterns` has only the view-query entry.

So `overlay.text = 'Loading'` yields zero output naming the file. The consumer sees `updated 0 file(s)` and a clean log, then discovers each assignment site from `tsc` (`Cannot assign to 'text' because it is a read-only property`) with no list to work from.

That contradicts `collection.json` / `migrations.json` ("Warns on … Reports …"), `README.md:40` (`overlay.text = …` → "Bind `[text]` in the template") and `docs/guides/migration.en.md` ("the `text` and `caption` reads are rewritten, **the rest is reported**"). `index.spec.ts:124-138` asserts only that the source is unchanged, never that a warning is emitted — so the gap is untested.

Simplest fix: drop `WRITABLE_MEMBERS` and log the write instead of silently returning.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
+        const next = content.replace(pattern, '$1.$2()');
```
lskramarov: **Template refs read through `?.` are silently left unmigrated — and fail silently at runtime.**

Verified in node: `{{ overlay?.text }}` → `changed = false`, left untouched. The TypeScript pass deliberately handles the optional chain (the comment on line 167 says so, and `index.spec.ts:63` tests `loaderOverlay?.caption`), so the two passes disagree about the same construct.

Interpolating a function is legal TypeScript, so Angular's strict template checker says nothing; `ɵɵtextInterpolate` → `renderStringify` then prints the `InputSignal` function's source into the page instead of the string. The migration reports the file as clean.

The pattern needs to admit the optional token: `` `\\b(${escapeRegExp(ref)})\\??\\.(${members})\\b(?!\\s*\\()` `` with `'$1?.$2()'` handling, or a capture group over `\\??\\.`.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return { content, changed };
+}
+
+/** Pass B (core) — parse a template, discover loaderOverlay refs, rewrite their value-safe signal reads. */
+async function migrateTemplate(template: string): Promise<{ content: string; changed: boolean }> {
+    if (!template.includes(OVERLAY_ELEMENT)) return { content: template, changed: false };
+
+    const parsed = await parseTemplate(template);
+
+    if (!parsed.tree) return { content: template, changed: false };
```
lskramarov: **An unparseable template is skipped in total silence.**

`parseTemplate` (`utils/typescript.ts:296-305`) returns `{ tree: undefined, errors }` both for HTML parse errors **and** from its `catch`, which covers the dynamic `await import('@angular/compiler')` failing. Either way this line drops the file with no message.

Failure: an external template contains `<kbq-loader-overlay #o />… {{ o.text }}` plus one unrelated unclosed tag elsewhere. The file is skipped, nothing is logged for it, and the run reports success — `{{ o.text }}` then renders the stringified signal function with no diagnostic pointing at the file. If `@angular/compiler` is unresolvable in the consumer's install, *every* template is skipped and the run still says `updated 0 file(s)`.

The sibling this was cloned from keeps that channel: `app-switcher-signals` threads an `unparseable` flag through `migrateTemplate` / `migrateInlineTemplates` and logs `UNPARSEABLE_TEMPLATE_MESSAGE` (`index.ts:459-461`, `475-477`; `data.ts:135`).

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    const receivers = collectReceivers(sourceFile, OVERLAY_TYPE);
+
+    if (receivers.length === 0) return;
+
+    const protectedAccess = collectProtectedAccess(sourceFile, receivers);
+
+    if (protectedAccess.size > 0) {
+        logMessage(context.logger, [
+            `${LABEL} ${filePath}`,
+            `  These KbqLoaderOverlay members are now \`protected\` and can't be read from outside the ` +
```
lskramarov: **This message says `protected` for three members that are `private`.**

`PROTECTED_MEMBERS` (`data.ts:41-43`) lumps `externalIndicator` / `externalText` / `externalCaption` in with the genuinely `protected` helpers, but at `loader-overlay.component.ts:68-70` all three are:

```ts
private readonly externalIndicator = contentChild(KbqLoaderOverlayIndicator);
private readonly externalText = contentChild(KbqLoaderOverlayText);
private readonly externalCaption = contentChild(KbqLoaderOverlayCaption);
```

A consumer reads the log, concludes the queries survive for subclasses, refactors toward `class MyOverlay extends KbqLoaderOverlay`, and hits `TS2341: Property 'externalText' is private`. Both migration guides get this right ("Now `private` signal queries") — only the schematic's own output is wrong.

`PROTECTED_HINT` has the same problem: "the template helpers behind the projection slots … what the overlay renders is the contract" describes `isExternal*` / `isEmpty` / `spinnerSize`, not the content queries. Splitting the list into two sets with two messages fixes both.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+            warnReceiverMembers(context, filePath, original);
+
+            let content = migrateTsExpressions(original, filePath);
+
+            content = await migrateInlineTemplates(content, filePath);
+
+            commit(filePath, original, content);
+        }
+
+        for (const filePath of htmlPaths) {
```
lskramarov, line 934: **HTML files get no warning pass at all.**

`logWarnings` and `warnReceiverMembers` are called only inside the `tsPaths` loop (lines 431-432) and both operate on the TypeScript AST, so they can never see a template read. This loop calls only `migrateTemplate`, which rewrites `SIGNAL_MEMBERS` (`text`, `caption`) and nothing else.

So a template with `<kbq-loader-overlay #o />` and `{{ o.isEmpty }}` or `[size]="o.spinnerSize"` is neither rewritten nor reported:

- with `strictTemplates`, the consumer's build fails with `TS2445` on a member the migration never mentioned;
- without it, `{{ o.isEmpty }}` now renders the `computed()` function's source text instead of `true`/`false`, because the getter became a signal.

The ref collector already resolves which refs point at an overlay — the same list could drive a `PROTECTED_MEMBERS` scan over the template text and emit the same per-file warning the `.ts` path gets. The same gap applies to inline templates.

lskramarov (commented):
Automated review of the loader-overlay signals rework and the new `loader-overlay-signals` schematic. 15 findings, most severe first; each is anchored inline. The component change itself is sound — almost everything below is in the migration schematic, plus one CI gate.

**Two that block CI / make the migration a no-op:**
1. `fix` is `undefined` under `ng update`, so the schematic rewrites nothing.
2. `tools/check-public-api-any/baseline.json` still records `"loader-overlay": 1`; the `booleanAttribute` transform makes it 2, and `.github/workflows/api.yml` runs that check on every PR.

Everything else is inline.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    };
+
+    visit(sourceFile);
+
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
```
lskramarov: **Receiver scope is text equality over the whole enclosing function, so a shadowing binding is rewritten.**

`collectReceivers` registers a parameter's scope as the entire enclosing function (`findAncestor(node, isFunctionLike)`), and this compares `node.expression.getText(sourceFile)` against the receiver text within that span — there is no binding resolution.

```ts
render(overlay: KbqLoaderOverlay) {
    this.rows.forEach((overlay) => console.log(overlay.text));
}
```

The arrow parameter shadows the typed one, but its read sits inside the outer method's span and matches by text → rewritten to `overlay.text()` → `TypeError` on a plain row object. Same for a nested `function` or `const` of the same name.

Narrower than the template case (#3) since it needs a real name collision, but `text` / `caption` on a variable named after the component is a plausible pairing. Trimming the receiver scope when a nested declaration re-binds the same identifier would close it.

packages/components/loader-overlay/loader-overlay.component.ts
```diff
@@ -51,69 +54,65 @@ export class KbqLoaderOverlayCaption {}
...
-    @ContentChild(KbqLoaderOverlayText) externalText: KbqLoaderOverlayText | null;
-    @ContentChild(KbqLoaderOverlayCaption) externalCaption: KbqLoaderOverlayCaption | null;
+    /** @docs-private */
+    protected readonly loaderSizeClass = computed(() => `kbq-loader-overlay_${this.size()}`);
 
     ngOnInit(): void {
-        this.parent = this.elementRef.nativeElement.parentElement;
+        this.parent = this.nativeElement.parentElement;
 
         this.renderer.addClass(this.parent, kbqLoaderOverlayParent);
```
lskramarov: **`parent` is typed `HTMLElement | null` and neither hook guards it.**

`DefaultDomRenderer2.addClass` is `el.classList.add(name)` (`@angular/platform-browser` `dom_renderer.mjs:617`) — it throws on `null`, it is not a no-op.

- `const ref = createComponent(KbqLoaderOverlay, { environmentInjector }); ref.changeDetectorRef.detectChanges();` before the host node is appended — `parentElement` is `null`, `ngOnInit` throws `TypeError: Cannot read properties of null (reading 'classList')`.
- `const ref = vcr.createComponent(KbqLoaderOverlay); ref.destroy();` with no CD in between — `ngOnInit` never ran, `parent` is still `null`, and `ngOnDestroy` throws in `removeClass`.

Pre-existing, but this hook is rewritten in the diff (`elementRef.nativeElement` → `kbqInjectNativeElement()`) and the PR is billed as a full review, so it is signed off now. An early `if (!this.parent) return;` in both hooks covers it.

packages/schematics/src/migrations/loader-overlay-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+    visitAll(collector, (parsed.tree as { rootNodes: unknown[] }).rootNodes);
+
+    if (collector.refs.size === 0) return { content: template, changed: false };
+
+    return rewriteRefReads(template, [...collector.refs]);
+}
+
+/** Interior `[start, end]` ranges of inline `@Component({ template: '…' })` string literals. */
+function collectInlineTemplateRanges(sourceFile: ts.SourceFile): Array<{ start: number; end: number }> {
```
lskramarov: **This is a byte-identical copy of an export the file already imports from.**

`packages/schematics/src/utils/typescript.ts:74` exports `collectInlineTemplateRanges`; `diff` between the two reports exactly one differing line — the `export` keyword. And line 7 already reads `import { forEachClass, parseTemplate } from '../../utils/typescript';`, where `forEachClass` is imported **only** to feed this copy (its sole use is line 322).

Seven migrations already import the shared one (`app-switcher-signals`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `list-tree-multiple-input`, `navbar-signals-and-aria`), as does `utils/icon-migration.ts:603`.

Fix: change line 7 to `import { collectInlineTemplateRanges, parseTemplate } from '../../utils/typescript';` and delete lines 318-353. −36 lines, zero behavior change — and the shared version documents a guarantee this copy silently won't inherit ("A template literal with a `${…}` substitution is skipped").

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **This sentence contradicts three of the four subsections it introduces.**

Section 18's subsections are `#### Loader overlay` (1052), `#### Search expandable` (1069), `#### Split button` (1091) and `#### Title` (1113). Search-expandable, split-button and title are all named in the *first*-wave list in this same sentence, yet "the second is the one each subsection below belongs to" assigns them to wave two. Loader overlay is named in neither list.

A reader migrating `KbqSplitButton` is told split-button was reviewed in wave one and, one clause later, that the `#### Split button` subsection belongs to wave two. And section 18 keeps growing — it went from 3 to 4 subsections in this PR — so every future addition either re-breaks the sentence or forces another rewrite of it. Wave membership is per-subsection data being encoded in a shared intro.

The original one-liner with loader-overlay added is self-consistent and has nothing to go stale:

```suggestion
Eleven components went through a full review in 20.3.0: loader-overlay, notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```

`docs/guides/migration.ru.md:1042` carries the identical defect ("ко второй относится каждый из подразделов ниже") and needs the mirrored change.

packages/schematics/src/migrations/loader-overlay-signals/data.ts
artembelik: Вынес предупреждение в AST-проход вместо регулярки: он видит алиасы импорта и `.required`, поэтому для `viewChild()` советует `?.`, а декораторную форму не трогает вовсе.

docs/guides/migration.en.md
artembelik: Взял формулировку из main без перечисления компонентов: список устаревал бы с каждым влившимся ревью, о чём ты и писал.

lskramarov (approved):

PR #1983, opened 2026-09-02
fix(link)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `link`, in the same shape as the 20.3.0 component reviews. It also closes the two `// @todo 20` markers the directive carried.

### Three accessor inputs, each doing more than storing a value

```ts
set disabled(value: boolean) { this.disabledSignal.set(value); }
get tabIndex(): number { return this.disabled ? -1 : this._tabIndex; }
set print(value: any) { this.printMode = value !== null; this._print = value; this.updatePrintUrl(); }
```

- **`tabIndex`** now reports what was bound. The host attribute still goes to `-1` while the link is disabled — that moved into a `hostTabIndex` computed — so nothing about focus behavior changed; only a programmatic read sees the difference.
- **`print`** accepts `string | null` instead of `any`. An unbound link no longer carries `kbq-link_print`: the old setter set `printMode = value !== null` and only ran when the input was bound, so the class depended on whether anyone bound `[print]` at all.
- **`disabled`** is a plain `booleanAttribute` input, and the `_disabled` mirror kept in sync through a `toObservable` subscription is gone.

### `disabledSignal` stays

`kbqTooltip` accepts a link through `forDisabledComponent`, typed `Record<'disabledSignal', WritableSignal<boolean>>`, and there is a docs example doing exactly that. So `disabledSignal` stays a public `WritableSignal<boolean>` — a `linkedSignal` over the input now, so binding `[disabled]` drives it and a direct write still wins. The host bindings read it rather than the input, which keeps that behavior identical.

### Closed internals

`icons`, `icon`, `hasIcon`, `printMode` and `printUrl` left the public surface. `icons` is a signal query driven by an effect instead of a `QueryList` + `changes` subscription, so an icon projected behind an `@if` now gets its spacing class — the same fix the badge got.

## Migration

`link-signals` runs from `ng update @koobiq/components@20`. It rewrites `disabled` reads to calls on receivers typed `KbqLink` and reports the rest. `tabIndex` is deliberately **not** rewritten: appending `()` would compile and hand back a different number for a disabled link.

There is no template-reference pass — `kbq-link` is an attribute on an anchor, so a `#ref="kbqLink"` read is not tied to an element name the schematic can match. The spec covers that decision.

## Testing

- `link.component.spec.ts`: 5 → 9 tests. New coverage for the disabled tab-order behavior, `tabIndex` reporting the bound value while disabled, a `forDisabledComponent`-style write to `disabledSignal`, and an unbound `print`.
- The existing print test needed its ordering corrected: it called `tick()` before the first `detectChanges()`, which only worked because the old constructor scheduled the microtask eagerly. In a real app the attribute has always landed on the change detection pass after the first, which is what the test asserts now.
- `link-signals/index.spec.ts`: 12 tests.
- Full `packages/components` (4997 tests) and `packages/schematics` (446 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


packages/schematics/src/migrations/link-signals/index.ts
```diff
@@ -0,0 +1,478 @@
...
+ * A `.ts` file is a link consumer if it names any of the exported symbols, imports the package, or renders
+ * the element in an inline template — a component that only imports `KbqLinkModule` names no type.
+ */
+function referencesLink(content: string): boolean {
+    return /\bKbqLink\w*\b/.test(content) || content.includes(LINK_PACKAGE) || content.includes(`<${LINK_ELEMENT}`);
+}
+
+export default function linkSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **`fix` is `undefined` under `ng update`, so the migration never writes anything.**

`@angular/cli` runs migrations as `this.executeSchematic(workflow, migration.collection.name, migration.name)` — no options argument, so `options = {}`. And `@angular-devkit/schematics/tools/schema-option-transform.js` applies schema defaults only `if (schematic.schema && schematic.schemaJson)`; the `link-signals` entry in `migrations.json` correctly has no `schema` key, so `schema.json`'s `"default": true` never reaches the rule.

Result: `ng update @koobiq/components@20` prints `[link-signals] would update src/app/x.ts (run with --fix to apply)` for every consumer file and rewrites nothing — while `migration.en.md:1042` promises "Every schematic named below runs automatically" and `:1070` says "the `disabled` reads are rewritten". The suggested `--fix` flag does not exist on `ng update`.

Twelve siblings already guard this, `app-switcher-signals/index.ts:413` with a comment spelling out the trap:

```suggestion
        const { project } = options;
        const fix = options.fix ?? true;
```

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
         '[class.kbq-link_pseudo]': 'pseudo()',
         '[class.kbq-link_multiline]': 'multiline()',
-        '[class.kbq-link_print]': 'printMode',
-        '[class.kbq-text-only]': '!hasIcon',
-        '[class.kbq-text-with-icon]': 'hasIcon',
-        '[class.kbq-disabled]': 'disabled',
-        '[attr.disabled]': 'disabled || null',
-        '[attr.tabindex]': 'tabIndex',
-        '[attr.print]': 'printUrl'
+        '[class.kbq-link_print]': 'printMode()',
```
lskramarov, line 43: **The `kbq-link_print` class this binding drives is dead CSS — nothing consumes it.**

`_link-theme.scss:193-194` nests the `@media print` block inside `.kbq-link { … }` but writes `.kbq-link_print:not(.kbq-disabled)` without a leading `&`, so it compiles to a **descendant** selector. Verified with the repo's own sass:

```css
@media print {
  .kbq-link .kbq-link_print:not(.kbq-disabled)::after { content: ":\a" attr(print); }
}
```

Both classes sit on the same host element, so the rule can never match. Every other same-element class in that file uses `&.` (`&.kbq-link_big`, `&.kbq-disabled`, `&.kbq-link_pseudo`, `&.cdk-keyboard-focused`) — this is the sole exception, and it is the only `@media print` rule in `packages/` that reads the class or the `print` attribute.

So for the shipped `link-print-example`, printing the page appends no URL: the directive sets `class="kbq-link_print"` and `print="…"` correctly, but no rule consumes them. That makes `printMode`, the `printUrl` signal, the new `should drop .kbq-link_print when print is unbound` spec, and the whole `kbq-link_print` migration note unobservable to any user. Pre-existing, but this PR reworked the entire print pipeline on top of it. Fix in `_link-theme.scss:194`: `&.kbq-link_print:not(.kbq-disabled)`.

packages/schematics/src/migrations/link-signals/index.ts
```diff
@@ -0,0 +1,478 @@
...
+        ]);
+    }
+}
+
+/** Collects template reference variable names bound to a `<kbq-link>` element. */
+class LinkRefCollector implements Visitor {
+    readonly refs = new Set<string>();
+
+    visitElement(element: any): void {
+        if (element.name === LINK_ELEMENT) {
```
lskramarov: **`LINK_ELEMENT` is matched as an element name, so the whole template pass is unreachable.**

`KbqLink` is the attribute selector `[kbq-link]` on an `<a>`, so `element.name` is always `a` and `collector.refs` is always empty. `data.ts:31` acknowledges this ("Kept for the shared machinery") — but the machinery is a private copy in this file, not shared, so there is nothing to keep it for.

Two concrete costs:

1. **Template reads are silently missed.** `<a #l="kbqLink" kbq-link [disabled]="d"></a> @if (l.disabled) { … }` — a pattern this repo uses twice (`packages/components/tooltip/tooltip.spec.ts:927`, `packages/docs-examples/components/tooltip/tooltip-disabled-for-component/tooltip-disabled-for-component-example.ts:28`). After the migration `l.disabled` is an `InputSignalWithTransform` function object: permanently truthy in a boolean position, no template type-check error, no rewrite — and no warning either, since `logWarnings`/`warnReceiverMembers` run only over `.ts` files.
2. **Wasted work.** `migrateTemplate`'s guard at :322 is `template.includes('kbq-link')`, which *is* true for every real `<a kbq-link>` consumer, so `parseTemplate` runs a full `@angular/compiler` parse on every such `.html` and every inline template and throws the result away (23 `.html` files, ~171 KB, in this repo alone).

Either match the attribute the way `button-truncation/index.ts:86` does (`hasAttr(node, BUTTON_ATTR)`) and rewrite the ref reads, or delete the pass along with `LINK_ELEMENT`, `LinkRefCollector`, `rewriteRefReads`, `migrateTemplate`, `migrateInlineTemplates`, `escapeRegExp`, the `htmlPaths` loop, and the `content.includes('<kbq-link')` disjunct at :404.

packages/schematics/src/migrations/link-signals/index.ts
```diff
@@ -0,0 +1,478 @@
...
+
+            edits.push({ start: node.getEnd(), end: rhs.getStart(sourceFile), text: '.set(' });
+            edits.push({ start: rhs.getEnd(), end: rhs.getEnd(), text: ')' });
+        }
+
+        return;
+    }
+
+    // Read (incl. optional chain `x?.compact`): append `()`.
+    edits.push({ start: node.getEnd(), end: node.getEnd(), text: '()' });
```
lskramarov: **Compound assignments and increments fall through to the read branch and produce a file that no longer parses.**

The write guard at :156 returns early only for `ts.SyntaxKind.EqualsToken`. `link.disabled ||= this.formDisabled` is a `BinaryExpression` with `BarBarEqualsToken`, so it reaches this line and the schematic writes:

```ts
link.disabled() ||= this.formDisabled;   // TS2364: left-hand side must be a variable or property access
```

Identical for `??=`, `&&=`, `+=`, and `link.disabled++` → `link.disabled()++` (a `PostfixUnaryExpression`, matched by none of the three guards). Once `fix` actually defaults to `true`, `ng update` writes this to disk — replacing a clear "cannot assign to a read-only input" error with a syntax-shaped one.

Widen the guard to `ts.isAssignmentExpression(parent)` (or an operator-token range check) and add prefix/postfix unary handling.

packages/schematics/src/migrations/link-signals/index.ts
```diff
@@ -0,0 +1,478 @@
...
+ * parameter-properties) and typed locals.
+ */
+function collectReceivers(sourceFile: ts.SourceFile, typeName: string): Receiver[] {
+    const receivers: Receiver[] = [];
+    const add = (text: string, scope: ts.Node) =>
+        receivers.push({ text, start: scope.getStart(sourceFile), end: scope.getEnd() });
+
+    const visit = (node: ts.Node): void => {
+        if (ts.isParameter(node) && ts.isIdentifier(node.name) && isTypeReference(node.type, typeName)) {
+            add(node.name.text, findAncestor(node, isFunctionLike) ?? sourceFile);
```
lskramarov: **Receiver scopes are flat ranges, so a shadowing inner binding of the same name is rewritten.**

A parameter's scope is the whole enclosing function; `this.<name>` (:117, :123) covers the whole class; a typed top-level local covers the whole `SourceFile`. Nothing re-resolves the identifier at the access site.

```ts
read(link: KbqLink) {
    this.buttons.forEach((link: KbqButton) => console.log(link.disabled)); // → link.disabled()
    return link.disabled;
}
```

The inner `link` is a `KbqButton`, whose `disabled` is a plain boolean getter, but it lies inside the outer parameter's range with the same source text, so both reads get `()`. Result: a compile error, or a runtime `TypeError: link.disabled is not a function` when the callback param is untyped. Same for a `for (const link of this.buttons)` loop inside such a method, and for a class declared inside a method of a class that has `link: KbqLink`.

packages/schematics/src/migrations/link-signals/index.ts
```diff
@@ -0,0 +1,478 @@
...
+    };
+
+    visit(sourceFile);
+
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
```
lskramarov: **Receivers are matched by raw source text, so `this.link!.disabled` is silently skipped — and `!` is the spelling strict mode forces.**

`@ViewChild(KbqLink) link!: KbqLink;` registers the receiver text `this.link`, but the read `this.link!.disabled` has `node.expression.getText()` === `'this.link!'`, so this returns `false`: no rewrite, and — because `collectReceiverWarnings` shares the same gate — no `tabIndex`/`printUrl`/`hasIcon` warning either.

The leftovers are not uniformly caught by `tsc`:

| leftover | result |
|---|---|
| `if (this.link!.disabled)` | TS2774 (caught) |
| `x: boolean = this.link!.disabled` | TS2322 (caught) |
| `!this.link!.disabled` | **no error — always `false`** |
| `this.link!.disabled \|\| fallback` | **no error — always truthy** |

A class mixing both spellings gets half-migrated. Note the spec's own fixture (`index.spec.ts:75`, `@ViewChild(KbqLink) link: KbqLink;`) would not compile in a consumer app with `strictPropertyInitialization`. Same hole for `(link as KbqLink).disabled` and `(link).disabled` — stripping `!`/parens/`as` from the receiver text before comparing fixes all three.

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
+    /** @docs-private */
+    protected readonly icon = contentChild(KbqIcon);
 
-    printMode: boolean;
+    /** @docs-private */
+    protected readonly hasIcon = computed(() => !!this.icon());
 
-    printUrl: string;
+    /** @docs-private */
+    protected readonly printMode = computed(() => this.print() != null);
```
lskramarov, line 117: **`!==` quietly became `!=`: a `[print]` binding whose expression is `undefined` now turns print mode off instead of on.**

Old: `this.printMode = value !== null`. For `<a kbq-link [href]="row.url" [print]="row.printUrl">` with `printUrl?: string` undefined until data loads, the setter ran with `undefined`, `undefined !== null` → `printMode = true` → the class applied and `printUrl` fell back to the href.

New: `undefined != null` → `false` → class absent, nothing printed, permanently if the value never becomes a string. The `print` attribute is still written (:123 keeps the href fallback), so the element looks right in DevTools while the print output is empty.

No test covers it, the schematic does not rewrite it, and neither migration guide mentions it — they document a *different*, non-existent change instead (see my comment on `migration.en.md:1068`).

Related: `printMode` is nullish (`!= null`) while the URL fallback at :123 is falsy (`||`), so the two disagree on `''` — `[print]="''"` turns print mode **on** and then prints the anchor's own `href`. That is the state `KbqLinkPrintTestApp` starts in, so the spec now bakes it in.

packages/schematics/src/migrations/link-signals/data.ts
```diff
@@ -0,0 +1,89 @@
...
+ * - `link.tabIndex` → a signal whose value changed: the getter folded in the disabled state (warn)
+ * - `link.print` → was write-only (a setter with no getter), so it has no read to rewrite (warn on writes)
+ * - `icons` / `icon` / `hasIcon` / `printMode` / `printUrl` → `protected` / `private` (warn)
+ *
+ * `disabledSignal` stays a public `WritableSignal<boolean>`: `kbqTooltip` accepts a link through
+ * `forDisabledComponent` and reads it, and that contract is typed on the tooltip side.
+ */
+
+/** Members of `KbqLink` whose value is unchanged; a read must become a call. Auto-fixed. */
+export const SIGNAL_MEMBERS: readonly string[] = ['disabled'];
```
lskramarov, line 17: **`disabled` is auto-fixed as "value unchanged", but its value did change.**

The old accessor returned the *effective* state:

```ts
get disabled(): boolean { return this._disabled; }
toObservable(this.disabledSignal).subscribe((value) => (this._disabled = value));
```

so `link.disabled` reflected a direct `disabledSignal.set(true)` — exactly the `forDisabledComponent` contract this PR keeps `disabledSignal` public for. The new input reports only what was bound, so after the rewrite:

```ts
link.disabledSignal.set(true);
if (link.disabled()) { … }   // false, while the host renders .kbq-disabled and tabindex="-1"
```

`tabIndex` is warned about for precisely this reason (`data.ts:42`: "a mechanical `()` append would silently change it") and then `disabled` is auto-fixed under the same condition. The caveat exists only in the once-per-project `SUMMARY` at :86 — printed *after* the edit has been applied, with no pointer to the file it was applied to. Either move `disabled` into `VALUE_CHANGED_MEMBERS`, or emit the caveat per-file at the rewritten call site.

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
-        '[class.kbq-text-with-icon]': 'hasIcon',
-        '[class.kbq-disabled]': 'disabled',
-        '[attr.disabled]': 'disabled || null',
-        '[attr.tabindex]': 'tabIndex',
-        '[attr.print]': 'printUrl'
+        '[class.kbq-link_print]': 'printMode()',
+        '[class.kbq-text-only]': '!hasIcon()',
+        '[class.kbq-text-with-icon]': 'hasIcon()',
+        '[class.kbq-disabled]': 'disabledSignal()',
+        '[attr.disabled]': 'disabledSignal() || null',
```
lskramarov: **`disabled` on an anchor is inert for assistive tech, and there is no `aria-disabled`.**

`[kbq-link]` is applied to `<a>` and `<span>` hosts, where `disabled` is not a valid HTML attribute. `<a href="…" kbq-link [disabled]="true">` renders `disabled="true" tabindex="-1" class="… kbq-disabled"`, and `tabindex="-1"` only removes it from the tab order — a screen reader in browse mode (NVDA/JAWS virtual cursor, VoiceOver rotor) still reaches the anchor through the links list and announces a plain, actionable link. `.kbq-disabled { pointer-events: none }` blocks the mouse but not <kbd>Enter</kbd> on a programmatically focused anchor.

The wave-1 button review already settled this — `packages/components/button/button.component.ts:460-468` splits it into `nativeDisabledAttribute` / `ariaDisabledAttribute` (bound at :326-327), and `packages/components/breadcrumbs/breadcrumbs.spec.ts:107` states the rule verbatim: "`disabled` is not a valid attribute on an anchor, so the state is exposed via ARIA."

This PR rewrote the binding and added `expect(link.getAttribute('disabled')).toBe('true')` at `link.component.spec.ts:112`, codifying the gap. AGENTS.md requires WCAG AA minimums and correct ARIA attributes.

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
+        const icons = this.icons();
 
-    private updateClassModifierForIcons = () => {
-        this.icons.forEach(({ nativeElement }) => {
+        icons.forEach(({ nativeElement }) => {
             this.renderer.removeClass(nativeElement, leftIconClassName);
             this.renderer.removeClass(nativeElement, rightIconClassName);
         });
 
         const filteredNodesWithoutComments = getNodesWithoutComments(this.nativeElement.childNodes as NodeList);
```
lskramarov: **Icon edge detection breaks under `preserveWhitespaces`; button and button-toggle already carry the fix.**

`getNodesWithoutComments` filters comments only. With `"preserveWhitespaces": true`, `<a kbq-link>\n  <i kbq-icon></i>\n  Text\n</a>` yields childNodes `[text(" "), i, text(" Text ")]` — `iconIndex` is 1, neither `0` nor `length - 1`, so the icon gets neither `kbq-icon_left` nor `kbq-icon_right` and the gap disappears. Buttons in the same app render correctly, so it reads as a link-specific styling bug.

`packages/components/button/button.component.ts:281` and `packages/components/button-toggle/button-toggle.component.ts:63` both carry a hardened `getContentNodes` with byte-identical JSDoc:

> Child nodes that take part in the icon detection: comments and whitespace-only text nodes are ignored so that detection does not depend on `preserveWhitespaces`.

This PR rewrote the function (`ngAfterContentInit` → `effect`) and reproduced the older walk. Since that helper is now duplicated in two components and needed in a third, the right move is to export it once next to `getNodesWithoutComments` in `packages/components/core/utils/utils.ts:28` and call it from all three.

lskramarov (commented):
## Review: `link` signal refactor + `link-signals` schematic

15 findings, most severe first. Full-effort pass over the component, the schematic, both migration guides and the API guard.

**Two blockers:**

1. **`index.ts:409` — the migration never writes anything.** `ng update` invokes migrations with no options and `migrations.json` declares no schema, so `fix` is `undefined`. Every consumer gets "would update … (run with `--fix`)" and no rewrite — and `--fix` is not a flag `ng update` accepts. Twelve siblings already guard this with `options.fix ?? true`.
2. **`_link-theme.scss:194` — the `kbq-link_print` rule is dead CSS.** The `@media print` block is nested in `.kbq-link` without a leading `&`, so it compiles to `.kbq-link .kbq-link_print` — a descendant selector for two classes on the same element. The whole print pipeline this PR reworked renders nothing. (Pre-existing; commented on `link.component.ts:42` since the SCSS is outside the diff.)

**Schematic correctness:** the template pass can never fire (`kbq-link` matched as an element name, not an attribute), compound assignments are rewritten into invalid syntax, shadowed receivers are rewritten, and `this.link!.disabled` — the spelling `strictPropertyInitialization` forces — is silently skipped by both the rewrite and the warnings.

**Component:** `printMode` loosened `!==` to `!=`, the disabled anchor has no `aria-disabled`, icon edge detection regresses under `preserveWhitespaces`, and `printUrl` goes stale on `[href]` changes.

**Docs:** the bolded "unbound link no longer carries `kbq-link_print`" change did not happen (old code behaved identically), a table cell is truncated mid-sentence, and the rewritten wave intro contradicts its own subsection list.

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
+        return this.nativeElement;
     }
 
-    ngAfterContentInit() {
-        this.updateClassModifierForIcons();
-
-        this.icons.changes.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(this.updateClassModifierForIcons);
-    }
+    private updateClassModifierForIcons(): void {
+        const icons = this.icons();
```
lskramarov, line 183: **The effect branches on DOM state it does not track.**

The only reactive dependency is `this.icons()`, but the gate at :160 (`filteredNodesWithoutComments.length > 1`) and the `findIndex` positioning both read `nativeElement.childNodes` imperatively.

`<a kbq-link><i kbq-icon="kbq-plus_16"></i>@if (label()) { {{ label() }} }</a>`: with `label()` false the non-comment children are `[i]` — length 1, gate false, no class. Flipping `label()` to true makes them `[i, text]`, which should give the icon `kbq-icon_left` — but `icons()` did not change, so the effect never re-runs and the icon stays unspaced. Symmetrically, a stale `kbq-icon_right` survives when the sibling disappears.

The comment added at :127-128 says the assignment now reacts to the signal; it reacts to the *icon query* only, not to the sibling nodes the branch actually reads. Inherited from the old `icons.changes` subscription, but the function was rewritten here and neither spacing class has a test — the spec only covers `kbq-text-with-icon`, which comes from the `hasIcon` computed.

packages/components/link/link.component.ts
```diff
@@ -39,133 +39,126 @@ export const baseURLRegex = /^http(s)?:\/\//;
...
+    /** @docs-private */
+    protected readonly hostTabIndex = computed(() => (this.disabledSignal() ? -1 : this.tabIndex()));
 
-    readonly icon = contentChild(KbqIcon);
+    /** @docs-private */
+    protected readonly printUrl = signal<string | undefined>(undefined);
 
     constructor() {
-        this.updatePrintUrl();
+        effect(() => {
```
lskramarov: **`printUrl` derives from `href` but only tracks `print()`, and moving it out of the constructor added a second failure mode.**

1. **Stale on `[href]` change.** `<a kbq-link [href]="url()" [print]>` with `url()` going `…/1` → `…/2`: `print()` never changes, the effect never re-runs, and the printed page shows the old URL next to the new link text — silently, while the documented default is "the `href` without its protocol".
2. **Never computed in a detached view.** This `effect()` resolves a `ViewContext`, so it is a view effect that runs only from `runEffectsInView` inside `refreshView`. Under `ChangeDetectorRef.detach()`, or in a view not attached to `ApplicationRef`, `printUrl` is now never set — the old constructor-scheduled microtask always ran.

The repo already uses the sanctioned primitive for a deferred DOM read: `afterNextRender` at ~65 call sites in `packages/components`, e.g. `core/form-field/field-sizing-content.ts:43` (same `kbqInjectNativeElement` + deferred read). It is tied to the injection context, runs in the render phase rather than an untracked microtask, and does not run during SSR — and it would drop the extra `fixture.detectChanges()` the spec had to add at :51 and :63.

packages/schematics/src/migrations/link-signals/data.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * `tabIndex` became a read-only `InputSignal` AND changed its value: the getter returned `-1` for a
+ * disabled link, it now reports what was bound. The host attribute still goes to `-1` while the link is
+ * disabled, so nothing about focus behavior changed — only a programmatic read sees the difference, and a
+ * mechanical `()` append would silently change it.
+ */
+export const VALUE_CHANGED_MEMBERS: readonly string[] = ['tabIndex'];
+
+/** Members that moved out of the public surface and can no longer be read from outside the directive. */
+export const PROTECTED_MEMBERS: readonly string[] = ['icons', 'icon', 'hasIcon', 'printMode', 'printUrl'];
```
lskramarov: **`PROTECTED_MEMBERS` misses several members that left the public surface, and the message's wording is wrong for `icons`.**

Diffing `git show 4bb1c9927:tools/public_api_guard/components/link.api.md` against the new file, these also left the surface with no entry here, no `warnPatterns` rule, and no row in either migration table:

- `protected readonly nativeElement: HTMLElement` → now `private`
- `protected readonly destroyRef: DestroyRef` → deleted
- `ngAfterContentInit(): void` → deleted, along with `implements AfterContentInit`
- `set tabIndex(value: number)` → gone; `link.tabIndex = 3` no longer compiles, while the `tabIndex` warning text only discusses *reading*

`class MyLink extends KbqLink { ngAfterContentInit() { super.ngAfterContentInit(); this.renderer.setAttribute(this.nativeElement, …); } }` breaks on three members with zero guidance.

Separately: `link.component.ts:58` declares `icons` **`private`**, not `protected`, so the message emitted at `index.ts:258` ("These KbqLink members are now `protected`") is wrong for it. Both doc tables get this right ("Now `protected`/`private`"); only the runtime message a consumer actually sees does not.

docs/guides/migration.en.md
```diff
@@ -1049,6 +1049,26 @@ Most of them report rather than rewrite: what replaces a removed member or a sig
...
+| ------------------------------- | -------------------------------------------------------------------------- |
+| `.disabled`                     | Read as `disabled()` — rewritten for you                                   |
+| `.tabIndex`                     | `tabIndex()`, and expect what was bound — not `-1` for a disabled link     |
+| `.print = …`                    | Bind `[print]`; it was a setter with no getter, so there is no read to fix |
+| `.icons` / `.icon` / `.hasIcon` | Now `protected`/`private`; the icon spacing classes are the contract       |
+| `.printMode` / `.printUrl`      | Now `protected`; the `kbq-link_print` class and `print` attribute are      |
+
+The host attribute still goes to `-1` while the link is disabled, so nothing about focus behavior changed — only a programmatic read of `tabIndex` sees the difference.
+
+**An unbound link no longer carries the `kbq-link_print` class.** The old setter set `printMode = value !== null` and only ran when the input was bound, so the class depended on whether anyone bound `[print]` at all. It is driven by the input now and is absent until you bind it. `print` accepts `string | null` instead of `any`.
```
lskramarov: **This breaking change never happened.**

In the old component `printMode: boolean;` had no initializer and was written only by `set print`. For `<a kbq-link>` with no binding the setter never ran, so `printMode === undefined` → `[class.kbq-link_print]` falsy → **class absent**. New: `print()` is `undefined` → `undefined != null` is `false` → **class absent**. Identical.

Consequences:

- Consumers are told to audit every unbound link for a change that did not occur.
- The new spec `should drop .kbq-link_print when print is unbound` (`link.component.spec.ts:68`) passes verbatim against the pre-PR component, so it pins nothing.
- The shape that *did* change — `[print]` bound to an `undefined`-valued expression, which used to turn print mode **on** — is documented nowhere.

The same claim is repeated in `docs/guides/migration.ru.md:1072`, `packages/schematics/src/migrations/link-signals/README.md:48`, `data.ts:82` (the `SUMMARY` printed to every consumer), and the descriptions in both `collection.json` and `migrations.json`.

docs/guides/migration.en.md
```diff
@@ -1049,6 +1049,26 @@ Most of them report rather than rewrite: what replaces a removed member or a sig
...
+
+`disabledSignal` stays a public `WritableSignal<boolean>` — `kbqTooltip` accepts a link through `forDisabledComponent` and reads it. It is a `linkedSignal` over the `disabled` input now, so binding still drives it and a direct write still wins.
+
+| Pattern                         | Manual migration                                                           |
+| ------------------------------- | -------------------------------------------------------------------------- |
+| `.disabled`                     | Read as `disabled()` — rewritten for you                                   |
+| `.tabIndex`                     | `tabIndex()`, and expect what was bound — not `-1` for a disabled link     |
+| `.print = …`                    | Bind `[print]`; it was a setter with no getter, so there is no read to fix |
+| `.icons` / `.icon` / `.hasIcon` | Now `protected`/`private`; the icon spacing classes are the contract       |
+| `.printMode` / `.printUrl`      | Now `protected`; the `kbq-link_print` class and `print` attribute are      |
```
lskramarov: **Truncated cell** — the sentence ends on a dangling verb: "the `kbq-link_print` class and `print` attribute **are**".

The RU counterpart at `migration.ru.md:1068` is complete ("контракт — класс `kbq-link_print` и атрибут `print`"), so the predicate was lost only on the EN side. `packages/schematics/src/migrations/link-signals/README.md:40` carries the identical truncated cell.

The suggestion spans the whole table because prettier realigns every row once the last cell grows — this is `prettier --parser markdown --print-width 120` output, so it is check-clean:

```suggestion
| Pattern                         | Manual migration                                                                   |
| ------------------------------- | ---------------------------------------------------------------------------------- |
| `.disabled`                     | Read as `disabled()` — rewritten for you                                           |
| `.tabIndex`                     | `tabIndex()`, and expect what was bound — not `-1` for a disabled link             |
| `.print = …`                    | Bind `[print]`; it was a setter with no getter, so there is no read to fix         |
| `.icons` / `.icon` / `.hasIcon` | Now `protected`/`private`; the icon spacing classes are the contract               |
| `.printMode` / `.printUrl`      | Now `protected`; the `kbq-link_print` class and `print` attribute are the contract |
```

---

Separately, one line up in the same section — **`migration.en.md:1038` contradicts its own subsection list.** It now reads:

> The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; **the second is the one each subsection below belongs to**.

But the subsections under `### 18` are Link (1052), **Search expandable** (1072), **Split button** (1094) and **Title** (1116) — three of the four are named in the wave-1 list in the same sentence. Only Link belongs to the second wave. `migration.ru.md:1042` carries the identical error, so both files are in sync and both wrong; the sentence this replaced was consistent.

packages/components/link/link.component.ts
artembelik: поправил в `_link-theme.scss`, скомпилированный селектор теперь `.kbq-link.kbq-link_print`

packages/schematics/src/migrations/link-signals/index.ts
artembelik: операторы присваивания вынес в набор, а из префиксных записью считаю только `++`/`--`: `!link.disabled` тоже префиксный, но это чтение

packages/components/link/link.component.ts
artembelik: сделал как в button, а набор тегов вынес в `core/utils`, чтобы он не дублировался

packages/components/link/link.component.ts
artembelik: эффект оставил на `icons()`, а соседние узлы отслеживаю CDK `ContentObserver` на хосте

packages/components/link/link.component.ts
artembelik: `afterNextRender` закрывает только отсоединённое представление, но не устаревание при смене `[href]`, поэтому читаю `href` прямо в host-биндинге - он вычисляется после биндингов шаблона и на каждой проверке

packages/schematics/src/migrations/link-signals/data.ts
artembelik: взял второй вариант, оговорка печатается в отчёте по каждому переписанному файлу

docs/guides/migration.en.md
artembelik: `suggestion` применил, добавил ещё строку про `ngAfterContentInit`; вступление про две волны переписано в main и пришло со слиянием

packages/components/link/link.component.ts
artembelik: гайд и тест на `[print]="undefined"` ушли ещё в `65d38acf9`. пустая строка и бесзначный `print` означают «печатать href по умолчанию», дописал это в JSDoc

lskramarov:
что то копилот перевозбудился, давно столько не находил..

lskramarov (approved):

packages/components/core/utils/utils.ts
```diff
@@ -31,6 +31,30 @@ export const getNodesWithoutComments = (nodes: NodeList): Node[] => {
...
 
+/**
+ * Child nodes that take part in the icon detection: comments and whitespace-only text nodes are
+ * ignored so that detection does not depend on `preserveWhitespaces`.
+ */
+export const getContentNodes = (element: Node): Node[] => {
+    const TEXT_NODE = 3;
+
+    return getNodesWithoutComments(element.childNodes).filter(
+        (node) => node.nodeType !== TEXT_NODE || !!node.textContent?.trim()
```
NikGurev: Посмотри, может лучше будет взять из NODE, чем определять тип ноды локально

```suggestion
        (node) => node.nodeType !== Node.TEXT_NODE || !!node.textContent?.trim()
```

packages/components/core/utils/utils.ts
```diff
@@ -31,6 +31,30 @@ export const getNodesWithoutComments = (nodes: NodeList): Node[] => {
...
+/** Host tags that support the native `disabled` attribute. */
+const nativelyDisableableTags = new Set([
+    'button',
+    'input',
+    'select',
+    'textarea'
+]);
+
+/** Whether the element supports the native `disabled` attribute; on any other host it is invalid HTML. */
+export const supportsNativeDisabled = (element: Element): boolean =>
```
NikGurev, line 51: Посмотри, не будет ли конфликтовать с нашими директивами, которые добавляют disabled атрибут даже если тег его не поддерживает ? Возможно, нужно будет их отдельно поправить, но поведение точно должно работать как прежде

Например в packages/components/dropdown/e2e.ts

```
<div kbq-dropdown-item disabled data-testid="e2eDropdownItemActionDisabled">
           Disabled row and action
          <a disabled kbq-icon-button kbqDropdownItemAction aria-label="Settings" href="/settings">
              <i kbq-icon="kbq-gear_16" [color]="'contrast-fade'"></i>
          </a>
</div>
```

NikGurev (approved):

packages/components/core/utils/utils.ts
artembelik: применил, и заодно перевёл соседний `getNodesWithoutComments`: иначе один хелпер читал бы `Node.TEXT_NODE`, а соседний - локальную `8`. локальные константы стояли на случай отсутствия `Node` на сервере, но `platform-server` кладёт типы domino в `globalThis` через `setDomTypes()`, так что опасение не подтвердилось

packages/components/core/utils/utils.ts
artembelik: проверил, не конфликтует

PR #1985, opened 2026-09-02
fix(autocomplete)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `autocomplete`, in the same shape as the 20.3.0 component reviews. Four accessor inputs and one write-target input survived the automated signal migration, across the panel and its trigger.

### `classList` was an `@Input('class')` that accumulated

```ts
set classList(value: string) {
    if (value && value.length) {
        const classList = { ...this._classList };
        value.split(' ').forEach((c) => (classList[c.trim()] = true));
        this._classList = classList;
        this.elementRef.nativeElement.className = '';
    }
}
```

It mixed three concerns: the consumer's classes, the internal `kbq-autocomplete_visible` / `_hidden` pair, and `cdk-keyboard-focused`. Those are three signals and one `computed` now, behind a `class` signal input. `class="…"` on `<kbq-autocomplete>` keeps working exactly as before, including the host-clearing side effect.

### `isOpen` disagreed with itself

```ts
get isOpen() { return this._isOpen && this.showPanel; }
set isOpen(value) { this._isOpen = value; }
```

Writing `true` and reading it back returned `false` whenever the panel had no options. It is a `computed` now — `attached() && showPanel()` — with `attached` as the writable half the trigger owns, which is what the setter always meant.

### The rest

- `displayWith`, `autoActiveFirstOption`, `openOnFocus` and the trigger's `kbqAutocompleteDisabled` are signal inputs. The last three gained `booleanAttribute` in place of `coerceBooleanProperty` or nothing.
- The panel id comes from the CDK `_IdGenerator` instead of a module-level counter. It is the value of the trigger's `aria-owns`, so the shape changes from `kbq-autocomplete-1` to `kbq-autocomplete-a1`.
- `setVisibility`, `emitSelectEvent`, `onKeydown`, `setScrollTop`, `getScrollTop`, `keyManager` and `showPanel` are marked `@docs-private`: they are the trigger's interface to the panel, not the consumer's.

### Deliberately left alone

`options` stays a `QueryList` content query. `ActiveDescendantKeyManager` takes one directly, and `subscribeToClosingActions` rebuilds the panel-closing stream off `options.changes` — a signal query would mean a `toObservable` per panel open, with the effect outliving the subscription. Not worth the risk in this PR.

## Migration

`autocomplete-signals` runs from `ng update @koobiq/components@20`. It rewrites reads on receivers typed `KbqAutocomplete` or `KbqAutocompleteTrigger` and reports the rest.

It also reports a gotcha worth calling out: **binding `[autoActiveFirstOption]` overrides `KBQ_AUTOCOMPLETE_DEFAULT_OPTIONS` even when the bound value is `undefined`.** The token default only applies to an input nobody bound. This one is easy to hit when converting a field write into a binding — it caught me while updating the spec.

## Testing

- `autocomplete.spec.ts`: 122 → 123 tests, all existing ones kept.
  - Six tests mutated inputs on the instance. Each now binds from its host — and deliberately **not** on the shared `SimpleAutocomplete`, because a bound `[autoActiveFirstOption]` there silently defeats the test that checks `KBQ_AUTOCOMPLETE_DEFAULT_OPTIONS`. Preselection has its own host now, plus a new test that the input really does override the token.
  - Fixed a pre-existing type error the migration surfaced: the `openOnFocus` describe declared `ComponentFixture<AutocompleteWithDisabledItems>` while creating `AutocompleteWithOpenOnFocus`.
- `autocomplete-signals/index.spec.ts`: 16 tests, covering both receiver types.
- Full `packages/components` (4994 tests) and `packages/schematics` (450 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


packages/components/autocomplete/autocomplete.component.ts
```diff
@@ -154,74 +160,58 @@ export class KbqAutocomplete implements AfterContentInit {
...
+
+    /** @docs-private */
+    protected readonly classList = computed<Record<string, boolean>>(() => ({
+        ...Object.fromEntries(
+            this.hostClass()
+                .split(' ')
+                .map((className) => className.trim())
+                .filter(Boolean)
+                .map((className) => [className, true])
+        ),
```
lskramarov, line 220: **`classList` throws when `[class]` is bound to anything but a string.**

The accessor this replaces was guarded — `set classList(value: string) { if (value && value.length) { … value.split(' ') … } }` — so `undefined`, `null` and object/array class maps were silent no-ops. The computed calls `.split(' ')` unconditionally.

Angular forwards the binding value to a styling-shadowing input **raw**. In `checkStylingMap` the `concatStringsWithSpace(staticPrefix, value ? value : '')` normalisation only runs `if (staticPrefix !== null)`, i.e. only when the element also carries a *static* `class="…"` attribute (`host: { class: 'kbq-autocomplete' }` does not count — `classesWithoutHost` excludes host contributions). Without one, `setDirectiveInputsWhichShadowsStyling` passes `undefined` / `null` / `{…}` / `[…]` straight into the input, and signal inputs do not fall back to their declared default for an explicit `undefined`.

```html
<!-- no static class= on the element -->
<kbq-autocomplete [class]="panelClass">
```
`panelClass === undefined` → `TypeError: Cannot read properties of undefined (reading 'split')`; `[class]="{ 'my-panel': cond }"` → `hostClass(...).split is not a function`. The error is thrown from `[class]="classList()"` in `autocomplete.html:2`, so the whole panel fails to render on open, and `computed` caches the throw until a dependency changes. Note AGENTS.md tells authors to prefer `[class]="{…}"` over `ngClass`.

Cheapest fix is to coerce at the input rather than in the computed:

```ts
readonly hostClass = input('', {
    alias: 'class',
    transform: (value: unknown) => (typeof value === 'string' ? value : '')
});
```
(or widen it to `string | string[] | Record<string, boolean>` the way Angular's own `[class]` accepts).

tools/public_api_guard/components/autocomplete.api.md
```diff
@@ -59,29 +59,22 @@ export const KBQ_AUTOCOMPLETE_VALUE_ACCESSOR: Provider;
 // @public (undocumented)
 export class KbqAutocomplete implements AfterContentInit {
     constructor();
-    get autoActiveFirstOption(): boolean;
-    set autoActiveFirstOption(value: boolean);
-    get classList(): string;
-    set classList(value: string);
+    readonly attached: _angular_core.WritableSignal<boolean>;
+    readonly autoActiveFirstOption: _angular_core.InputSignalWithTransform<boolean, unknown>;
```
lskramarov, line 63: **The `api` CI job fails on this PR:** `check-public-api-any` reports `autocomplete: 9 → 12`.

The three `booleanAttribute` transforms each add an `unknown` to the published surface — this line plus `openOnFocus` (:67) and `KbqAutocompleteTrigger.autocompleteDisabled` (:106) — and `tools/check-public-api-any/baseline.json` still records `"autocomplete": 9`.

`InputSignalWithTransform<boolean, unknown>` is the repo-wide shape (103 occurrences across the guards), so narrowing is not the intent here — the baseline just needs re-recording:

```bash
yarn run approve-public-api-any
```

(The `unit_tests` failure is unrelated: `tree-selection.component.spec.ts:1666` TS2445, already fixed on `main` by 0f6a3d9e0. The branch needs a rebase.)

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
```
lskramarov: **`--fix` writes an unparseable template.** `rewriteRefReads` has no write detection — unlike `classifyAccess`, which deliberately bails out on `x.member = RHS` (:151). The lookahead only skips an existing call.

```html
<!-- before -->
<kbq-autocomplete #auto /><button (click)="auto.isOpen = false">x</button>
<!-- after --fix -->
<kbq-autocomplete #auto /><button (click)="auto.isOpen() = false">x</button>
```

`auto.isOpen() = false` is rejected by Angular's action parser, so the component stops compiling. Same for `(click)="auto.showPanel = true"` — and both were legal markup before this PR (`isOpen` had a public setter, `showPanel` was a public writable `boolean`).

The sibling this was forked from states the premise out loud — `button-toggle-signals-and-aria/index.ts:440`: *"A template can only read these, so an assignment never needs excluding."* True there; false for the autocomplete, and copied without re-checking.

Two aggravating factors: `logWarnings`/`warnReceiverMembers` are only called in the `tsPaths` loop (:431-432) — the `htmlPaths` loop (:441-452) calls neither, so `.html` files get no warning at all; and the `isOpen =` warn anchor `\bKbqAutocomplete\b` does not match `KbqAutocompleteModule`, so a component that only imports the module gets the corrupt rewrite silently.

lskramarov (commented):
## Code review — `max` effort

Ten finder angles plus a gap sweep over the component, the trigger, the spec and the new `autocomplete-signals` schematic. 15 findings, most severe first. Verified against the installed `@angular/core` 20.3.29 / `@angular/cdk` sources and the PR’s own CI logs.

Headline: the `api` check is red because `tools/check-public-api-any/baseline.json` was not re-approved (`autocomplete: 9 → 12`), and `KbqAutocomplete.classList` now throws on a non-string `[class]` binding that the old accessor silently ignored. The rest are in the schematic — it rewrites consumers’ source, so a wrong rewrite is worth more than a missed one.

_The `unit_tests` failure is not this PR: `tree-selection.component.spec.ts:1666` is already fixed on `main` by 0f6a3d9e0. Rebase._

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
+}
+
+/** Classifies a matched property access and appends the resulting edit(s). */
+function classifyAccess(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, edits: Edit[]): void {
+    const parent = node.parent;
+
+    // Already migrated: a call, or a `.set(...)` write — leave alone (idempotent).
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
```
lskramarov: **`displayWith` is function-valued, so "a call means already migrated" is exactly inverted for it.**

`displayWith` is the only `SIGNAL_MEMBER` whose value is itself a function, so `x.displayWith(value)` is a legitimate *pre*-migration invocation — the shape the library itself used (`autocomplete-trigger.directive.ts:569` at the base commit). This guard skips it, and `rewriteRefReads`' `(?!\s*\()` skips it in templates too.

```ts
// before
return a.displayWith ? a.displayWith(v) : v;
// after --fix  (half-migrated)
return a.displayWith() ? a.displayWith(v) : v;
```

In TypeScript that degrades to a loud arity error. In a template it fails **silently**: `createInputSignal`'s getter ignores extra arguments, so

```html
{{ auto.displayWith(state) }}   <!-- unchanged by --fix -->
```
now interpolates the function object instead of the label. The correct rewrite is `auto.displayWith()(state)`.

Nothing covers it — `displayWith` is not in `PROTECTED_MEMBERS` and no `warnPattern` matches a call.

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+export const SUMMARY = [
+    '  `autoActiveFirstOption`, `openOnFocus` and `kbqAutocompleteDisabled` are `booleanAttribute` inputs. ' +
+        'A valueless attribute means true now, where it used to pass the empty string — `coerceBooleanProperty` ' +
+        'already treated that as false, so only markup that relied on the empty string reading as false changes.',
+    '  Binding `[autoActiveFirstOption]` now overrides KBQ_AUTOCOMPLETE_DEFAULT_OPTIONS even when the bound ' +
+        'value is `undefined`. The default only applies to an input nobody bound, so leave it unbound to let ' +
+        'the token decide.',
+    '  Generated panel ids come from the CDK `_IdGenerator`, so their shape changed from `kbq-autocomplete-1` ' +
+        "to `kbq-autocomplete-a1`. It is the value of the trigger's `aria-owns`, so anything asserting on that " +
+        'shape needs updating.'
```
lskramarov: **Both halves of this note are false**, and it is repeated verbatim in `README.md:50-51`, `docs/guides/migration.en.md:1072`, `docs/guides/migration.ru.md:1076` and the commit message.

1. **The shape did not change.** `_IdGenerator.getId(prefix)` appends the app id *only* when `APP_ID !== 'ng'` (`if (this._appId !== 'ng') { prefix += this._appId; }`, then `prefix + counters[prefix]++`), and `DEFAULT_APP_ID = 'ng'`. For a default app that is `kbq-autocomplete-0`, `kbq-autocomplete-1`, … — byte-identical to the deleted `uniqueAutocompleteIdCounter`. The `a` in `kbq-autocomplete-a1` comes from `@angular/platform-browser/testing`, which provides `{ provide: APP_ID, useValue: 'a' }` — i.e. the shape only differs *in tests*, or in an app that sets a custom `APP_ID`.

2. **There is no `aria-owns`.** Grepping `packages/` and `apps/` finds it only in the stale JSDoc at `autocomplete.component.ts:102` and here. The trigger's `host` block (`autocomplete-trigger.directive.ts:119-129`) sets no `aria-owns`, `aria-controls` or `aria-activedescendant`; `id` reaches the DOM only as `[id]` on the panel div (`autocomplete.html:2`).

As written, a consumer audits ARIA assertions that do not exist, for an id shape their app never produced.

```suggestion
    '  Generated panel ids come from the CDK `_IdGenerator`. With the default `APP_ID` the shape is unchanged ' +
        '(`kbq-autocomplete-0`, `kbq-autocomplete-1`, …); an app or test that sets a custom `APP_ID` now gets it ' +
        "embedded, e.g. `kbq-autocomplete-a1` under TestBed. The id is the panel element's `id`, so selectors and " +
        'snapshots targeting it may need updating.'
```

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+            'are now signals — reading one is a double call, e.g. `this.autocomplete().isOpen()`. Verify ' +
+            'query reads manually.'
+    }
+];
+
+/** Printed once per project, after the per-file reports. */
+export const SUMMARY = [
+    '  `autoActiveFirstOption`, `openOnFocus` and `kbqAutocompleteDisabled` are `booleanAttribute` inputs. ' +
+        'A valueless attribute means true now, where it used to pass the empty string — `coerceBooleanProperty` ' +
+        'already treated that as false, so only markup that relied on the empty string reading as false changes.',
```
lskramarov: **This is inverted, and it hides the one input that actually changed.**

`coerceBooleanProperty(value)` is `value != null && String(value) !== 'false'`, so `coerceBooleanProperty('')` is **`true`** — same as `booleanAttribute('')`. For `autoActiveFirstOption` and `kbqAutocompleteDisabled`, which both already used `coerceBooleanProperty`, the transform is a no-op and *nothing* changed.

The input that did change is `openOnFocus`: its old setter did no coercion at all, and `handleFocus()` truth-tested the raw value. Both directions flip and neither is documented:

| markup | before | after |
| --- | --- | --- |
| `<kbq-autocomplete openOnFocus>` (→ `''`) | does not open on focus | opens |
| `[openOnFocus]="0"` | does not open | opens |
| `[openOnFocus]="'false'"` | opens | does not open |

`README.md:44-46` states the premise correctly ("used `coerceBooleanProperty` **or nothing at all**"), so the two disagree — and this is the text users actually see in the `ng update` output. `docs/guides/migration.{en,ru}.md` carry the same claim.

```suggestion
    '  `autoActiveFirstOption`, `openOnFocus` and `kbqAutocompleteDisabled` are `booleanAttribute` inputs. ' +
        'For `autoActiveFirstOption` and `kbqAutocompleteDisabled` that matches the `coerceBooleanProperty` they ' +
        'already used, so nothing changes. `openOnFocus` had no coercion at all: a valueless attribute or `0` ' +
        "used to read as false and now means true, and `'false'` used to read as true and now means false.",
```

packages/components/autocomplete/autocomplete.component.ts
```diff
@@ -154,74 +160,58 @@ export class KbqAutocomplete implements AfterContentInit {
...
-        return this._isOpen && this.showPanel;
-    }
-
-    set isOpen(value: boolean) {
-        this._isOpen = value;
-    }
-
-    private _isOpen: boolean = false;
+    /** Whether the panel opens as soon as the trigger receives focus. */
+    readonly openOnFocus = input(true, { transform: booleanAttribute });
```
lskramarov, line 196: **This is a behavior change, not a no-op like the other two.**

The accessor it replaces did no coercion — `set openOnFocus(value: boolean) { this._openOnFocus = value; }` — and the only consumer is a bare truthiness test (`autocomplete-trigger.directive.ts:449`). `autoActiveFirstOption` and `kbqAutocompleteDisabled` used `coerceBooleanProperty`, which is semantically identical to `booleanAttribute`, so those two really are no-ops. `openOnFocus` flips in both directions:

- `<kbq-autocomplete openOnFocus>` → `''` → was falsy, **suppressed** open-on-focus; now `true`.
- `[openOnFocus]="0"` → was falsy; now `true` (`0 !== 'false'`).
- `[openOnFocus]="'false'"` → was truthy, opened; now `false`.

Worth calling out explicitly in the migration guide (see the note on `data.ts:81`), since the guide currently says the opposite.

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+}
+
+/** Collects template reference variable names bound to a `<kbq-autocomplete>` element. */
+class AutocompleteRefCollector implements Visitor {
+    readonly refs = new Set<string>();
+
+    visitElement(element: any): void {
+        if (element.name === AUTOCOMPLETE_ELEMENT) {
+            for (const attr of element.attrs ?? []) {
+                if (typeof attr.name !== 'string') continue;
```
lskramarov: **Trigger refs are never collected**, so `autocompleteDisabled` — the one `SIGNAL_MEMBER` that lives on `KbqAutocompleteTrigger` — is never migrated in templates.

The collector matches `element.name === 'kbq-autocomplete'`, but the trigger is a directive on an `<input>`/`<textarea>` reached through `exportAs: 'kbqAutocompleteTrigger'`. The idiom is already in this repo: `packages/docs-examples/components/autocomplete/autocomplete-with-footer/autocomplete-with-footer-example.ts:29`.

```html
<input kbqInput #t="kbqAutocompleteTrigger" [kbqAutocomplete]="auto" />
@if (t.autocompleteDisabled) { … }   <!-- left untouched -->
```
Post-migration that reads the `InputSignalWithTransform` object, which is always truthy — the branch renders unconditionally, with no compile error and no warning (the `.html` loop calls neither `logWarnings` nor `warnReceiverMembers`).

Two siblings already solve this: `app-switcher-signals/index.ts:250` (`TriggerRefCollector`, matching `attr.value === TRIGGER_EXPORT_AS`) and `navbar-signals-and-aria/index.ts:362-375` (an `EXPORT_AS_TO_TYPE` map feeding a collector that returns `Map<refName, typeName>` — exactly the two-type shape needed here).

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.member = RHS`. Every KbqAutocomplete signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
+    ) {
```
lskramarov: **Only `EqualsToken` counts as a write**, so every other assignment form falls through to the read branch at :168 and `--fix` writes syntactically invalid TypeScript into the consumer's file:

```ts
a.openOnFocus ||= false;  →  a.openOnFocus() ||= false;   // TS2364, invalid LHS
a.displayWith ??= fn;     →  a.displayWith() ??= fn;      // TS2364
a.showPanel &&= false;    →  a.showPanel() &&= false;     // TS2364
delete a.showPanel;       →  delete a.showPanel();        // TS2703
a.showPanel++;            →  a.showPanel()++;             // invalid
```

The file ends up in a *worse* state than untouched — mangled and non-compiling. `||=`/`&&=`/`??=` on `showPanel`/`displayWith` are plausible because both were writable before this PR.

Also silently missed with no rewrite and no warning: `const { isOpen } = a;` and `a!.isOpen`.

Cheap guard: treat any `ts.isBinaryExpression(parent) && parent.left === node` with an assignment operator (`ts.SyntaxKind.FirstAssignment..LastAssignment`), plus `isDeleteExpression` / `isPostfixUnaryExpression` / `isPrefixUnaryExpression` parents, as writes.

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+    'autocompleteDisabled',
+    'showPanel',
+    'isOpen'
+];
+
+/**
+ * Signal members that are writable via `.set(...)`. Every migrated member is either an `input()` or a
+ * read-only `computed`, so this is empty — writes are left untouched and covered by the warnings.
+ */
+export const WRITABLE_MEMBERS: ReadonlySet<string> = new Set<string>();
```
lskramarov: **The premise is false, so writes to `showPanel` get neither a rewrite nor a warning.**

`showPanel` is not an `input()` or a read-only `computed` — it is `readonly showPanel = signal(false)` (`autocomplete.component.ts:117`), published as `readonly showPanel: WritableSignal<boolean>` (`autocomplete.api.md:90`). Pre-PR it was `showPanel: boolean = false`, a plain public mutable field, so consumer writes are exactly what to expect. (`attached` is writable too, but it is new so nobody wrote it.)

Consequences:
- `WRITABLE_MEMBERS` being empty makes the `.set(…)` branch at `index.ts:157-162` unreachable dead code.
- `autocomplete.showPanel = x` is left as-is, and `warnPatterns` (:59-77) only matches `\.\s*isOpen\s*=`, so the consumer gets an unexplained compile error on a write that could have been auto-fixed to `.set(x)`.
- `index.spec.ts:124` tests the write path with `openOnFocus` — the one member for which the read-only claim *is* true — so the gap is untested.

Putting `showPanel` in `WRITABLE_MEMBERS` fixes the rewrite; `displayWith`/`autoActiveFirstOption`/`openOnFocus` writes still deserve a warn pattern, since the pre-PR source explicitly carried *"Your application code writes to the input. This prevents migration."* on `displayWith`.

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+/** Rewrites `ref.member` reads to `ref.member()` for the given refs, scoped to those exact identifiers. */
+function rewriteRefReads(template: string, refs: string[]): { content: string; changed: boolean } {
+    const members = SIGNAL_MEMBERS.join('|');
+    let content = template;
+    let changed = false;
+
+    for (const ref of refs) {
+        // `\bref\.(member)\b(?!\s*\()` — skip anything already invoked, so the rewrite is idempotent.
+        const pattern = new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()`, 'g');
+        const next = content.replace(pattern, '$1.$2()');
```
lskramarov: **The `SIGNAL_API_METHODS` negative lookahead that three siblings carry was dropped**, so the schematic corrupts code that was hand-migrated the way its own warning tells you to.

`app-switcher-signals/index.ts:287-293` (used by its template rewrite at :302), `button-toggle-signals-and-aria/index.ts:446-453` and `navbar-signals-and-aria` all build it as:
```ts
new RegExp(`\\b(${escapeRegExp(ref)})\\.(${members})\\b(?!\\s*\\()(?!\\s*\\.\\s*(?:${methods})\\b)`, 'g')
// methods = set, update, asReadonly, subscribe
```

`showPanel` and `attached` are public `WritableSignal`s, so `.set`/`.update` are reachable:
```html
<!-- before -->  (click)="auto.showPanel.set(true)"
<!-- after -->   (click)="auto.showPanel().set(true)"   → TypeError: …set is not a function
```
The docblock above claims idempotency; that holds only for `.set` on the TS side (`classifyAccess`:148) and for nothing at all here.

Two smaller regex issues while you're in here: the siblings capture the dot with the whitespace around it (`button-toggle-signals-and-aria/index.ts:450`, with a comment explaining that Angular's grammar allows `group . multiple` and a binding wrapped over two lines), so `{{ auto . isOpen }}` and `auto\n  .isOpen` are silently skipped here; and `\b` before a `$`-prefixed ref never matches, so `#$auto` refs are never migrated (`#_auto` works).

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
+    const start = node.getStart(sourceFile);
+    const end = node.getEnd();
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
+}
```
lskramarov: **Receivers are matched by source text over the whole enclosing function or class**, so any binder that shadows the name inside that range is wrongly rewritten:

```ts
// before
m(a: KbqAutocomplete) { [1].forEach((a: Other) => console.log(a.isOpen)); }
// after --fix
m(a: KbqAutocomplete) { [1].forEach((a: Other) => console.log(a.isOpen())); }  // TypeError at runtime
```
A class field's scope is the entire class (:116), so a nested class shadows too.

`navbar-signals-and-aria` already fixed both halves of this, with comments explaining why: `variableScope` (index.ts:122, over `isBlockScopeContainer` at :107) confines a block-scoped local to its block, and `receiverOf` (index.ts:176) picks the innermost enclosing declaration while `collectReceivers` (index.ts:136) deliberately collects declarations of *any* type so a shadowing binder is seen. This migration was forked from the older `alert-signals` shape instead.

The converse also bites: `@ViewChild(KbqAutocomplete) a: KbqAutocomplete | undefined` is a `UnionTypeNode`, not a `TypeReferenceNode`, so `isTypeReference` (:78) rejects it and `a.isOpen` is left reading an always-truthy signal object — no rewrite, no warning. `index.spec.ts:91` only covers the differently-*named* case.

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    return { content, changed };
+}
+
+/** Pass B (core) — parse a template, discover autocomplete refs, rewrite their value-safe signal reads. */
+async function migrateTemplate(template: string): Promise<{ content: string; changed: boolean }> {
+    if (!template.includes(AUTOCOMPLETE_ELEMENT)) return { content: template, changed: false };
+
+    const parsed = await parseTemplate(template);
+
+    if (!parsed.tree) return { content: template, changed: false };
```
lskramarov: **An unparseable template is skipped in silence.** `parseTemplate` returns `{ tree: undefined }` for any template `HtmlParser` rejects, and this returns `changed: false` with no log line — so the consumer's file is left un-migrated and they discover it as a broken build with no hint from `ng update`.

The three newest siblings all report it: `app-switcher-signals/index.ts:351` returns `{ …untouched(template), unparseable: true }`, threaded through :379 and surfaced at :459/:475 as `UNPARSEABLE_TEMPLATE_MESSAGE`. `button-toggle-signals-and-aria` and `navbar-signals-and-aria` do the same. This one was forked from the older `alert-signals` shape, before that flag existed.

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+            'back returned false whenever the panel had no options.'
+    },
+    {
+        anchor: AUTOCOMPLETE_ANCHOR,
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqAutocomplete\\b',
+        message:
+            'A KbqAutocomplete view/content query returns the component instance, whose `isOpen` and inputs ' +
+            'are now signals — reading one is a double call, e.g. `this.autocomplete().isOpen()`. Verify ' +
+            'query reads manually.'
+    }
```
lskramarov, line 129: **This warning fires on the case the schematic auto-fixes correctly, and the advice it gives would break it.**

The pattern matches the decorator form as well as the signal-query form:

| input | warning |
| --- | --- |
| `readonly a = viewChild(KbqAutocomplete);` | correct |
| `@ViewChild(KbqAutocomplete) a: KbqAutocomplete;` | **false positive** |

For the second, `collectReceivers` picks the field up (that is exactly what `index.spec.ts:73` asserts) and emits the correct `this.autocomplete.openOnFocus()` — while this warning simultaneously tells the user *"reading one is a double call, e.g. `this.autocomplete().isOpen()`"*. Following it yields `this.autocomplete()` on a non-callable field.

Anchoring on `viewChild|contentChild` only (the lowercase signal-query functions) would remove the false positive; `index.spec.ts:191` exercises only that form, and `index.spec.ts:73` never inspects `messages`, so nothing catches this today.

packages/schematics/src/migrations/autocomplete-signals/index.ts
```diff
@@ -0,0 +1,463 @@
...
+    visitAll(collector, (parsed.tree as { rootNodes: unknown[] }).rootNodes);
+
+    if (collector.refs.size === 0) return { content: template, changed: false };
+
+    return rewriteRefReads(template, [...collector.refs]);
+}
+
+/** Interior `[start, end]` ranges of inline `@Component({ template: '…' })` string literals. */
+function collectInlineTemplateRanges(sourceFile: ts.SourceFile): Array<{ start: number; end: number }> {
+    const ranges: Array<{ start: number; end: number }> = [];
```
lskramarov: `collectInlineTemplateRanges` is already exported from `packages/schematics/src/utils/typescript.ts:74` — and **this file already imports from that exact module** on line 7 (`forEachClass, parseTemplate`), where `forEachClass` is used *only* by this duplicate. The body is byte-identical to the shared one minus its JSDoc.

Seven siblings import it instead (`app-switcher-signals`, `button-state-and-styles`, `button-supported-colors`, `button-toggle-signals-and-aria`, `dropdown-demote-overlay`, `list-tree-multiple-input`, `navbar-signals-and-aria` — all on their line 7).

The fix is one word on line 7 plus deleting lines 318-353. The cost of keeping it: this becomes the 6th private copy, and a fix to the shared version (the `${…}` template-literal case its docblock reasons about, or `templateUrl`) reaches seven migrations and silently skips this one. Drift is already visible — `tag-slots/index.ts:141` renamed its locals and no longer hashes the same as the rest.

While you're here: `migrateInlineTemplates` (:357) parses the file before checking for the element at all — the `if (!template.includes(AUTOCOMPLETE_ELEMENT))` guard lives inside `migrateTemplate` (:303) and only fires after the parse and the `forEachClass` walk. `referencesAutocomplete` admits any file that merely imports the package, so most of those parses find no inline template at all.

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+];
+
+/** Printed once per project, after the per-file reports. */
+export const SUMMARY = [
+    '  `autoActiveFirstOption`, `openOnFocus` and `kbqAutocompleteDisabled` are `booleanAttribute` inputs. ' +
+        'A valueless attribute means true now, where it used to pass the empty string — `coerceBooleanProperty` ' +
+        'already treated that as false, so only markup that relied on the empty string reading as false changes.',
+    '  Binding `[autoActiveFirstOption]` now overrides KBQ_AUTOCOMPLETE_DEFAULT_OPTIONS even when the bound ' +
+        'value is `undefined`. The default only applies to an input nobody bound, so leave it unbound to let ' +
+        'the token decide.',
```
lskramarov: **"now" is wrong — a bound `undefined` already beat the token before this PR**, so this note documents a non-change.

The accessor this replaces coerced on every binding write:
```ts
set autoActiveFirstOption(value: boolean) { this._autoActiveFirstOption = coerceBooleanProperty(value); }
```
and the token default was assigned once in the constructor (`this._autoActiveFirstOption = !!defaults.autoActiveFirstOption`). Angular calls a setter for every binding update including the first, whatever the value, so `[autoActiveFirstOption]="undefined"` ran `coerceBooleanProperty(undefined)` → `undefined != null` → `false` and overwrote the token exactly as it does now.

With this and the two notes above it, **all three entries in `SUMMARY` are factually wrong** — that is the whole of what `ng update` prints once per project. Same claim in `README.md:47`, `docs/guides/migration.en.md:1070`, `docs/guides/migration.ru.md:1074` and both JSON descriptions; `README.md` even builds advice on it ("This one bites when converting a field write into a binding").

The behaviour is worth documenting — it is just not new, so drop the "now", or drop the note.

packages/schematics/src/migrations/autocomplete-signals/data.ts
```diff
@@ -0,0 +1,90 @@
...
+    'Set `class` on <kbq-autocomplete> as before — the classes still land on the overlay panel.';
+
+export interface WarnPattern {
+    /** Owner of the member. The pattern is only evaluated for files that also name it. */
+    anchor: string;
+    pattern: string;
+    message: string;
+}
+
+const AUTOCOMPLETE_ANCHOR = '\\bKbqAutocomplete\\b';
```
lskramarov: **This anchor cannot match the file shape the migration deliberately widened its filter to catch**, so neither warn pattern ever fires for it.

`\b` fails between `KbqAutocomplete` and the `M` of `KbqAutocompleteModule`. Meanwhile `referencesAutocomplete` (`index.ts:384`) uses `/\bKbqAutocomplete\w*\b/` plus the package path plus `<kbq-autocomplete`, with a docblock saying exactly why: *"a component that only imports `KbqAutocompleteModule` names no type."*

So a standalone component that imports `KbqAutocompleteModule`, renders `<kbq-autocomplete #auto="kbqAutocomplete">` in an inline template and does `this.auto().isOpen = false` **is** scanned (it increments `consumers` and gets the project summary) but never sees the `isOpen =` warning that exists precisely to explain the error it is about to hit. That is the most likely shape for such a write to live in.

`AUTOCOMPLETE_ANCHOR` should use the same `\w*` the file filter uses, or reuse `referencesAutocomplete` outright.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **The rewritten intro now contradicts its own subsections.**

Section 18 has four: `#### Autocomplete` (:1052), `#### Search expandable` (:1078), `#### Split button` (:1100), `#### Title` (:1122). This sentence says *"the second is the one each subsection below belongs to"* — but the first-wave list in the same sentence names `search-expandable`, `split-button` and `title`, i.e. three of those four. And `autocomplete` appears in neither enumeration, so a reader scanning the two lists never places it.

The pre-PR sentence ("Ten components went through a full review in 20.3.0: … search-expandable, … split-button, title …") was consistent with the subsections because there was only one wave. If the two-wave framing is worth keeping, the second wave needs its own explicit list (just `autocomplete`); otherwise adding `autocomplete` to the original enumeration is a one-word fix.

`docs/guides/migration.ru.md:1042` has the identical problem, with the same four subsections at :1056/:1082/:1104/:1126.

lskramarov (commented):
### Follow-up: gap sweep

A final pass over the diff, run against a real build of the PR head, turned up three more. Two of them land in the same `SUMMARY` array as earlier comments — with this one, **all three notes `ng update` prints are factually wrong**.

Things the sweep checked and found clean, for the record: `api-extractor onlyCheck` (the `.api.md` guard itself is up to date — the red `api` job is the separate `check-public-api-any` ratchet), `prettier`, `cspell` and `eslint --max-warnings=0` on all 15 changed files, and `jest` on `autocomplete.spec.ts` (123), `tags` (222) and `autocomplete-signals/index.spec.ts` (16). `should let the input override the token` is not vacuous, `zone` is correctly reassigned after `TestBed.resetTestingModule()`, panel ids are unique and stable per instance, and dynamic `[class]` does correctly add *and* remove classes on the panel.

packages/components/autocomplete/autocomplete.component.ts
artembelik: Взял второй вариант: вход принимает всё то же, что и обычная привязка `[class]` - строку, массив, `Set` и объект, - и сводит это к строке в трансформе.

packages/schematics/src/migrations/autocomplete-signals/index.ts
artembelik: Затенение чиню разрешением имени через объявления, а union и прочие неразрешимые формы не резолвлю, а печатаю списком строк с пометкой посмотреть руками.

tools/public_api_guard/components/autocomplete.api.md
artembelik: Записал baseline в мерж-коммите `0e9da28c` вместе с подтягиванием main, так что TS2445 в tree-selection ушёл тем же коммитом.

docs/guides/migration.en.md
artembelik: Убрал деление на волны целиком, а не дописал autocomplete в список: перечисление первой волны устаревало с каждым влившимся ревью.

lskramarov (approved):

PR #1988, opened 2026-09-03
fix(code-block)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `code-block`, in the same shape as the 21.0.0 component reviews.

### `maxHeight` lied about being required

```ts
readonly maxHeight = input<number, unknown>(undefined!, { transform: numberAttribute });
```

A code block with no `[maxHeight]` binding reported `undefined` from a non-nullable type: an assignment to a `number` held `undefined`, and `maxHeight() > 0` was a NaN comparison that never matched. It reports `number | undefined` now — the same pathology, and the same fix, as `KbqSplitButton.disabled` in the first review wave.

### `KbqCodeBlockHighlight.file` was write-only

```ts
@Input({ required: true })
set file(file: KbqCodeBlockFile) { … this.highlight(file); }
```

A required input with no getter, whose setter kicked off highlighting as a side effect — which is exactly why the automated signal migration skipped it. It is `input.required()` driven by an effect now, so it can finally be read.

### The max-height binding never re-evaluated on its own

`calculatedMaxHeight` was a getter read from `[style.max-height.px]`, so it only recomputed when something else marked the view dirty. It is a `computed` over `maxHeight` and `viewAll` now.

That binding turned out to have no test coverage at all — the new test caught a mistake I made in the very same change (binding the computed without calling it), which is worth knowing about the component.

### What deliberately stays an accessor input

`softWrap`, `viewAll`, `canDownload`, `files`, `activeFileIndex` and `hideTabs` are `WritableSignal`s over a `@docs-private` backing `input()` that carries the transform and aliases the public attribute name. A `model()` would have been the obvious shape and does not fit: `ModelOptions` carries no `transform`, and without one a valueless attribute passes the empty string and silently turns the feature off — `<kbq-code-block softWrap>` is in `e2e.ts` under a screenshot baseline and `<kbq-code-block canDownload />` is in the published docs in both languages.

**Template bindings are untouched** — `[softWrap]`, `[(viewAll)]` and the valueless form all keep working. What changes is programmatic access: a read becomes a call and a write becomes `.set(…)`. The deprecated `canLoad` and `codeFiles` stop writing into their replacements and fill in for them instead, so which of each pair wins no longer depends on attribute order.

## Migration

`code-block-signals` runs from `ng update @koobiq/components@21` (registered for `21.0.0-0`). It rewrites reads to calls and plain writes to `.set(…)` on the six signal members, in TypeScript and through template reference variables on `<kbq-code-block>`. It reports what has no mechanical translation: compound assignments and increments, the backing inputs, the deprecated `canLoad` / `codeFiles` aliases, `maxHeight` — which reports `number | undefined` and never NaN now — and `KbqCodeBlockHighlight.file`, which was a write-only setter and is a readable required input.

## Testing

- `code-block.spec.ts`: 50 → 66 tests. New coverage for the valueless boolean attribute, the deprecated `canLoad` / `codeFiles` fallbacks, an out-of-range and a non-numeric `activeFileIndex`, an empty `files`, a `maxHeight` that arrives after init, a failed `highlight.js` load, and the tab bar coming back once the files are named.
- `code-block-signals/index.spec.ts`: 13 tests.
- Full `packages/components` (4996 tests) and `packages/schematics` (438 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


lskramarov (commented):
## Code review — `max` effort

15 findings, most severe first. Everything below was reproduced by running code in a worktree at `21a06a93` and, where the question was "is this new?", re-run against the merge base `4bb1c9927` — each comment says which.

Headline: **the migration is registered at `20.3.0-0`, which `ng update` will never select** now that `20.3.0` is tagged and `main` has moved every sibling to `21.0.0-0`. That is the semantic conflict behind the textual one in `migrations.json`.

Two behaviour changes look like regressions against the merge base:

1. `set files` now emits `activeFileIndexChange` and calls `scrollTo()` from inside an input setter — a parent that shrinks `files` and moves `[(activeFileIndex)]` in the same tick gets the wrong file and has its own field silently overwritten (`NG0100` in the other binding order).
2. The highlight effect's dependency set is history-dependent: `startFrom`/`singleLine` are dead until a `file` change routes through the synchronous branch, then live forever.

The rest split into incomplete fixes in touched code (`setupContentOverflowDetection` never re-arms; `set activeFileIndex` has no clamp; empty `files` still crashes; `hideTabs` still latches; `maxHeight` can still be `NaN`), one altitude call (`loadOnce()` is scoped to the directive while `initLineNumbersPlugin` is global), and notes on the guide, the schematic's matching, and two of the new tests.

Verified clean, for the record: `check-api` passes for `components/code-block` after a real build, the `any`/`unknown` ratchet is unchanged, `prettier`/`eslint`/`cspell` pass on every changed file, the Jest suite is 55/55, effects still flush before `ngAfterViewInit` so the `pending` contract is preserved, the `model()`-cannot-carry-a-transform justification in the code comment is correct, and no `maxHeight()` or `.file =` call site in this repo breaks on the new types.

_Generated with Claude Code_

packages/schematics/src/migrations.json
```diff
@@ -105,6 +105,11 @@
             "version": "20.3.0-0",
             "description": "Reports the KbqSplitButton members whose type changed in the split-button review. disabled reports boolean | undefined instead of boolean: the backing field has no initializer, so a control with no [disabled] binding always returned undefined behind a non-nullable type, and the call sites that assigned it to a boolean or compared it against false were quietly wrong. The protected buttons content query moved from QueryList<KbqButton> to a signal query, so a subclass reading buttons.changes or buttons.length has to read buttons() instead. Warn-only: narrowing boolean | undefined back to boolean is a decision the call site owns. Notes the silent part too — an empty <kbq-split-button> no longer throws outside dev mode.",
             "factory": "./migrations/split-button-optional-disabled/index"
+        },
+        "code-block-optional-max-height": {
+            "version": "20.3.0-0",
```
lskramarov: **`20.3.0-0` is already behind the released tag — `ng update` will skip this migration.**

`origin/main` is `"version": "20.3.0"`, the `20.3.0` tag exists, and main has moved 30 migrations to `21.0.0-0` (including `split-button-optional-disabled`, the sibling this one was copied from), leaving a single leftover at `20.3.0-0`. This branch predates that bump, so all 18 review entries here still read `20.3.0-0` and the new one blends in.

The CLI builds the range `>${from} <=${to.split('-')[0]}`. A consumer on the published `20.3.0` running `ng update @koobiq/components@21` gets `>20.3.0 <=21.0.0`, which `20.3.0-0` does not satisfy — the migration is filtered out and the report this PR exists to deliver never reaches anyone.

This is the semantic conflict hiding behind the textual conflict in `migrations.json`/`collection.json`: a merge that keeps both sides leaves this entry as the only stale one. The same staleness is in `README.md:3` ("registered for `20.3.0-0`", `ng update @koobiq/components@20`) and in both migration guides.

packages/components/code-block/code-block.ts
```diff
@@ -249,31 +277,36 @@ export class KbqCodeBlock implements AfterViewInit {
...
-        return this._files;
+        return this._files();
     }
 
     set files(files: KbqCodeBlockFile[]) {
-        this._files = files;
+        this._files.set(files);
 
-        if (this._files.length < this.activeFileIndex) {
+        if (files.length <= this.activeFileIndex) {
```
lskramarov: **Regression: this emits `activeFileIndexChange` from inside an input setter and clobbers a `[(activeFileIndex)]` update made in the same tick.**

`onSelectedTabChange(0)` emits the output and calls `scrollTo()`, all while the parent's template function is still executing.

Verified in Jest at both revisions. Host `<kbq-code-block [files]="files" [(activeFileIndex)]="index" />`, 3 files, `index = 2`; parent then does `files = files.slice(0, 2); index = 1;` in one tick:

| | rendered | parent `index` |
|---|---|---|
| `4bb1c9927` (base) | `b.txt` ✅ | `1` |
| this PR | `a.txt` ❌ | silently rewritten to `0` |

`2 <= 2` fires, `activeFileIndexChange.emit(0)` writes `ctx.index = 0`, and `ɵɵtwoWayProperty('activeFileIndex', …)` then reads the clobbered `0`. `checkNoChanges` sees `0 === 0`, so nothing is reported.

With `[(activeFileIndex)]` declared **before** `[files]` — the order used by `packages/components/code-block/e2e.ts` and both `code-block-with-tabs` examples — the same emit throws `NG0100: ExpressionChangedAfterItHasBeenCheckedError … Previous value: '2'. Current value: '0'`. Pre-PR that only fired for `length < index`; the `<=` widens it into the `length === index` case the new test advertises.

`scrollTo()` during the input-write phase also allocates a `toObservable` effect + `ReplaySubject` (when `pending()` is true) that only unwind on component destroy.

Suggest clamping the index in a derived value rather than driving tab-change side effects from an input setter.

packages/components/code-block/code-block.ts
```diff
@@ -500,7 +531,11 @@ export class KbqCodeBlock implements AfterViewInit {
         if (!this.maxHeight()) return;
```
lskramarov: **A `maxHeight` that arrives after init clips the content with no way to expand it.**

This gate runs once, from `ngAfterViewInit`, and nothing re-arms it — so `checkOverflow()`, the `toObservable(pending)` subscription and `sharedResizeObserver.observe(...)` are all skipped and `contentExceedsMaxHeight` stays `false` for the component's life.

Verified in Jest with `MockSharedResizeObserver` (`<pre>` offsetHeight 500):

- control, `[maxHeight]="200"` from init → `.kbq-code-block__view-all__button` **renders**
- `limit` starting `undefined`, set to `200` after init → `.kbq-code-block__main` gets `max-height: 200px` but the button is **absent**

`canCodeContentBeFocused` then returns `!this.calculatedMaxHeight()` → `false` → `tabIndex = -1`, so the clipped remainder is unreachable by keyboard too.

The same one-shot wiring leaves a *changed* limit stale in the other direction: raising it 200 → 1000 over 300px of content keeps the button showing (also verified).

This PR made `calculatedMaxHeight` a `computed` so the style follows `maxHeight`, and added `should follow a maxHeight that changes after init` (spec:986), which drives exactly this path and asserts only `main.style.maxHeight` — so the two halves now disagree and a test cements the broken one. `maxHeight` is a signal; this belongs in an `effect`.

packages/components/code-block/code-block-highlight.ts
```diff
@@ -124,6 +140,18 @@ export class KbqCodeBlockHighlight {
     /** Whether to display line numbers for single line code block. */
     readonly singleLine = input<boolean, unknown>(false, { transform: booleanAttribute });
 
+    /** Loads highlight.js at most once, so overlapping `file` changes share a single import. */
+    private loadOnce(): Promise<void> {
+        this.hljsLoading ??= this.load(this.config ?? {}).finally(() => {
```
lskramarov: **A failed load un-latches `hljsLoading` but never `_pending`, so `pending()` stays `true` forever.**

`load()` sets `this._pending.set(true)` at line 156 and only `highlight()` clears it at line 213.

Verified with a rejecting `core` loader: after the first failure `pending()` is `true`; a later `file` change **does** retry (attempts 1 → 2, so the `hljsLoading` reset works) but `pending()` is still `true` afterwards, and the rejection escapes `this.loadOnce().then(…)` (line 129, no `.catch`) through zone.js as an uncaught error on every attempt.

Consequences in the host:

- `KbqCodeBlock.scrollTo()` takes the `highlight?.pending()` branch and subscribes to `filter((pending) => !pending), take(1)`, which never emits — `toggleViewAll()`'s collapse-and-scroll-to-top and `onSelectedTabChange()`'s scroll reset are silently dead for the rest of the page, and each call leaks another `toObservable` effect + `ReplaySubject`.
- `setupContentOverflowDetection`'s post-load `checkOverflow` never runs.

Triggers: a 404 on the lazy `import('highlight.js')` chunk after a deploy, offline, or one rejecting entry in `kbqCodeBlockHighlightJsConfigProvider({ languages })`.

The comment on the next line says the failure is not latched — the *load* is not, `pending` is. Clearing it where it is set covers both halves:

```ts
private async load({ core, languages }: KbqCodeBlockHighlightJsConfig): Promise<void> {
    this._pending.set(true);

    try {
        // …
    } catch (error) {
        this._pending.set(false);
        throw error;
    }
}
```

…plus a rejection handler on the `loadOnce()` call at line 129 so the failure does not escape as an uncaught error.

packages/components/code-block/code-block.ts
```diff
@@ -249,31 +277,36 @@ export class KbqCodeBlock implements AfterViewInit {
...
     /** Defines which file (index) is active. */
-    // TODO: Skipped for migration because:
-    //  Your application code writes to the input. This prevents migration.
-    @Input({ transform: numberAttribute }) activeFileIndex = 0;
+    @Input({ transform: numberAttribute })
+    get activeFileIndex(): number {
+        return this._activeFileIndex();
+    }
+
+    set activeFileIndex(value: number) {
```
lskramarov: **The out-of-range fix is one-sided: this setter has no clamp, and `numberAttribute` lets `NaN` walk past the new guard.**

Two verified paths, both ending in `TypeError: Cannot read properties of undefined (reading 'link')` from `@if (file.link)` (`code-block.html:78`) over `@let file = files[activeFileIndex]`:

1. **Ordering.** `<kbq-code-block [files]="files" [activeFileIndex]="index" />` — `[files]` is written first, so the new guard reads the *stale* index (`2 <= 0` → false, no reset), then this setter installs the out-of-range value. Shrinking 3 files → 2 while moving `index` 0 → 2 in one tick crashes; identical at the merge base, so the `<` → `<=` fix never reaches this half. The new test passes only because `BaseCodeBlock` moves `activeFileIndex` to 2 in a *prior* `detectChanges()`.
2. **NaN.** `maxHeight` one input above got `value == null ? undefined : numberAttribute(value)`, but `activeFileIndex` kept the bare transform — and `numberAttribute(undefined)` is `NaN`. With `index?: number` unset, `_activeFileIndex` becomes `NaN` (verified), the guard `3 <= NaN` is `false`, and the template throws — behind a type that says `number`.

A single clamped `computed` over `_files()` and `_activeFileIndex()` would cover both halves and the empty-`files` case together, and would remove the need to emit and scroll from `set files`.

packages/components/code-block/code-block-highlight.ts
```diff
@@ -105,17 +106,32 @@ export class KbqCodeBlockHighlight {
...
-            this.highlight(file);
-        }
+    constructor() {
+        effect((onCleanup) => {
+            const file = this.file();
+
+            if (!this.window) return;
+
+            if (this.hljs) {
+                this.highlight(file);
```
lskramarov: **The effect's dependency set is history-dependent: `startFrom`/`singleLine` are tracked here but not on the `.then()` path.**

`highlight()` reads `this.startFrom()` (line 206) and `this.singleLine()` (line 207). Called here it runs inside the effect's consumer, so both become dependencies; called from `loadOnce().then(…)` (line 131) it runs in a microtask and neither is.

Verified by counting `highlight()` calls on `<code kbqCodeBlockHighlight [file]="f" [startFrom]="n">`:

| step | base `4bb1c9927` | this PR |
|---|---|---|
| after initial load | 1 | 1 |
| `startFrom` 1 → 10 | 1 | 1 (dead) |
| `f` changes | 2 | 2 |
| `startFrom` 10 → 20 | **2** | **3** |

`hljs` is null on every instance's first run, so the first pass always takes the untracked path and the set stays `{file}`; once a `file` change routes through this branch it is rebuilt as `{file, startFrom, singleLine}`. Same input change, two behaviours depending on what happened earlier.

Once live, each gutter-only step also costs a full `hljs.highlight()` re-parse, a `DomSanitizer.sanitize` pass, a `lineNumbersValue()` DOM rebuild and a whole-subtree `innerHTML` write.

`KbqCodeBlockHighlight` is exported from `public-api.ts` with both inputs public, so this is consumer-reachable even though `code-block.html:111` binds only `[file]`. Pick one: read both inputs in the effect body, or wrap this call in `untracked()`.

packages/components/code-block/code-block.ts
```diff
@@ -249,31 +277,36 @@ export class KbqCodeBlock implements AfterViewInit {
...
         }
 
-        if (this._files.length === 1 && !this._files[0].filename) {
+        if (files.length === 1 && !files[0].filename) {
             this.hideTabs = true;
         }
     }
 
-    private _files: KbqCodeBlockFile[] = [];
+    private readonly _files = signal<KbqCodeBlockFile[]>([]);
```
lskramarov: **The default empty `files` crashes the template, and the widened guard cannot rescue it.**

Verified at both revisions: `<kbq-code-block />` and `<kbq-code-block [files]="[]" />` both throw `TypeError: Cannot read properties of undefined (reading 'link')` from `@if (file.link)` (`code-block.html:78`) over `@let file = files[activeFileIndex]` (line 1). `{{ file.content }}` and `[file]="file"` on line 111 feed the same `undefined` into `input.required<KbqCodeBlockFile>()`.

The new guard *does* run (`0 <= 0`), but `onSelectedTabChange(0)` opens with `if (this.activeFileIndex !== index)` — `0 !== 0` is false — so the "reset to 0" fix no-ops here.

Reachable from the repo's own consumer: `apps/docs/.../docs-live-example-viewer.ts:246` binds `[codeFiles]` to `codeFiles.filter((file) => file.content)` with no non-empty floor, and any `[files]="items$ | async"` starts empty.

Pre-existing, but the migration note this PR adds names "the block then rendered from an undefined file" as the bug being fixed. Guarding the render (`@if (activeFile(); as file)`) covers it; guarding only the setter does not.

packages/components/code-block/code-block.ts
```diff
@@ -249,31 +277,36 @@ export class KbqCodeBlock implements AfterViewInit {
...
+        this._files.set(files);
 
-        if (this._files.length < this.activeFileIndex) {
+        if (files.length <= this.activeFileIndex) {
             this.onSelectedTabChange(0);
         }
 
-        if (this._files.length === 1 && !this._files[0].filename) {
+        if (files.length === 1 && !files[0].filename) {
             this.hideTabs = true;
```
lskramarov: **`hideTabs` latches to `true` and never comes back, and `hideTabsChange` re-emits on every `files` write.**

Verified in Jest with `<kbq-code-block [files]="files" (hideTabsChange)="emissions.push($event)" />`:

- start with one unnamed file → tabs hidden, `emissions = [true]`
- assign two **named** files → `.kbq-code-block__header-tabs` is *still* absent; the tab bar never returns, so the second file is unreachable and only `files[0]` renders
- two further single-unnamed-file assignments → `emissions = [true, true, true]`

`set hideTabs` (line 328) is `this._hideTabs.set(value); this.hideTabsChange.emit(value);` with no equality check, so a `[(hideTabs)]` consumer gets a write-back on every `files` assignment even when nothing changed. The existing suite works around the latch by explicitly resetting `componentInstance.hideTabs = false` (spec:305).

Both lines sit inside the setter this PR rewrites. The derived value wants to be `hideTabs || (files.length === 1 && !files[0].filename)`, and the emit wants an equality guard.

packages/components/code-block/code-block.ts
```diff
@@ -200,31 +221,40 @@ export class KbqCodeBlock implements AfterViewInit {
      * Maximum height of the code block content, other parts will be hidden.
      * Can be toggled by `viewAll` property.
      */
-    readonly maxHeight = input<number, unknown>(undefined!, { transform: numberAttribute });
+    readonly maxHeight = input<number | undefined, unknown>(undefined, {
+        transform: (value) => (value == null ? undefined : numberAttribute(value))
+    });
```
lskramarov, line 223: **`maxHeight()` can still be `NaN` behind the new `number | undefined`, and the documented `?? 0` does not catch it.**

`numberAttribute` (`core.mjs:3895`) falls back to `NaN` for anything not cleanly numeric — verified: `''` → `NaN`, `'200px'` → `NaN`, `'abc'` → `NaN`. So `<kbq-code-block maxHeight>` (valueless attribute, Angular passes `''`) or `[maxHeight]="cfg.height"` with a non-numeric string still yields `NaN` from a type that promises `number | undefined`.

A consumer following the migration table (`migration.en.md:1067`, `README.md:32`, `data.ts:48` — "`?? 0` for the common reading") writes `block.maxHeight() ?? 0` and gets `NaN` — exactly the NaN-arithmetic bug the migration says it fixes. Internally the PR papers over the same hole twice with guards it had to add (`maxHeight && maxHeight > 0` at line 234, `!!maxHeight &&` at line 537).

Separately, the guide's "Nothing about the runtime value changed" is not accurate: `[maxHeight]="undefined"` / `[maxHeight]="null"` returned `NaN` at the merge base and returns `undefined` now — which is the only reason the new spec `'should report undefined for an unbound maxHeight'` passes.

```suggestion
    readonly maxHeight = input<number | undefined, unknown>(undefined, {
        transform: (value) => {
            const parsed = numberAttribute(value);

            return Number.isFinite(parsed) ? parsed : undefined;
        }
    });
```

`numberAttribute(null)` and `numberAttribute(undefined)` are already `NaN`, so the explicit null branch is not needed. That makes the published type true and lets both falsiness guards go.

packages/components/code-block/code-block-highlight.ts
```diff
@@ -124,6 +140,18 @@ export class KbqCodeBlockHighlight {
     /** Whether to display line numbers for single line code block. */
     readonly singleLine = input<boolean, unknown>(false, { transform: booleanAttribute });
 
+    /** Loads highlight.js at most once, so overlapping `file` changes share a single import. */
+    private loadOnce(): Promise<void> {
```
lskramarov: **`loadOnce()` de-duplicates per directive instance, but the work it guards is application-global.**

Verified: a host with three `code[kbqCodeBlockHighlight]` elements produces `loadCalls = 3` and **3 new `<style>` elements in `<head>`**.

`load()` ends in `initLineNumbersPlugin` (line 172), which does `this.window['hljs'] = instance` (line 227 — already global) and then runs `highlightJSLineNumbersPlugin`, whose body unconditionally calls `addStyles()` (line 394, `head.appendChild(css)`) and `d.addEventListener('copy', …)` (line 360). There is no `ngOnDestroy`/`DestroyRef`/`removeEventListener` in the file, so the listeners outlive their directives: a docs component page renders 10–30 code blocks and every "show source" toggle adds another, so the count grows monotonically with clicking and SPA navigation, and all of them run `preventDefault()` + `setData()` on a single copy gesture.

The intra-instance race this fixes is real and the fix works (3 rapid `file` changes now yield 1 load and 1 highlight of the last file — verified), but it is the rare case. A `providedIn: 'root'` loader holding the promise and running `initLineNumbersPlugin` once would kill both duplications and give the copy listener an owner with a lifetime.

packages/components/code-block/code-block.ts
```diff
@@ -345,7 +376,7 @@ export class KbqCodeBlock implements AfterViewInit {
 
         const element = this.scrollableCodeContent()?.getElementRef().nativeElement;
 
-        return element && this.hasScroll(element) && !this.calculatedMaxHeight;
+        return element && this.hasScroll(element) && !this.calculatedMaxHeight();
```
lskramarov: **Four layout-forcing reads happen before the cheap memoized guard that would have skipped them.**

`hasScroll` destructures `{ scrollHeight, scrollWidth, clientHeight, clientWidth }` (line 561). This getter is reached from `codeContentTabIndex` (line 364), bound as `[tabIndex]="codeContentTabIndex"` at `code-block.html:106`, so it re-evaluates on every update pass. When `maxHeight` is set and `viewAll` is off, `calculatedMaxHeight()` is non-null and the whole expression is `false` — but the reflow already happened.

Worse, the binding on the line above writes `[style.max-height.px]="calculatedMaxHeight()"` to the *same* element, so a changed max-height guarantees a style-write-then-layout-read on one element in one pass.

`calculatedMaxHeight` is a `computed` now and is already evaluated by `code-block.html:105` in the same pass, so reading it first is just a cached version check:

```suggestion
        return !this.calculatedMaxHeight() && !!element && this.hasScroll(element);
```

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **This sentence contradicts the subsections it points at, and `main` has already replaced the whole paragraph.**

§18 contains exactly four `####` subsections: `Code block` (1052), `Search expandable` (1078), `Split button` (1100), `Title` (1122). The sentence lists search-expandable, split-button and title as *first* wave and then says "the second is the one each subsection below belongs to" — so a reader concludes those three were reviewed twice in 20.3.0. `migration.ru.md:1042` repeats it verbatim.

Meanwhile `origin/main` already retitled the section to `### 18. Component review (21.0.0)` and rewrote the intro to:

> Components went through a full review in 21.0.0. … Only the changes that reach a consumer are listed here, so a component whose review changed nothing a consumer can see has no subsection below.

which solves the same problem properly — so this hunk is both a merge conflict and a regression on top of it. Rebasing onto main's rewrite drops the wave framing entirely.

One more inaccuracy in the same hunk: it says `softWrap`, `viewAll`, `canDownload`, `activeFileIndex` and `files` "stay accessor inputs". At the merge base four of the five were plain public fields (`@Input({ transform: booleanAttribute }) softWrap: boolean = false;`) — they *became* accessors here, which moves them off the instance (`Object.keys`, spread, `JSON.stringify`) and makes a subclass's `override softWrap = true` shadow the accessor under a consumer's default `useDefineForClassFields: true`.

packages/schematics/src/migrations/code-block-optional-max-height/index.ts
```diff
@@ -0,0 +1,65 @@
+import { Path } from '@angular-devkit/core';
+import { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { logMessage } from '../../utils/messages';
+import { setupOptions } from '../../utils/package-config';
+import { CODE_BLOCK_PACKAGE, CODE_BLOCK_TYPE, HIGHLIGHT_TYPE, SUMMARY, warnPatterns } from './data';
+import { Schema } from './schema';
+
+const LABEL = '[code-block-optional-max-height]';
+const EXTENSIONS = ['.ts'];
```
lskramarov: **`.ts`-only plus the `\bKbqCodeBlock\b` anchor silences the migration for the two call-site shapes it exists to report.**

Verified the regex semantics in node: `\bKbqCodeBlock\b` is **false** for `KbqCodeBlockModule`, `KbqCodeBlockHighlight`, `KbqCodeBlockFile` and the `kbqCodeBlock` exportAs — true only for the bare identifier. So

```ts
import { KbqCodeBlockModule } from '@koobiq/components/code-block';
// template: <kbq-code-block #b="kbqCodeBlock" [files]="f" />{{ b.maxHeight() }}
```

passes `referencesCodeBlock` on the package string, increments `consumers`, and is then folded into "N file(s) reference the component" with no `maxHeight` warning.

And because `KbqCodeBlock` carries `exportAs: 'kbqCodeBlock'` (`code-block.ts:140`) with `strictTemplates: true` (`tsconfig.json:117`), the breakage most likely to bite — `<kbq-code-block #cb="kbqCodeBlock" /> <div [style.height.px]="cb.maxHeight() * 2">` in an external `.html` — is never opened at all. The file comment's justification ("the component's template surface did not change, and every binding accepts exactly what it did before") answers the *inbound* question; the migration is about reads going *out*.

In the other direction the patterns are loose (also verified): `\.\s*maxHeight\b` matches `dialogConfig.maxHeight`, `// note about .maxHeight` and `'.maxHeight'` inside a string; `\.\s*file\s*=[^=]` matches `this.file = e.target.files[0];` and `// old: d.file = {}` in any file that also happens to mention `KbqCodeBlockHighlight`.

Minor: `reported++` runs at most once per (file, pattern) pair, so "N call site(s) reported" is neither call sites nor files.

packages/components/code-block/code-block.spec.ts
```diff
@@ -939,4 +939,81 @@ describe(KbqCodeBlock.name, () => {
...
+            .injector.get(KbqCodeBlockHighlight);
+
+        expect(highlight.file().filename).toBe('index.html');
+
+        fixture.componentInstance.activeFileIndex = 1;
+        fixture.detectChanges();
+
+        expect(highlight.file().filename).toBe('main.ts');
+    });
+    it('should follow a maxHeight that changes after init', () => {
```
lskramarov: **This duplicates the first half of the test above it and pins nothing the PR changed.**

Lines 987–995 are statement for statement lines 952–962: same `createComponent(BaseCodeBlock)`, same `maxHeight = 200`, same `querySelector('.kbq-code-block__main')`, same `expect(main.style.maxHeight).toBe('200px')`. Only the `400` step is new — and `maxHeight` was already an `input()` signal at the merge base, so writing the binding marks the OnPush view for refresh and the old `protected get calculatedMaxHeight()` is re-read and returns 400. The test goes green on pre-PR `code-block.ts`/`code-block.html`, so it pins none of the `computed` change. `'should drop the calculated max height once viewAll is on'` (line 951) is what actually pins it, because `toggleViewAll()` wrote a plain field before and dirtied nothing.

It is also the test that runs entirely inside the dead zone of `setupContentOverflowDetection` (see the comment on `code-block.ts:531`), so it documents the half-fixed state as correct.

Suggest folding the `400` step into the test above and spending the slot on what currently has no coverage at all: `loadOnce()`'s de-duplication, the failed-load path, and the effect's `cancelled` guard.

packages/components/code-block/code-block.spec.ts
```diff
@@ -939,4 +939,81 @@ describe(KbqCodeBlock.name, () => {
             expect(scrollSpy).toHaveBeenCalledWith({ top: 100 });
         }));
     });
+    it('should report undefined for an unbound maxHeight', () => {
```
lskramarov: **`maxHeight` is bound here, not unbound — the path the name describes stays untested.**

`BaseCodeBlock` declares `[maxHeight]="maxHeight"` (line 102) over `maxHeight: number | undefined = undefined` (line 137), so the transform *runs* and takes its new `value == null` branch. That is the interesting assertion — at the merge base the same binding produced `numberAttribute(undefined)` → `NaN` — but it is not what the name says.

A genuinely unbound `<kbq-code-block [files]="files" />` never invokes the transform: Angular uses the declared default, which was `undefined!` before and is `undefined` now, so that path returned `undefined` on both sides and nothing covers it.

Rename to say the input is bound to `undefined`, or add the unbound host so the name is true.

packages/components/code-block/code-block.ts
artembelik: клампящий `computed` сделал, а NaN отсекаю в сеттере: геттер обещает `number`, поэтому хранить NaN нельзя даже до клампа

packages/components/code-block/code-block-highlight.ts
artembelik: загрузка теперь не бросает, а возвращает результат - обработчику отказа неоткуда взяться в эффекте; `pending` снимается там же, где ставится, плюс `warn` в dev

packages/components/code-block/code-block-highlight.ts
artembelik: пока сделал минимум - плагин ставится один раз на инстанс hljs, поэтому `style` и слушатель `copy` заводятся по разу; корневой загрузчик вынес в отдельную задачу, он меняет область видимости `KBQ_CODE_BLOCK_HIGHLIGHT_JS_CONFIG`

packages/components/code-block/code-block.ts
artembelik: сделал как предложено. заодно заменил тест `should NOT hide tabs ... when set outside`: он фиксировал обратное и проходил только из-за порядка атрибутов в шаблоне хоста

packages/components/code-block/code-block.ts
artembelik: `suggestion` применил дословно, но оба falsy-guard оставил: без них TS ругается на `number | undefined` в сравнении, так что теперь они сужают тип, а не прикрывают NaN

packages/components/code-block/code-block-highlight.ts
artembelik: читаю оба входа в теле эффекта и передаю их в `highlight()` аргументом, чтобы на микрозадачной ветке значения были те же

docs/guides/migration.en.md
artembelik: вступление про волны уехало со слиянием main, а утверждение про аксессоры переписал в обоих гайдах, README и отчёте схематика

packages/components/code-block/code-block.spec.ts
artembelik: слот отдал под гейт переполнения при `maxHeight` после init и под провалившуюся загрузку; дедупликация `loadOnce` и `cancelled`-guard пока без покрытия

lskramarov (approved):

packages/components/code-block/code-block.ts
```diff
@@ -173,23 +174,36 @@ export class KbqCodeBlock implements AfterViewInit {
...
      * Output to support two-way binding on `[(softWrap)]` property.
      */
     readonly softWrapChange = output<boolean>();
 
+    /**
+     * Backing input of `viewAll`, in the same shape as `softWrapInput`.
+     *
+     * @docs-private
+     */
+    readonly viewAllInput = input(false, { alias: 'viewAll', transform: booleanAttribute });
```
NikGurev: есть смысл сделать model ,чтобы объединить viewAllInput и viewAllChange или есть ограничения? или выходит за рамки задачи?

NikGurev (approved):

PR #1990, opened 2026-09-04
fix(textarea)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `textarea`, in the same shape as the 20.3.0 component reviews.

### What stays a plain accessor, and why

`KbqTextarea` implements `KbqFormFieldControl`, which declares `value`, `id`, `placeholder`, `required`, `disabled`, `focused`, `empty` and `errorState` as plain members. That interface is how the form field reads them, and it went through its own review in 20.3.0 — so those stay accessors. What moved are the four inputs the textarea owns.

### `canGrow` disagreed with what was bound

```ts
get canGrow(): boolean {
    return !this.maxRowLimitReached && this._canGrow;
}
```

It reported `false` once the textarea hit `maxRows`, even though the consumer had asked for growth. The folded value is what drives the resize handle, so it became an internal `growing` computed; `canGrow()` reports what was bound. Same shape as `KbqLink.tabIndex` in the earlier review.

### `freeRowsHeight` defaulted itself by writing its own input

```ts
this.freeRowsHeight = this.freeRowsHeight ?? this.lineHeight;
```

That is exactly why the automated migration skipped it. The fallback is a `computed` over the measured line height now, so binding `[freeRowsHeight]` later actually takes effect instead of being overwritten the next time `ngOnInit` runs — and it does run again, because the parent animation re-invokes it.

### `maxRows` and `freeRowsHeight` lied about being required

Both were declared `number` while an unbound textarea held `undefined`. `maxRowLimitReached` compared `rowsCount > undefined`, which is `false` — that accident is what made unlimited growth work. They report `number | undefined` now and the comparison is explicit.

### The row-limit class waited for someone else's change detection

`rowsCount` is written inside `runOutsideAngular`, so `[class.kbq-textarea_max-row-limit-reached]` only appeared on the next unrelated change detection pass. It is a signal now and the class follows it directly.

### Also

- The `parent.animationDone` subscription had no teardown; it is `takeUntilDestroyed()` now.
- The generated id comes from the CDK `_IdGenerator` instead of a module counter, so the shape changes from `kbq-textarea-1` to `kbq-textarea-a1`.
- `this.id = this.id` in the constructor — a trick to force the setter — is replaced by initialising the backing field.

## Migration

`textarea-signals` runs from `ng update @koobiq/components@20`. It rewrites the value-safe reads and reports the rest. `canGrow` is deliberately **not** rewritten: appending `()` would compile and hand back a different boolean at the row limit.

There is no template pass — `kbqTextarea` is an attribute on a native `<textarea>`, so a reference variable is not tied to an element name the schematic can match.

## Testing

- `textarea.component.spec.ts`: 24 → 27 tests. New coverage for the valueless `canGrow` attribute, `canGrow()` reporting the bound value, and the unbound `maxRows`/`freeRowsHeight` defaults.
- `textarea-signals/index.spec.ts`: 10 tests.
- Full `packages/components` (5164 tests) and `packages/schematics` (492 tests) suites pass.
- `check-api` is in sync.

The row-limit reactivity fix is not unit-testable here: jsdom reports no line height, so `rowsCount` never leaves 0.

BREAKING CHANGE


tools/public_api_guard/components/textarea.api.md
```diff
@@ -32,12 +32,11 @@ export const KBQ_TEXTAREA_VALUE_ACCESSOR: InjectionToken<{
 // @public (undocumented)
 export class KbqTextarea implements KbqFormFieldControl<any>, OnInit, OnChanges, OnDestroy, DoCheck, CanUpdateErrorState {
     constructor();
-    readonly autofilled: i0.Signal<boolean>;
-    get canGrow(): boolean;
-    set canGrow(value: boolean);
+    readonly autofilled: _angular_core.Signal<boolean>;
+    readonly canGrow: _angular_core.InputSignalWithTransform<boolean, unknown>;
```
lskramarov, line 36: **`check-public-api-any` baseline is stale — the API workflow will fail.**

The three transform-carrying signal inputs add three `unknown`s to the guarded surface (`canGrow` here, `freeRowsHeight` at line 51, `maxRows` at line 60), but `tools/check-public-api-any/baseline.json:45` still records `"textarea": 3`.

Running the exact algorithm from `tools/check-public-api-any/index.ts` over both goldens: base = 3, PR = 6. Signal-input transform types are not covered by the `ngAcceptInputType_` exemption in `isGenerated`, unlike the two `unknown`s the accessor inputs shed. So the gate prints `❌ The published type surface gained any / unknown: - textarea: 3 → 6` and exits 1.

Fix: `yarn run approve-public-api-any`, or narrow the write types (e.g. `input<number | undefined, number | string | null | undefined>`), which removes 2 of the 3.

packages/schematics/src/migrations/textarea-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+ * A `.ts` file is a textarea consumer if it names any of the exported symbols or imports the package. There
+ * is no element to look for: `kbqTextarea` is an attribute on a native `<textarea>`.
+ */
+function referencesTextarea(content: string): boolean {
+    return /\bKbqTextarea\w*\b/.test(content) || content.includes(TEXTAREA_PACKAGE);
+}
+
+export default function textareaSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov: **The migration rewrites nothing when it runs from `ng update`.**

`packages/components/package.json` points `ng-update.migrations` at `migrations.json`, whose `textarea-signals` entry carries no `"schema"` key. `@angular/cli/src/commands/update/cli.js:329` calls `executeSchematic(workflow, collection, name)` with no options, and `executeSchematic` defaults `options = {}` (line 233) — so no schema defaults are applied and `fix` is `undefined`.

`commit()` then takes the `else` branch and the consumer sees `[textarea-signals] would update src/app/app.ts (run with --fix to apply)` with zero files modified — contradicting the PR body and both registration descriptions ("Rewrites programmatic reads … to calls").

12 sibling migrations guard exactly this (`app-switcher-signals:413`, `navbar-signals-and-aria:596`, `list-tree-multiple-input:222`, `tag-slots:196`, …). `index.spec.ts` can't catch it: `run(fix = true)` always materialises the option, and it resolves through `collection.json`, which *does* declare the schema.

```suggestion
        const { project } = options;
        const fix = options.fix ?? true;
```

packages/schematics/src/migrations/textarea-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+            const rhs = parent.right;
+
+            edits.push({ start: node.getEnd(), end: rhs.getStart(sourceFile), text: '.set(' });
+            edits.push({ start: rhs.getEnd(), end: rhs.getEnd(), text: ')' });
+        }
+
+        return;
+    }
+
+    // Read (incl. optional chain `x?.compact`): append `()`.
```
lskramarov: **Compound and increment writes are rewritten into TypeScript that does not compile.**

`classifyAccess` recognises only `EqualsToken` as a write (line 146), so everything else falls through to this unconditional `()` append:

| input | output | diagnostic |
| --- | --- | --- |
| `textarea.freeRowsHeight += 4` | `textarea.freeRowsHeight() += 4` | TS2364 |
| `textarea.maxRows++` | `textarea.maxRows()++` | TS2357 |
| `textarea.maxRows ??= 3` | `textarea.maxRows() ??= 3` | TS2364 |
| `delete textarea.maxRows` | `delete textarea.maxRows()` | TS2703 |
| `[textarea.maxRows] = arr` | `[textarea.maxRows()] = arr` | TS2364 |

The safety net misses them too: `data.ts:48`'s pattern `\.\s*(?:freeRowsHeight|maxRows|canGrow)\s*=[^=]` requires `=` immediately after the member, so `+=`, `-=`, `??=`, `||=`, `&&=` and `++` all test `false`. The consumer gets a mangled file with no diagnostic — the opposite of the stated contract two lines up ("leave the write untouched (it becomes a compile error the consumer fixes by hand)"), which only holds for the plain `=` form.

Widening the guard to "parent is a `BinaryExpression` whose `left === node` and whose operator is an assignment token, or a `Prefix`/`PostfixUnaryExpression`, or a `DeleteExpression`, or a destructuring target" and returning early would cover all of them.

(Also on this line: `` `x?.compact` `` — `compact` is a `KbqAlert` member, carried over verbatim from `alert-signals/index.ts:158`.)

packages/schematics/src/migrations/textarea-signals/data.ts
```diff
@@ -0,0 +1,79 @@
...
+ * `required`, `disabled`, `focused`, `empty` and `errorState` as plain members — so those stay plain
+ * accessors. What moved are the inputs that belong to the textarea itself.
+ *
+ * - `textarea.canGrow` → a signal whose value changed: the getter folded in the row limit (warn)
+ * - `textarea.maxRows` / `freeRowsHeight` / `maxRowLimitReached` → calls (value unchanged — auto-fixed)
+ * - `textarea.freeRowsHeight = …` → the input is read-only now (warn)
+ */
+
+/** Members whose value is unchanged; a read must become a call. Auto-fixed. */
+export const SIGNAL_MEMBERS: readonly string[] = ['maxRows', 'freeRowsHeight', 'maxRowLimitReached'];
```
lskramarov: **`freeRowsHeight` is auto-fixed as "value unchanged", but its value did change.**

Before this PR, `ngOnInit` ran `this.freeRowsHeight = this.freeRowsHeight ?? this.lineHeight;`, so an unbound textarea's `freeRowsHeight` read back the *measured line height* (e.g. `20`) after the first microtask. Now the fallback lives in the private `resolvedFreeRowsHeight` computed (`textarea.component.ts:216`), so `freeRowsHeight()` returns `undefined` forever.

A consumer with

```ts
const gap = this.textarea.freeRowsHeight;
el.style.paddingBottom = gap + 'px';
```

is silently rewritten to `this.textarea.freeRowsHeight()` and starts emitting `"undefinedpx"` — no compile error, no warning. That is exactly the criterion this file uses to keep `canGrow` in `VALUE_CHANGED_MEMBERS` instead of auto-fixing it; `freeRowsHeight` belongs there for the same reason.

The prose repeats the same claim — data.ts:67, `README.md:47`, `migration.en.md:1128` all say "an unbound textarea held `undefined`", which is true of `maxRows` but false of `freeRowsHeight` after init.

packages/components/textarea/textarea.component.ts
```diff
@@ -294,13 +292,18 @@ export class KbqTextarea
...
 
-    /** Grow textarea height to avoid vertical scroll  */
-    grow = () => {
-        if (!this.isBrowser || !this._canGrow) return;
+    /**
+     * Grow textarea height to avoid vertical scroll.
+     *
+     * @docs-private
+     */
+    grow(): void {
```
lskramarov, line 312: **`grow` went from a bound arrow property to a prototype method — a silent runtime break with no migration coverage.**

The golden records it (`grow: () => void;` → `grow(): void;`), but both forms type as `() => void`, so a consumer holding it detached still compiles:

```ts
el.addEventListener('input', textarea.grow);
this.resize$.subscribe(textarea.grow);
setTimeout(textarea.grow, 0);   // ← the idiom this very file used at its old line 259
```

All of those now run with `this === undefined` (class bodies are strict mode) and throw `TypeError: Cannot read properties of undefined (reading 'isBrowser')` on line 306.

`grow` appears in none of `data.ts`'s `SIGNAL_MEMBERS` / `VALUE_CHANGED_MEMBERS` / `warnPatterns` / `SUMMARY`, nor in the §18 Textarea subsections of `migration.{en,ru}.md`. AGENTS.md §Schematics: *"A breaking change to a component ships with a migration there, or at least a warning that points at the manual steps."* A `warnPatterns` entry matching `\.\s*grow\b(?!\s*\()` on a `KbqTextarea`-anchored file would cover it.

packages/schematics/src/migrations/textarea-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+        const projectDefinition = await setupOptions(project, tree);
+        const root = projectDefinition?.root ?? '';
+        const rootDir = root ? tree.getDir(root as Path) : tree.root;
+
+        const tsPaths: string[] = [];
+
+        rootDir.visit((filePath) => {
+            if (filePath.includes('node_modules') || filePath.includes('/dist/')) return;
+
+            if (filePath.endsWith(TS_EXT)) tsPaths.push(filePath);
```
lskramarov, line 433: **No template pass, and the stated reason for skipping it is false.**

`README.md:34` justifies it as: *"There is no template pass: `kbqTextarea` is an attribute on a native `<textarea>`, so a reference variable is not tied to an element name the schematic can match."*

But the directive declares `exportAs: 'kbqTextarea'` (`textarea.component.ts:58`), and the siblings match reference variables by **exportAs value on any element**, never by element name — `button-toggle-signals-and-aria/index.ts:403` (`attr.value !== GROUP_EXPORT_AS`), `app-switcher-signals/index.ts:249`, `navbar-signals-and-aria/index.ts:361`.

So this is left unmigrated and, unlike the TypeScript case, it does not fail to compile:

```html
<textarea kbqTextarea #t="kbqTextarea" [maxRows]="5"></textarea>
<div [class.at-limit]="t.maxRowLimitReached">…</div>
```

`t.maxRowLimitReached` is now a `Signal<boolean>` function object — permanently truthy — so `at-limit` is applied from the first render, silently inverting the binding. `{{ t.maxRows }}` likewise renders a function.

lskramarov (commented):
Automated review (`/code-review max`) of the textarea component review. 15 findings: 10 correctness, 3 test-coverage, 1 simplification, 1 reuse — posted inline.

One of them is CI-blocking (`check-public-api-any` baseline), and one makes the new schematic a no-op under `ng update`.

Verified locally at `54f91dd`: `prettier`, `eslint --max-warnings=0` and `cspell` pass on every changed file, and both new suites are green (`textarea.component.spec.ts` 27/27, `textarea-signals` 10/10).

packages/components/textarea/textarea.component.ts
```diff
@@ -31,15 +34,17 @@ import { asapScheduler, observeOn, Subject } from 'rxjs';
 
 export const KBQ_TEXTAREA_VALUE_ACCESSOR = new InjectionToken<{ value: any }>('KBQ_TEXTAREA_VALUE_ACCESSOR');
 
-let nextUniqueId = 0;
+/** Coerces an optional numeric input, keeping `undefined` distinguishable from `0`. */
+const optionalNumberAttribute = (value: unknown): number | undefined =>
```
lskramarov: **`optionalNumberAttribute` leaks `NaN`, so the declared `number | undefined` is not what it returns.**

Angular's `numberAttribute` defaults `fallbackValue = NaN`, and `NaN` is not nullish, so `??` downstream does not catch it. With `<textarea kbqTextarea freeRowsHeight>` (bare attribute → `''`):

- `freeRowsHeight()` → `NaN`
- `resolvedFreeRowsHeight()` (line 216) → `NaN ?? lineHeight()` → `NaN`
- `height` (line 321) → `NaN`
- `coerceCssPixelValue(NaN)` → `min-height: "NaNpx"`, which the CSSOM drops

The textarea stops growing and shows a scrollbar instead, while `rowsCount` is pinned to `0` by the `lineHeight > 0` guard so `maxRowLimitReached()` never fires either. Same for a bare `maxRows` or any binding that evaluates to a non-number — which the new `unknown` write type now accepts silently where `input<number>` was a template type error.

The repo already has the idiom: `packages/components/core/overlay/panel-width.ts:83` keeps `numberAttribute` and guards on `Number.isFinite`. A `Number.isFinite(n) ? n : undefined` here would close it.

Minor, same mechanism: because `maxRowLimitReached` already folds in `maxRows !== undefined`, the second conjunct at line 330 can never be `false` — swapping the operands (`maxRows !== undefined && this.maxRowLimitReached()`) makes it read as the narrowing it actually is.

docs/guides/migration.en.md
```diff
@@ -1113,6 +1113,28 @@ A `<kbq-split-button>` with no projected button no longer throws outside dev mod
...
+| `.canGrow`                                              | `canGrow()`, and expect what was bound — not `false` at the row limit |
+| `.canGrow = …` / `.maxRows = …` / `.freeRowsHeight = …` | Bind them in the template; the inputs are read-only                   |
+
+**`maxRows` and `freeRowsHeight` report `number | undefined`.** Both were declared non-nullable while an unbound textarea held `undefined`, and `maxRowLimitReached` compared against it — `rowsCount > undefined` is false, which is why unlimited growth worked at all.
+
+**`freeRowsHeight` no longer writes itself.** It defaulted to the measured line height by assigning its own input in `ngOnInit`; the fallback is a computed now, so binding it later actually takes effect instead of being overwritten on the next init.
+
+**The `kbq-textarea_max-row-limit-reached` class follows the row count directly.** It is derived from a signal written inside `runOutsideAngular`, so the class used to wait for an unrelated change detection pass to appear.
+
+**Generated ids changed shape**, from `kbq-textarea-1` to `kbq-textarea-a1`.
```
lskramarov: **The generated-id change does not happen for consumers — `kbq-textarea-a1` is a TestBed artifact.**

CDK's `_IdGenerator.getId()` is:

```js
```

with the CDK's own comment *"Omit the app ID if it's the default `ng`"*, and Angular's `APP_ID` defaults to `'ng'`. The `a` only appears because `@angular/platform-browser/testing` provides `{provide: APP_ID, useValue: 'a'}`.

The per-prefix counter also still starts at `0`, exactly like the removed `let nextUniqueId = 0` — so a real app gets `kbq-textarea-0` before *and* after. Nothing changed.

A consumer who greps their snapshots for `kbq-textarea-a…` on the strength of this note finds nothing. And `data.ts:76`'s rationale — *"the app id is part of the prefix now, which keeps two Angular apps on one page from colliding"* — is false at the default `APP_ID`, which is precisely the case where the CDK strips it.

Mirrored at `migration.ru.md:1138`, `README.md:56` and `data.ts:76`.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **The rewritten intro contradicts the subsections it introduces.**

> "The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; **the second is the one each subsection below belongs to**."

The `####` subsections actually present in §18 are: Popover, Search expandable, Split button, Textarea, Title, Toast, Tooltip. Six of those seven are named in the *first*-wave list in the same sentence, so the guide now assigns six already-shipped migrations to the wrong release wave. Only the Textarea subsection is new.

Mirrored verbatim at `migration.ru.md:1042`. Naming the second wave explicitly ("the second covered textarea") avoids the contradiction and survives the next component review.

packages/components/textarea/textarea.component.ts
```diff
@@ -204,23 +194,39 @@ export class KbqTextarea
...
+    readonly maxRowLimitReached = computed(() => {
+        const maxRows = this.maxRows();
 
-    protected uid = `kbq-textarea-${nextUniqueId++}`;
+        return maxRows !== undefined && this.rowsCount() > maxRows;
+    });
+
+    /**
+     * Whether the textarea still grows with its content. It stops once `maxRows` is reached, which is
+     * when the native resize handle takes over.
```
lskramarov: **"the native resize handle takes over" — the stylesheet says otherwise.**

At the row limit with `canGrow` true, `growing()` is `false`, so the element carries **both** classes. In `textarea.scss`:

```scss
&.kbq-textarea-resizable        { resize: vertical; }   // line 29
&.kbq-textarea_max-row-limit-reached { resize: unset; }  // line 38
```

Equal specificity, the second wins by source order, and `resize: unset` on a non-inherited property resolves to `none`. So no handle appears — the fold's only surviving effect is dropping `overflow-y: hidden` (which is the behaviour you want, just not the one described).

The same claim is in both migration guides ("The folded value drives the resize handle") and `README.md:22`. Worth correcting in all four places, since the next person debugging this will go looking for a handle and conclude the class binding is broken.

packages/schematics/src/migrations/textarea-signals/index.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+            'class Demo {\n' +
+            '    read(textarea: KbqTextarea) {\n' +
+            '        return textarea.canGrow;\n' +
+            '    }\n' +
+            '}\n';
+
+        appTree.overwrite(ts, source);
+
+        expect((await run()).readText(ts)).toContain('return textarea.canGrow;');
+        expect(messages.join('\n')).toContain('maxRows');
```
lskramarov: **This assertion is vacuous — `SUMMARY` always prints `maxRows`.**

`SUMMARY[0]` (`data.ts:67`) begins ``​`maxRows` and `freeRowsHeight` report `number | undefined`​``, and `index.ts:299` prints `SUMMARY` unconditionally whenever `consumers > 0` — which this fixture guarantees, since it imports `@koobiq/components/textarea`.

Deleting the `warnReceiverMembers(...)` call at `index.ts:291` leaves this test green, so the migration's headline warning has no real guard. The neighbouring test at line 165 gets it right (`toContain('read-only signal inputs')` is unique to `warnPatterns[0]`).

```suggestion
        expect(messages.join('\n')).toContain('read-only InputSignal');
```

packages/schematics/src/migrations/textarea-signals/index.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+                '    read() {\n' +
+                '        return this.textarea.maxRows;\n' +
+                '    }\n' +
+                '}\n'
+        );
+
+        expect((await run()).readText(ts)).toContain('return this.textarea.maxRows();');
+    });
+
+    it('leaves reads on a receiver of an unrelated type alone', async () => {
```
lskramarov, line 93: **This test never reaches the logic it names.**

The fixture declares only `read(other: Other)` — there is no `KbqTextarea`-annotated receiver anywhere in it. So `collectReceivers` returns `[]` and `migrateTsExpressions` returns at `index.ts:211` (`if (receivers.length === 0) return content;`) before any receiver discrimination runs.

Mutation-verified: replacing the body of `inReceiverScope` with an always-`true` expression keeps all 10 tests in this file green — i.e. the type-discrimination that decides which reads get rewritten has zero coverage.

Adding a `KbqTextarea`-typed receiver to the same file alongside the `Other`-typed one (which is what the `alert-signals` spec this was copied from does) makes the test bite.

packages/components/textarea/textarea.component.spec.ts
```diff
@@ -308,6 +308,40 @@ describe('KbqTextarea', () => {
...
+
+            fixture.detectChanges();
+
+            const textarea = fixture.debugElement.query(By.directive(KbqTextarea)).injector.get(KbqTextarea);
+
+            expect(textarea.canGrow()).toBe(true);
+            expect(getTextareaElement(fixture).classList.contains('kbq-textarea-resizable')).toBe(false);
+        });
+
+        it('should report the bound canGrow rather than folding in the row limit', () => {
```
lskramarov: **The regression test for the headline change passes identically against the old code.**

It asserts `canGrow()` is `true` *together with* `maxRowLimitReached()` being `false` — conditions under which the old getter `!this.maxRowLimitReached && this._canGrow` also returned `true`. The distinguishing case (limit reached → old getter `false`, new input `true`) is never reached.

It can't be reached here at all: jsdom returns `line-height: normal`, so `parseInt(…)` is `NaN`, the `lineHeight > 0` guard at line 325 is `false`, and `rowsCount` is pinned at `0` for every unit test. That also leaves the `maxRows * lineHeight` clamp (line 330), the `kbq-textarea_max-row-limit-reached` class, and the PR's "the class follows the row count directly" claim with no unit coverage — and makes the `maxRowLimitReached()` assertions at lines 309, 331 and 343 trivially true.

Stubbing `KBQ_WINDOW.getComputedStyle` to return a real line height makes the limit reachable and turns all four assertions into real ones.

packages/schematics/src/migrations/textarea-signals/data.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Signal members that are writable via `.set(...)`. Every migrated member is an `input()` or a read-only
+ * `computed`, so this is empty — a programmatic write is left untouched and becomes a compile error.
+ */
+export const WRITABLE_MEMBERS: ReadonlySet<string> = new Set<string>();
```
lskramarov: **Empty `WRITABLE_MEMBERS` makes the whole `.set(` path dead code.**

Every member in `SIGNAL_MEMBERS` is an `input()` or a read-only `computed`, so `WRITABLE_MEMBERS.has(...)` at `index.ts:148` is always `false`. That ships ~10 lines of unreachable rewrite code plus this 5-line comment explaining why the set is empty — and `index.ts:139` (`parent.name.text === 'set'`) guards against output this migration can never produce.

Simpler: drop `WRITABLE_MEMBERS` and its import, collapse the write branch in `classifyAccess` to a bare `return` with the comment that's already there, and delete the `.set` idempotency check.

packages/schematics/src/migrations/textarea-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+/** A receiver whose static type is a textarea, valid within `[start, end]` of the source. */
+interface Receiver {
+    /** Source text of the receiver expression, e.g. `textarea` or `this.textarea`. */
+    text: string;
+    start: number;
+    end: number;
+}
+
+/** Applies text-span edits to `content`, right-to-left, so earlier edits don't shift later offsets. */
+function applyEdits(content: string, edits: Edit[]): string {
```
lskramarov, line 43: **This block is the 10th copy — and this PR is the one that made it fully generic.**

`applyEdits` is declared in 10 migration folders (`alert-signals`, `app-switcher-signals`, `button-toggle-signals-and-aria`, `button-truncation`, `filter-bar-signals`, `form-field-signals`, `list-tree-multiple-input`, `navbar-signals-and-aria`, `tag-slots`, `textarea-signals`) — 8 byte-identical, 2 already drifted. `isFunctionLike`, `findAncestor` and `FIELD_MODIFIERS` are byte-identical in 7; `inReceiverScope` in 4; `collectAccessEdits` matches `alert-signals`' verbatim.

This copy generalized the last two component-specific pieces — `isAlertType(type)` → `isTypeReference(type, typeName)` and `collectReceivers(sourceFile)` → `collectReceivers(sourceFile, typeName)` — so nothing in lines 21–183 is textarea-specific any more, and it still landed in the leaf folder. Consequence: the compound-assignment bug above has to be fixed in 10 places, and migration #11 starts from whichever copy gets pasted.

The shared home already exists and is already imported by these migrations: `packages/schematics/src/utils/typescript.ts` (`forEachClass`, `canMigrateFile`, `collectInlineTemplateRanges`). Extracting there would also drop the per-file double parse — `index.ts:208` and `:220` each call `ts.createSourceFile` **and** `collectReceivers` on the same content, and only ~4 of ~33 matched files in a workspace this size actually carry a `: KbqTextarea` annotation.

packages/schematics/src/migrations/textarea-signals/index.ts
artembelik: сделал по вашему списку, только из префиксных записью считаю `++`/`--`: `!textarea.maxRowLimitReached` тоже префиксный, но это чтение

packages/components/textarea/textarea.component.ts
artembelik: сам метод оставил прототипным, добавил предупреждение по вашему шаблону

tools/public_api_guard/components/textarea.api.md
artembelik: зафиксировал новое значение, а не сузил типы: второй параметр `InputSignalWithTransform` выводит Angular из сигнатуры `numberAttribute`, сужение отвергало бы валидные привязки

docs/guides/migration.en.md
artembelik: вступление приехало со слиянием `main` - там абзац уже переписан без деления на волны

packages/components/textarea/textarea.component.ts
```diff
@@ -294,13 +299,18 @@ export class KbqTextarea
...
 
-    /** Grow textarea height to avoid vertical scroll  */
-    grow = () => {
-        if (!this.isBrowser || !this._canGrow) return;
+    /**
+     * Grow textarea height to avoid vertical scroll.
+     *
+     * @docs-private
+     */
+    grow(): void {
```
NikGurev: Насколько я понимаю, снаружи он не нужен, можно сделать `protected`

```suggestion
    protected grow(): void {
```

NikGurev (approved):

lskramarov (approved):

packages/components/textarea/textarea.component.ts
artembelik: Сделал `grow` защищённым: вызов и отвязанная ссылка теперь не компилируются, схематик сообщает об обоих и предлагает `stateChanges.next()` для изменения раскладки, которого textarea не видит; спеки обращаются к нему через `textarea['grow']()`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1992, opened 2026-09-04
fix(timepicker)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `timepicker`, in the same shape as the 20.3.0 component reviews.

### What stays a plain accessor, and why

`KbqTimepicker` implements `KbqFormFieldControl`, which declares `value`, `id`, `placeholder`, `required`, `disabled`, `focused`, `empty` and `errorState` as plain members. Those stay accessors — the interface is how the form field reads them. The six of them carried `// TODO: Skipped for migration because:` comments from the automated tooling; they say the actual reason now.

### `min` and `max` reported something other than what was bound

```ts
set min(value: D | null) {
    this._min = this.getValidDateOrNull(this.dateAdapter.deserialize(value));
    this.validatorOnChange();
}
```

A value the date adapter could not read came back as `null` from the getter. They report what was bound now; the parsed values are internal computeds that still drive the validators, re-run from an effect.

### `kbqValidationTooltip` leaked a subscription per binding

```ts
set kbqValidationTooltip(tooltip: KbqTooltipTrigger) {
    …
    this.incorrectInput.subscribe(() => { … });   // never unsubscribed
}
```

Re-binding the input stacked another subscription, and the last one outlived the directive. It is a signal input driven by an effect with an `onCleanup` teardown.

### `format`'s side effects were already covered

Its setter clamped the value, refreshed the placeholder and re-rendered the view. The clamp is the input's transform; the other two are subsumed by the locale effect that already re-read `timeFormatPlaceholder` — which reads `format()`, so it re-runs on a format change too.

That effect also **stopped returning early on a consumer-provided placeholder**. It used to skip the value reformat along with the placeholder update, so a locale change left the rendered time in the old locale whenever the consumer had set their own placeholder.

The existing "Display time corresponding to timeformat" tests — including the unsupported-format fallback and the format-change re-render — pass unchanged, which is the evidence that moving the side effects preserved behavior.

### Also

- The generated id comes from the CDK `_IdGenerator` instead of a module counter (`kbq-timepicker-1` → `kbq-timepicker-a1`), and `this.id = this.id` in the constructor is replaced by initialising the backing field.
- `disabled` and `required` declare `booleanAttribute` instead of calling `coerceBooleanProperty` inside their setters.

## Migration

`timepicker-signals` runs from `ng update @koobiq/components@20`. It rewrites `format` reads and reports the rest. `min` and `max` are deliberately **not** rewritten: appending `()` would compile and hand back a different value for anything the adapter could not parse.

There is no template pass — `kbqTimepicker` is an attribute on a native `<input>`.

## Testing

- `timepicker.spec.ts`: 64 → 67 tests. New coverage for the format clamp through the transform, `min`/`max` reporting the bound value, and the validators re-running when `min` changes.
- `timepicker-signals/index.spec.ts`: 10 tests.
- Full `packages/components` (5164 tests) and `packages/schematics` (492 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


packages/components/timepicker/timepicker.directive.ts
```diff
@@ -111,14 +114,16 @@ export const kbqTimepickerLocaleConfigurationProvider = (
...
-let uniqueComponentIdSuffix: number = 0;
-
 const shortFormatSize: number = 5;
 const fullFormatSize: number = 8;
 
 /** Maximum number of digits in a single time part */
 const timePartLength: number = 2;
 
+/** Coerces a time format, falling back to the default for anything the enum does not name. */
+const timeFormatAttribute = (value: unknown): TimeFormats =>
```
lskramarov, line 124: **`check-public-api-any` fails: timepicker 4 → 5.**

Typing the transform's parameter `unknown` emits `readonly format: _angular_core.InputSignalWithTransform<TimeFormats, unknown>;` into the golden file. That is a hand-written line — the ratchet only skips `/ɵ|ngAcceptInputType_/` — so the count goes from 4 to 5 while `tools/check-public-api-any/baseline.json:47` still records `"timepicker": 4`, and that file is not in this PR.

I counted it with the tool's own filter against both revisions of `tools/public_api_guard/components/timepicker.api.md`: base 4 (`KBQ_TIMEPICKER_VALIDATORS`, `KBQ_TIMEPICKER_VALUE_ACCESSOR`, `get ngControl()`, `onPaste($event)`), PR 5. The API workflow runs `check-api` then `check-public-api-any`, which exits 1 with `timepicker: 4 → 5` — and its own advice is *"Narrow the new members instead"*, so approving the baseline is the wrong fix here.

Narrowing also restores a compile-time check the old `set format(formatValue: TimeFormats)` had: with `unknown` as the write type, `[format]="42"` and `[format]="someUnrelatedObject"` now type-check under `strictTemplates` and silently clamp to `HH:mm`. The sibling does it narrowly — `alert.api.md:15` is `InputSignalWithTransform<"error"|"warning"|"success"|"info", "error"|…|null|undefined>`.

packages/components/timepicker/timepicker.directive.ts
```diff
@@ -457,17 +396,48 @@ export class KbqTimepicker<D>
...
+                // Assigned through the private field so the setter does not mark it consumer-provided.
+                this._placeholder = placeholder;
+            }
 
-            // Assigned through the private field so that the setter does not mark it as consumer-provided.
-            this._placeholder = placeholder;
-            // Re-assigning the value re-runs it through the date adapter, which formats on the new locale.
+            // Re-assigning the value re-runs it through the date adapter, which formats on the new locale
+            // and the new format.
             this.value = this._value;
```
lskramarov, line 409: **A `[format]` change now erases partially typed text — the `if (this.value)` guard was dropped.**

`timeFormatPlaceholder` (line 348) reads `this.format()`, so this effect is now tracked on `format`; the old code read a plain field and was not. The deleted setter ended with:

```ts
if (this.value) {
    this.updateView();
}
```

The effect has no such guard, and `updateView()` → `getTimeStringFromDate(null, …)` returns `''` (line 922) → `setViewValue('')`.

Measured, PR vs base, on `<input kbqTimepicker [formControl]="control" [format]="timeFormat">`: type `1` (incomplete, so `_value` stays null), then flip the format —

| | base | this PR |
|---|---|---|
| input value after typing | `"1"` | `"1"` |
| after `[format]` change + CD | `"1"` | `""` |

This is directly reachable from the shipped docs example `packages/docs-examples/components/timepicker/timepicker-variations/timepicker-variations-example.html`, which puts a format radio group next to the input, and from `packages/components-dev/timepicker/template.html`.

The same line also re-derives `lastValueValid` from `_value` (setter, line 283), so a pending `kbqTimepickerParse` error silently clears at the same moment.

`if (this._value) this.value = this._value;` keeps the intended re-format for complete values (verified: `10:05` → `10:05:07` still works).

packages/components/timepicker/timepicker.directive.ts
```diff
@@ -457,17 +396,48 @@ export class KbqTimepicker<D>
...
         });
 
+        // `min` and `max` feed the validators, which Angular only re-runs when it is told to.
+        effect(() => {
+            this.minDate();
+            this.maxDate();
+
+            this.validatorOnChange();
```
lskramarov, line 417: **This fires for every timepicker on first CD, even with no `[min]`/`[max]` bound — and it runs inside the effect's reactive context.**

An `effect()` always runs once. The old `set min`/`set max` only ran when the input was actually bound, so a timepicker with no bounds never called `validatorOnChange()`. Effects run in `refreshView` *after* `FormControlDirective.ngOnChanges`, so by then `validatorOnChange` is the real `() => control.updateValueAndValidity()`, which emits `valueChanges` + `statusChanges` unconditionally and recurses into every ancestor.

Measured on `<input kbqTimepicker [formControl]="control">` inside a `FormGroup`, **no min/max bound**, after the first `detectChanges()`:

| | base | this PR |
|---|---|---|
| `control.valueChanges` | 0 | 1 |
| `control.statusChanges` | 0 | 1 |
| `group.valueChanges` | 0 | 1 |

Any autosave / dirty-guard / `valueChanges` subscriber misfires on form open. `packages/components/time-range/time-range-editor.html:32` mounts exactly this shape.

Follow-on: `setControl()` (line 1043) subscribes `control.valueChanges → (value) => (this._value = value)`, so the spurious emission overwrites the *parsed* `_value` with the **raw** control value. With a form patched from an API (`new FormControl('1970-01-01T12:34:00.000')`), `timepicker.value` is a string after the first CD, and because `onBlur()` (line 484) compares `viewValue !== getTimeStringFromDate(this.value, format())`, a plain focus + blur with no typing rewrites the field.

Two more consequences of the same move:

**It runs inside the effect's tracking scope.** `updateValueAndValidity()` synchronously re-runs every validator in the form tree, so any signal those validators read silently becomes a dependency of this effect. With two migrated controls in one form, each one's `minDate()`/`maxDate()` lands in the other's dependency set. Wrap it: `untracked(() => this.validatorOnChange())`.

**It moved past the host template in change-detection order.** `refreshView` runs `executeTemplate` → pre-order hooks → `runEffectsInView`; the old setters ran inside `executeTemplate` at the input's own binding site, so a validity flip landed *before* the rest of the template was read. `packages/components/filter-bar/pipes/pipe-datetime.html:89` binds `[max]="startMax"` on a timepicker, where `startMax` is `min(max, end)` (`pipe-date-base.ts:255`) and so changes whenever the user edits the end time — while lines 134/138 of the same template read `formGroup.errors?.kbqDateIntervalMin` / `kbqDateIntervalMax`. Those `@else if` branches are now evaluated before the effect flips `formGroup.errors`, so `checkNoChanges` re-reads a changed condition: `ExpressionChangedAfterItHasBeenCheckedError` in dev, a one-pass-late hint in production.

Guarding the effect (skip the run when neither parsed bound actually changed) removes the first problem; `untracked` removes the second; the third needs the revalidation to stay synchronous with the write.

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+ * A `.ts` file is a timepicker consumer if it names any of the exported symbols or imports the package.
+ * There is no element to look for: `kbqTimepicker` is an attribute on a native `<input>`.
+ */
+function referencesTimepicker(content: string): boolean {
+    return /\bKbqTimepicker\w*\b/.test(content) || content.includes(TIMEPICKER_PACKAGE);
+}
+
+export default function timepickerSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov, line 255: **`fix` is `undefined` under `ng update`, so the migration never writes anything.**

`migrations.json` registers `timepicker-signals` with only `version`/`description`/`factory` — no `"schema"` (`grep -c '"schema"' migrations.json` → 0 for every entry) — and the CLI invokes migrations with no options, so `schema.json`'s `"fix": {"default": true}` never reaches the factory.

`commit()` then takes the `else` branch for every consumer file and logs `would update <path> (run with --fix to apply)`, so the one thing this migration auto-fixes (`timepicker.format` → `format()`) never runs in the context it was built for. The user's reads are left as-is and become compile errors after the upgrade.

`index.spec.ts` cannot catch it because `run()` always passes `fix` explicitly.

The current sibling documents this exact trap — `packages/schematics/src/migrations/checkbox-signals/index.ts:843`:

```ts
// `ng update` invokes migrations with no options at all, and migrations.json declares no schema, so
// the schema default never reaches us — applying the fix is the intended behaviour there.
```

```suggestion
        const { project } = options;
        const fix = options.fix ?? true;
```

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+    // Already migrated: a call, or a `.set(...)` write — leave alone (idempotent).
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.member = RHS`. Every KbqTimepicker signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
```
lskramarov, line 146: **Only `=` counts as a write, so every other write form is rewritten into TypeScript that does not parse.**

Everything that is not a bare `=` falls through to the read branch at line 159 and gets `()` appended:

| input | output after the migration | diagnostic |
|---|---|---|
| `timepicker.format ??= f;` | `timepicker.format() ??= f;` | TS2364 |
| `timepicker.format += x;` / `\|\|=` | `timepicker.format() += x;` | TS2364 |
| `delete timepicker.format;` | `delete timepicker.format();` | TS2703 |
| `timepicker.format++;` | `timepicker.format()++;` | TS2357 |
| `({ format: timepicker.format } = src);` | `({ format: timepicker.format() } = src);` | TS2364 |

Worse, `warnPatterns[0]` in `data.ts` does not match `??=`/`+=` either, so the consumer gets neither the warning nor a correct rewrite — just corrupted source, reported as a successful migration.

`checkbox-signals/index.ts` handles this with an `ASSIGNMENT_OPERATORS` set plus explicit `PostfixUnaryExpression` / `PrefixUnaryExpression` / `DeleteExpression` / destructuring-target branches, and states the reason: *"appending `()` would produce unparseable TypeScript instead of the read-only error the consumer should see."*

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+    let result = content;
+
+    for (const { start, end, text } of sorted) {
+        result = result.slice(0, start) + text + result.slice(end);
+    }
+
+    return result;
+}
+
+const isFunctionLike = (node: ts.Node): boolean =>
```
lskramarov, line 47: **Hand-rolled `isFunctionLike` misses type-position parameters, so a receiver's scope silently widens to the whole file.**

This 7-way OR omits `MethodSignature`, `FunctionTypeNode`, `CallSignature`, `ConstructSignature` and `IndexSignature`. A `KbqTimepicker` parameter in a *type* position therefore finds no ancestor and hits the `?? sourceFile` fallback at line 92 — the receiver name becomes valid for the entire file:

```ts
import { KbqTimepicker } from '@koobiq/components/timepicker';
type Attach = (timepicker: KbqTimepicker) => void;   // receiver scope = whole file

function unrelated() {
    const timepicker = { format: 'HH:mm:ss' };
    return timepicker.format;   // → timepicker.format()  ✗ TypeError at runtime
}
```

`ts.isFunctionLike` is public TypeScript API and six siblings already use it (`autocomplete-signals`, `badge-signals`, `checkbox-signals`, `loader-overlay-signals`, `markdown-signals`, `progress-spinner-signals`); `checkbox-signals` names this exact bug in its own comment. Swapping to `ts.findAncestor(node.parent, ts.isFunctionLike)` removes 19 lines and fixes it.

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
+    const start = node.getStart(sourceFile);
+    const end = node.getEnd();
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
```
lskramarov, line 130: **Receiver matching is raw source-text comparison, so the most common `@ViewChild` idiom is neither rewritten nor reported.**

`node.expression.getText(sourceFile)` is compared verbatim against the collected receiver text, so anything that decorates the receiver misses:

```ts
@ViewChild(KbqTimepicker) timepicker!: KbqTimepicker;
read() { return this.timepicker!.format; }   // receiver text is "this.timepicker!" → never matches
```

Same for `(timepicker).format`, `this . timepicker`, an aliased import (`import { KbqTimepicker as TP }`), and receivers typed through `inject(KbqTimepicker)` / `viewChild(KbqTimepicker)` rather than an explicit annotation. `warnReceiverMembers` uses the same predicate, and the coarse `data.ts` regexes only cover writes and query declarations — so the consumer gets a clean migration report and a TS2322 on their next build.

`badge-signals/index.ts:332` normalizes this (*"`this.badge!.compact` and `(badge).compact` name the same receiver"*); `checkbox-signals` adds `TYPING_FACTORIES` for `inject`/`viewChild`/`contentChild`, `localTypeNames` for aliases, and — the important part — an `UNRESOLVED_RECEIVER_MESSAGE` so misses are *reported* instead of silent.

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+        const projectDefinition = await setupOptions(project, tree);
+        const root = projectDefinition?.root ?? '';
+        const rootDir = root ? tree.getDir(root as Path) : tree.root;
+
+        const tsPaths: string[] = [];
+
+        rootDir.visit((filePath) => {
+            if (filePath.includes('node_modules') || filePath.includes('/dist/')) return;
+
+            if (filePath.endsWith(TS_EXT)) tsPaths.push(filePath);
```
lskramarov, line 265: **No template pass, and the stated reason answers the wrong question.**

Only `.ts` is collected, and `README.md:34` justifies it with *"There is no element to look for: `kbqTimepicker` is an attribute on a native `<input>`."* A template pass is not about the element — it is about the template reference variable. `KbqTimepicker` declares `exportAs: 'kbqTimepicker'` (`timepicker.directive.ts:149`), so this is a supported shape:

```html
<input kbqTimepicker #tp="kbqTimepicker" [format]="fmt" />
{{ tp.format }}        <!-- stringifies the InputSignal function -->
@if (tp.min) { … }     <!-- a function is always truthy: the branch never turns off -->
```

Interpolation imposes no assignability constraint, so AOT does not catch it either, and `.html` files are never read — nothing is rewritten and nothing is warned.

`app-switcher-signals/index.ts:252` solves exactly this, element-agnostically, by matching the export name instead of the tag:

```ts
if (typeof attr.name !== 'string' || attr.value !== TRIGGER_EXPORT_AS) continue;
if (attr.name.startsWith('#')) this.refs.add(attr.name.slice(1));
```

This migration was cloned from `alert-signals`, whose collector is hardcoded to `element.name === ALERT_ELEMENT`, and the gap was rationalised rather than resolved with the sibling that already handles attribute-selector directives.

packages/components/timepicker/timepicker.directive.ts
```diff
@@ -457,17 +396,48 @@ export class KbqTimepicker<D>
...
+
+            const subscription = this.incorrectInput.subscribe(() => {
+                if (tooltip.isOpen) return;
+
+                tooltip.show();
+
+                setTimeout(() => tooltip.hide(), validationTooltipHideDelay);
+            });
+
+            onCleanup(() => subscription.unsubscribe());
```
lskramarov, line 438: **The teardown restores half of what the effect set up, so an unbound tooltip is left permanently inert.**

Lines 425–428 mutate a foreign directive — `enterDelay`, `trigger = 'manual'`, `initListeners()` — and `onCleanup` only unsubscribes. `initListeners()` starts with `clearListeners()` (`core/pop-up/pop-up-trigger.ts:561`) and `'manual'` matches none of Click/Hover/Focus/Keydown, so every hover/focus handler the trigger bound in its own `ngOnInit` is removed and none is bound back. Nothing in `KbqTooltipTrigger.ngOnChanges` restores them.

So in this PR's own `TimepickerWithValidationTooltip` fixture, after `useSecondTooltip = true` the `<span #first kbqTooltip="first">` keeps `trigger: 'manual'`, `enterDelay: 10` and zero DOM listeners forever — hovering it never opens its tooltip again. Same on destroy when the anchor outlives the timepicker (timepicker inside an `@if`, anchor outside).

The new spec cannot see this. `should stop driving a validation tooltip once it is unbound` never unbinds anything — `[kbqValidationTooltip]="useSecondTooltip ? second : first"` is total, so it only ever *re-binds*, and the effect's `if (!tooltip) return;` branch is never entered. The one state the test's name promises to cover is the one left broken.

Snapshot and restore `trigger`/`enterDelay` in `onCleanup` — or better, give `KbqTooltipTrigger` a narrow API for this so the timepicker never touches those fields.

Also: **line 428 is redundant** — `set trigger` already ends with `this.initListeners()` (`tooltip.component.ts:445`). It was carried over verbatim from the old setter.

lskramarov (commented):
## Code review — `max` effort

Ten finder angles plus a gap sweep, with the behavioural findings measured against `main` in a throwaway checkout.

Two things fail or regress outside the component:

- **`check-public-api-any` will fail the API job** — `timepicker` goes 4 → 5 and `tools/check-public-api-any/baseline.json` is not in this PR.
- **`timepicker-signals` never applies its fix under `ng update`** — `fix` has no `?? true` default and `migrations.json` declares no schema.

Two measured behaviour regressions in the directive, both traceable to a guard that moved into an `effect()`:

| scenario | base | this PR |
| --- | --- | --- |
| `[format]` change while `1` is typed | input keeps `1` | input is cleared |
| first CD, no `[min]`/`[max]` bound | 0 `valueChanges` / `statusChanges` | 1 each, propagated to the parent `FormGroup` |

The remaining schematic findings share one root cause, which is the last comment: this migration was cloned from `alert-signals` rather than `checkbox-signals`, which already fixes all of them.

Two smaller things I did not spend an inline comment on:

- `migration.en.md:1038` / `migration.ru.md:1042` — "the second is the one each subsection below belongs to" is wrong for six of the seven subsections under section 18 (Popover, Search expandable, Split button, Title, Toast and Tooltip are all named in the same sentence's first-wave list). Scoping it to Timepicker fixes it.
- `timepicker.spec.ts:1210` and `:1220` assert less than their names promise. `should clamp an unsupported format to the default` expects `DEFAULT_TIME_FORMAT`, which is also the input's own initial value and the host field's initial value, so it cannot tell a clamp from a binding that never arrived — set `HHmmss` first, then write garbage. `should report the bound min and max` only exercises `min`; the single `max` assertion is `toBeNull()`, which is the input default and passes with no binding at all.

Verified clean, for the record: no in-repo consumer reads or writes the four migrated members programmatically; `migration.ru.md` is a faithful translation of the English section; prettier, eslint and cspell pass on every new file; the `i0.` → `_angular_core.` change in the golden file is genuine api-extractor output, not drift (23 other goldens already use it); the Playwright baselines are unaffected; and no existing test was deleted or weakened.

packages/schematics/src/migrations/timepicker-signals/data.ts
```diff
@@ -0,0 +1,73 @@
...
+            '`max` are now signals — reading one is a double call, e.g. `this.timepicker().format()`. ' +
+            'Verify query reads manually.'
+    }
+];
+
+/** Printed once per project, after the per-file reports. */
+export const SUMMARY = [
+    '  `kbqValidationTooltip` unsubscribes. The setter subscribed to `incorrectInput` every time it ran ' +
+        'and never unsubscribed, so re-binding the input stacked another subscription and the last one ' +
+        'outlived the directive. It is an effect with a teardown now.',
```
lskramarov, line 66: **Two of the three SUMMARY bullets state something that is not true, and each is repeated in five user-facing places.**

**"the last one outlived the directive"** — it never did. `incorrectInput` is an `output()`, and `OutputEmitterRef`'s constructor registers a teardown:

```js
this.destroyRef.onDestroy(() => { this.destroyed = true; this.listeners = null; });
```

Every listener is dropped when the directive is destroyed, and `emit()` on a destroyed ref returns early. The real defect the effect fixes is the **stacking**: N re-binds meant N live subscriptions, so one rejected keystroke opened every tooltip ever bound — which is exactly what the new spec observes. As written, a consumer goes looking for a post-destroy leak that cannot happen, while the symptom they actually saw is not described.

**The id-shape claim is wrong for the default `APP_ID`.** `_IdGenerator.getId` (`@angular/cdk/fesm2022/id-generator.mjs`) appends the app id only when it is *not* the default, and its per-prefix counter starts at 0:

```js
if (!counters.hasOwnProperty(prefix)) { counters[prefix] = 0; }
```

`APP_ID` defaults to `'ng'`, so a real app gets `kbq-timepicker-0`, `kbq-timepicker-1`, … — byte-identical to the removed `uniqueComponentIdSuffix` counter, which also started at 0. The `a` prefix appears only under Jest, because `@angular/platform-browser/testing.mjs:71` provides `{ provide: APP_ID, useValue: 'a' }`; a TestBed measurement is not evidence here. Both siblings get it right — `checkbox-signals/data.ts:99` (*"The shape is unchanged for a default `APP_ID`"*) and `autocomplete-signals/data.ts:155`.

A consumer following this rewrites working `#kbq-timepicker-0` selectors into broken `#kbq-timepicker-a1` ones. Both claims also appear in `README.md`, `docs/guides/migration.en.md`, `docs/guides/migration.ru.md` and verbatim in `collection.json` and `migrations.json`.

packages/schematics/src/migrations/timepicker-signals/data.ts
```diff
@@ -0,0 +1,73 @@
...
+ * `required`, `disabled`, `focused`, `empty` and `errorState` as plain members — so those stay plain
+ * accessors. The four inputs the timepicker owns are signals now.
+ *
+ * - `timepicker.format` → `timepicker.format()` (value unchanged — auto-fixed)
+ * - `timepicker.min` / `max` → signals whose value changed: the getters returned the parsed date (warn)
+ * - `kbqValidationTooltip` → a plain input driven by an effect that finally unsubscribes (warn on writes)
+ */
+
+/** Members whose value is unchanged; a read must become a call. Auto-fixed. */
+export const SIGNAL_MEMBERS: readonly string[] = ['format'];
```
lskramarov, line 14: **Leaving `min`/`max` unrewritten turns a wrong-value bug into a silently-truthy one.**

`SIGNAL_MEMBERS` is `['format']` only; `min`/`max` are warn-only. That mirrors how `alert-signals` treats `alertColor` — but the hazard is not the same. `alertColor` went string → string, so truthiness was unchanged. `min`/`max` went `D | null` → a function that is **always** truthy:

```ts
if (this.timepicker.min && this.adapter.compareDateTime(this.timepicker.min, v) < 0) { … }
```

compiles fine after the upgrade (`compareDateTime(a: any, …)`), the guard now always passes, and `compareDateTime` is handed the `InputSignal` function instead of a date.

The warning text never tells the consumer to add `()` — it says only *"They report what was bound. Migrate by hand."* At minimum the message should say "add `()` **and** expect the bound value rather than the parsed one"; better, rewrite the read and warn about the value change separately, which is the only part that genuinely needs a human.

packages/schematics/src/migrations/timepicker-signals/data.ts
```diff
@@ -0,0 +1,73 @@
...
+    pattern: string;
+    message: string;
+}
+
+const TIMEPICKER_ANCHOR = '\\bKbqTimepicker\\b';
+
+export const warnPatterns: WarnPattern[] = [
+    {
+        anchor: TIMEPICKER_ANCHOR,
+        pattern: '\\.\\s*(?:format|min|max|kbqValidationTooltip)\\s*=[^=]',
```
lskramarov, line 47: **This regex is receiver-blind: it fires on any `.min =` / `.max =` in a file that merely names `KbqTimepicker`.**

`anchor` only requires `\bKbqTimepicker\b` somewhere in the file — an unrelated import or a comment is enough. Verified `true` for all of `this.slider.max = 10;`, `this.chart.scales.y.min = 0;`, `Math.max = noop;`.

Verified `false` (missed): `timepicker.format ??= f`, `timepicker.format += x` — the same gap as the `classifyAccess` write detection.

And the anchor is *stricter* than the file-admission gate this PR deliberately loosened: `TIMEPICKER_ANCHOR` is `'\\bKbqTimepicker\\b'`, but `referencesTimepicker` (`index.ts:250`) is `/\bKbqTimepicker\w*\b/`. A file whose only mention is `import { KbqTimepickerModule } …` is admitted, counted in `consumers` and gets the project summary — yet `\bKbqTimepicker\b` is `false` (the boundary fails before `M`), so **neither** warn pattern can fire there. Combined with the AST pass needing an explicit annotation, a `this.tp.format = …` in such a file is reported nowhere and silently stops compiling after the upgrade.

The AST pass right next to it is already receiver-scoped, so the write report can come from there instead of a regex (`checkbox-signals` does exactly that with a `writes: Set<string>`). That would also fix the double-reporting: today `this.timepicker.min = x` produces one block from `logWarnings` *and* one from `warnReceiverMembers` (whose walk has no write filter), and `logMessage` does no dedup.

packages/components/timepicker/timepicker.spec.ts
```diff
@@ -1205,4 +1206,102 @@ describe(KbqTimepicker.name, () => {
...
+
+            // The getter used to hand back the parsed value, so an unparseable bound value read as null.
+            expect(timepicker.min()).toBe('not a time');
+        });
+
+        it('should stop driving a validation tooltip once it is unbound', () => {
+            const fixture = createStandaloneComponent(TimepickerWithValidationTooltip);
+            const { componentInstance } = fixture;
+            const [first, second] = componentInstance.tooltips();
+            const firstShow = jest.spyOn(first, 'show');
```
lskramarov, line 1238: **Real 3-second timers escape this test.**

`jest.spyOn(first, 'show')` without `mockImplementation` calls through to the real `KbqPopUpTrigger.show()`, and the effect's handler then schedules `setTimeout(() => tooltip.hide(), validationTooltipHideDelay)` where `validationTooltipHideDelay === 3000` (`core/validation/validation.ts:9`).

The test body is synchronous, `createStandaloneComponent` uses `autoDetectChanges()` with real timers, and this repo's `testTimeout` is **2000 ms** with `jest-fail-on-console` on. Both `incorrectInput.emit()` calls arm a timer that fires roughly a second *after* the test finished, once the next `TestBed.resetTestingModule()` has destroyed the fixture — landing in whatever test is running at that moment.

The directive side is the root cause: `onCleanup` (line 438) disposes only the subscription, never the pending timeout, so in production a timepicker destroyed within 3 s of a rejected keystroke still calls `hide()` on a tooltip it no longer drives. Capture the handle and `clearTimeout` it in the teardown; wrap this test in `fakeAsync` + `tick(validationTooltipHideDelay)`, as the rest of the file does.

packages/schematics/src/migrations/timepicker-signals/index.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+                '    read() {\n' +
+                '        return this.timepicker.format;\n' +
+                '    }\n' +
+                '}\n'
+        );
+
+        expect((await run()).readText(ts)).toContain('return this.timepicker.format();');
+    });
+
+    it('leaves reads on a receiver of an unrelated type alone', async () => {
```
lskramarov, line 85: **This test is vacuous — it never reaches the code it is named for, which is why the receiver-scope bugs above have no guard.**

The fixture declares no `: KbqTimepicker`-annotated receiver anywhere, so `collectReceivers` returns `[]` and `migrateTsExpressions` bails at `if (receivers.length === 0) return content;` before `inReceiverScope` is ever consulted. What it proves is "a file with no receivers is untouched" — not "reads are discriminated by receiver".

No test in this file ever puts a timepicker receiver **and** a foreign receiver of the same member name in one source. Drop the `r.text === receiverText` comparison from `inReceiverScope`, or widen a receiver's scope so it leaks across classes, and all ten tests stay green.

A fixture that would actually guard it:

```ts
import { KbqTimepicker } from '@koobiq/components/timepicker';
class Other { format = 'HH:mm'; }
class Demo {
    read(timepicker: KbqTimepicker, other: Other) {
        return [timepicker.format, other.format];  // only the first may gain ()
    }
}
```

The same gap covers the other rewrite hazards: nothing here exercises `??=`/`+=`/`++`/`delete`, `this.tp!.format`, a parameter in a type position, or `fix` being *absent* (`run(fix: boolean = true)` always passes it explicitly, which is why the `ng update` default above slips through).

packages/schematics/src/migrations/timepicker-signals/index.ts
```diff
@@ -0,0 +1,305 @@
...
+/** A receiver whose static type is a timepicker, valid within `[start, end]` of the source. */
+interface Receiver {
+    /** Source text of the receiver expression, e.g. `timepicker` or `this.timepicker`. */
+    text: string;
+    start: number;
+    end: number;
+}
+
+/** Applies text-span edits to `content`, right-to-left, so earlier edits don't shift later offsets. */
+function applyEdits(content: string, edits: Edit[]): string {
```
lskramarov, line 36: **Altitude: this migration was cloned from the superseded `alert-signals` rather than the current `checkbox-signals`, which is the root cause of most of the findings above.**

Normalising the component name and diffing against `alert-signals/index.ts`: **188 of 241 non-blank lines are identical (78%)**. The fingerprints are unmistakable — `WRITABLE_MEMBERS` is an empty `Set` (so the `.set(` branch at line 148 is unreachable), and line 158 reads `// Read (incl. optional chain \`x?.compact\`)` where `compact` is a **`KbqAlert`** member (`alert-signals/data.ts:19`), not a `KbqTimepicker` one.

`checkbox-signals` shipped one commit earlier (`dec52b292`) and already carries every fix this review asks for: `options.fix ?? true`, `ASSIGNMENT_OPERATORS`, `ts.isFunctionLike`, `unwrapReceiver`, `TYPING_FACTORIES`, `UNRESOLVED_RECEIVER_MESSAGE`, and one parse + one receiver walk per file instead of two of each.

Two ways out, in order of preference:

1. Lift the engine — `applyEdits`, `findAncestor`, `isTypeReference`, `collectReceivers`, `inReceiverScope`, `classifyAccess`, `collectAccessEdits`, `logWarnings`, `WarnPattern` — into `packages/schematics/src/utils/` (the established home is `utils/typescript.ts`), parameterised by the `data.ts` shape that already exists. Copy counts today: `applyEdits` 16–17, `logWarnings` 15–16, `findAncestor` 13–14, `collectReceivers` 12–13, `WarnPattern` ~20. There is no shared module for any of it, so each of the bugs above has to be found and fixed in a dozen places.
2. Failing that, re-clone this one from `checkbox-signals`.

(`applyEdits` itself re-implements `tree.beginUpdate()`/`UpdateRecorder`, which the repo already uses in `utils/angular-parsing.ts:95` and `utils/icon-migration.ts` — and which throws on overlapping edits where the hand-rolled version silently produces garbage.)

packages/components/timepicker/timepicker.directive.ts
artembelik: вместо эффекта с проверкой и `untracked` перевалидация идёт из `ngOnChanges` и только при смене `min`/`max` - синхронно с привязкой, как у старого сеттера, поэтому закрыт и порядок CD

packages/components/timepicker/timepicker.spec.ts
artembelik: кроме `clearTimeout` очистка ещё и закрывает тултип, если скрытие было отложено - иначе при перепривязке в эти 3 секунды он оставался бы открытым навсегда

packages/schematics/src/migrations/timepicker-signals/data.ts
artembelik: `min`/`max` оставил в отчёте, а не переписал: значение изменилось, как у `freeRowsHeight` в textarea, а сообщение теперь прямо говорит добавить `()` и ждать привязанное значение

lskramarov (approved):

PR #1993, opened 2026-09-04
fix(tags)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

A full review of `tags` — the largest component in this wave, 21 decorator inputs across three interdependent classes.

### Most of the surface is constrained, and stays constrained

`KbqTagList` implements `KbqFormFieldControl`, `KbqTagInput` implements `KbqTagTextControl`, and both interfaces declare their members as plain properties. On top of that, most of `KbqTag`'s inputs fold in the parent's state:

```ts
get disabled(): boolean {
    return this._disabled || (this.tagList?.disabled ?? false);
}
```

A `model()` cannot carry the `booleanAttribute` transform a valueless attribute needs, so those stay accessors — the shape the reviewed `KbqButtonToggle` settled on and the one `checkbox` and `code-block` use in this wave.

**They are backed by signals now.** Read and write syntax is unchanged, so nothing at a call site moves; what changes is that the host bindings reading them re-render on their own instead of waiting for an unrelated change detection pass.

### What actually moved

The inputs on `KbqTagInput` that answer to nobody else:

- **`separatorKeyCodes`** was a setter with no getter — write-only. It is an `input()` with a `value ?? []` transform, so it can finally be read.
- **`separators`** was a getter that re-filtered the separator list on every call, from `onKeydown` and `onPaste`. It is a `computed`.
- **`addOnBlur`** was an accessor with manual `coerceBooleanProperty`; it is a plain `booleanAttribute` input.
- **`distinct`** was `input<boolean>(false)` with no transform, so `<input distinct>` passed the empty string — falsy — and duplicate tags were still accepted.

### Also

- Generated ids come from the CDK `_IdGenerator` in both the tag list and the tag input (`kbq-tag-list-1` → `kbq-tag-list-a1`). The tag list reports the id of its input when it has one, so both surface through the form field.
- The 21 `// TODO: Skipped for migration because:` comments across the three files are replaced by the actual reason each member kept its shape — the interface that names it, or the parent state it folds in.

## Migration

`tags-signals` runs from `ng update @koobiq/components@20`. It rewrites `addOnBlur` and `separators` reads on receivers typed `KbqTagInput` and reports the rest. The spec includes a test asserting that the interface-constrained accessors on `KbqTag` and `KbqTagList` are left completely alone.

## Testing

- `tag-input.spec.ts`: 3 new tests for the `null` separator coercion, the valueless `addOnBlur`/`distinct` attributes, and the generated id.
- Nine tests wrote `separatorKeyCodes` / `addOnBlur` directly on the directive; they bind from their hosts now. `TestTagInput` grew a bound `separatorKeyCodes` defaulting to `[ENTER]`, which is what `KbqTagsModule` provides — so the test that checks the globally configured separators still exercises the token rather than being overridden by a binding.
- `tag.component.spec.ts` and `tag-list.component.spec.ts` are unchanged and pass, which is the evidence that backing the accessors with signals preserved their behavior.
- `tags-signals/index.spec.ts`: 11 tests.
- Full `packages/components` (5164 tests) and `packages/schematics` (493 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


lskramarov (commented):
## Code review — `max` effort

15 findings, most severe first. Ten independent finder angles plus a gap sweep; framework claims were checked against `@angular/core` and `@angular/cdk` 20.2.x in `node_modules`, and the schematic's transforms were run against real inputs rather than reasoned about.

Two clusters dominate:

- **The schematic is the ~7th verbatim copy of the signal-read walker**, and it inherited the pre-fix version of it. Five separate findings below (`fix ?? true`, shadowed receivers, assignment shapes, receiver normalization, the missing template pass) were each already fixed in *some* sibling migration; none of the fixes reached this copy. The most consequential is that **`fix` arrives `undefined` under `ng update`, so the migration writes nothing at all** — the auto-fix path it exists for never runs.
- **`check-public-api-any` will fail** (`tags` 19 → 21) — `yarn run approve-public-api-any`.

Also worth a look: the documented `_IdGenerator` id-shape break doesn't happen in a default app (the `-a1` shape is a Jest artifact), and the section-18 intro's "two waves" framing contradicts 6 of its own 7 subsections.

Things I checked and cleared, so they're *not* below: `ngOnChanges` still fires for signal inputs (`writeToDirectiveInput` → `def.setInput` = `ngOnChangesSetInput`); no field-initializer ordering hazard in `KbqTagInput`; no `ngAcceptInputType_*` gaps; no NG0600 path; no external writer of `KbqTagList.uid`; the `i0.` → `_angular_core.` churn in the golden is toolchain-wide, not hand-editing; and no e2e baseline, `e2e.ts` scenario or `.md` docs page was invalidated.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+ * no element pass: `kbqTagInputFor` is an attribute on a native `<input>`, and nothing on `<kbq-tag>` or
+ * `<kbq-tag-list>` changed its read syntax.
+ */
+function referencesTags(content: string): boolean {
+    return /\bKbqTag\w*\b/.test(content) || content.includes(TAGS_PACKAGE);
+}
+
+export default function tagInputSignals(options: Schema): Rule {
+    return async (tree: Tree, context: SchematicContext) => {
+        const { project, fix } = options;
```
lskramarov, line 207: **`fix` arrives `undefined` under `ng update`, so the migration never writes anything.**

`migrations.json` declares no `"schema"` on any entry (`grep -c '"schema"' migrations.json` → 0), and `ng update` invokes the factory with no options, so `schema.json`'s `"default": true` never reaches the rule. `fix` is `undefined` → `commit()` takes the else branch for every file: the consumer's `addOnBlur`/`separators` reads stay broken and they get `[tags-signals] would update src/app/x.ts (run with --fix to apply)` — a flag `ng update` cannot pass.

12 siblings already guard against this; `navbar-signals-and-aria/index.ts:592` even carries the comment explaining it. The spec can't catch it because `runSchematic` goes through `collection.json`, which *does* declare the schema — a test that calls the rule directly (`runner.callRule(tagInputSignals({ project } as Schema), appTree)`) would.

```suggestion
        const { project } = options;
        // `ng update` invokes migrations with no options at all, and migrations.json declares no schema, so the
        // schema default never reaches us — applying the fix is the intended behaviour there.
        const fix = options.fix ?? true;
```

`schema.ts` needs `fix?: boolean` to match — `tag-slots/schema.ts` is the shape to copy.

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    // Already migrated: a call, or a `.set(...)` write — leave alone (idempotent).
+    if (ts.isCallExpression(parent) && parent.expression === node) return;
+    if (ts.isPropertyAccessExpression(parent) && parent.expression === node && parent.name.text === 'set') return;
+
+    // Write target: `x.member = RHS`. Every KbqTagInput signal member is `input()` (read-only), so there is no
+    // writable member — leave the write untouched (it becomes a compile error the consumer fixes by hand).
+    if (
+        ts.isBinaryExpression(parent) &&
+        parent.left === node &&
+        parent.operatorToken.kind === ts.SyntaxKind.EqualsToken
```
lskramarov, line 138: **Only `=` is treated as a write, so every other assignment shape is rewritten into source that no longer parses.**

`addOnBlur` was a real setter before this PR, so `tagInput.addOnBlur ||= this.fallback;` is legal pre-upgrade code. Its parent is a `BinaryExpression` with `BarBarEqualsToken`, the guard misses, and the read branch emits `tagInput.addOnBlur() ||= this.fallback;` — TS2779. Same for:

| Input | Output | Error |
| --- | --- | --- |
| `t.separators += x` | `t.separators() += x` | TS2364 |
| `t.separators++` | `t.separators()++` | TS2357 |
| `delete t.separators` | `delete t.separators()` | TS2703 |
| `[t.separators] = src` | `[t.separators()] = src` | invalid assignment target |

The warn pattern in `data.ts:45` needs `=` immediately after the member, so `||=` never matches either — the file is corrupted with no accompanying message.

The fix is to treat every assignment operator, the increment/decrement operators, `delete`, and destructuring targets as writes and report them instead of rewriting.

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
+    const start = node.getStart(sourceFile);
+    const end = node.getEnd();
+
+    return receivers.some((r) => r.text === receiverText && start >= r.start && end <= r.end);
```
lskramarov, line 122: **Receivers are matched by source text inside an offset range, so a shadowing binding of the same name gets rewritten.**

```ts
configure(tagInput: KbqTagInput) {
    this.legacy.forEach((tagInput: LegacyInput) => {
        if (tagInput.addOnBlur) { … }   // → tagInput.addOnBlur()
    });
}
```

The outer parameter's scope is the whole method, which contains the arrow, so the unrelated object gets `()` appended → `TypeError: tagInput.addOnBlur is not a function`.

Worse at module scope: `collectReceivers` gives a top-level `const tagInput: KbqTagInput` the whole `sourceFile` as its scope (line 105), so one such local turns *every* `tagInput.addOnBlur` in the file — in unrelated classes — into a call. For `this.` receivers the scope is the entire `ClassDeclaration` with no `this`-rebinding barrier, so a nested `class Inner` or a `function (this: Other) {…}` callback is rewritten too.

Resolving bindings per scope and picking the innermost declaration is what this needs.

tools/public_api_guard/components/tags.api.md
```diff
@@ -174,21 +173,22 @@ export type KbqTagFocusEvent = KbqTagEvent & {
 // @public
 export class KbqTagInput implements KbqTagTextControl, OnChanges {
     constructor();
-    get addOnBlur(): boolean;
-    set addOnBlur(value: boolean);
-    readonly addOnPaste: i0.InputSignalWithTransform<boolean, unknown>;
+    readonly addOnBlur: _angular_core.InputSignalWithTransform<boolean, unknown>;
```
lskramarov, line 176: **`check-public-api-any` fails: `tags` goes 19 → 21 and `tools/check-public-api-any/baseline.json` was not re-approved.**

Running the tool's own counting algorithm (`tools/check-public-api-any/index.ts:29-35`) over both goldens: base `e1464f997` → 19, this head → 21, while `baseline.json:44` still says `"tags": 19`. The API workflow exits 1 with `tags: 19 → 21`.

The two new `unknown`s are this line and `readonly distinct: …InputSignalWithTransform<boolean, unknown>` (line 183) — the accessor pair and the plain `InputSignal<boolean>` they replaced contributed 0. The added `ngAcceptInputType_*` lines are excluded by `isGenerated`.

`yarn run approve-public-api-any`.

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    for (const { anchor, pattern, message } of warnPatterns) {
+        if (!new RegExp(anchor).test(content) || !new RegExp(pattern).test(content)) continue;
+
+        logMessage(context.logger, [`${LABEL} ${filePath}`, `  ${message}`]);
+    }
+}
+
+/**
+ * A `.ts` file is a tags consumer if it names any of the exported symbols or imports the package. There is
+ * no element pass: `kbqTagInputFor` is an attribute on a native `<input>`, and nothing on `<kbq-tag>` or
```
lskramarov, line 198: **The "no element pass" justification overlooks `exportAs`.**

`tag-input.ts:123` is `exportAs: 'kbqTagInput, kbqTagInputFor'`, so a template reference variable can read the migrated members:

```html
<input kbqTagInputFor="list" #ti="kbqTagInput" />
@if (ti.addOnBlur) { … }
{{ ti.separators.length }}
```

After the upgrade `ti.addOnBlur` is the `InputSignal` function — the `@if` branch is permanently taken and `.length` is `0`. Neither `warnPatterns` anchor (`\bKbqTagInput\b`) matches a template that only names the exportAs string, and the migration never visits `.html` at all (line 217 collects only `.ts`). `alert-signals/index.ts:255-380` ships exactly this pass (ref collector + inline-template ranges + `.html` sweep) for the analogous case.

Being on a native `<input>` doesn't prevent a template reference — `exportAs` is what makes one possible.

docs/guides/migration.en.md
```diff
@@ -1113,6 +1113,24 @@ A `<kbq-split-button>` with no projected button no longer throws outside dev mod
...
+
+| Pattern                      | Manual migration                                                      |
+| ---------------------------- | --------------------------------------------------------------------- |
+| `.addOnBlur` / `.separators` | Read as calls — rewritten for you                                     |
+| `.separatorKeyCodes = …`     | Bind `[kbqTagInputSeparatorKeyCodes]`; in exchange it can be read now |
+| `.addOnBlur = …`             | Bind `[kbqTagInputAddOnBlur]`                                         |
+
+**`distinct` is a `booleanAttribute` input now.** A valueless `distinct` attribute used to pass the empty string, which is falsy, so duplicate tags were still accepted.
+
+**Generated ids changed shape**, from `kbq-tag-list-1` / `kbq-tag-list-input-1` to `kbq-tag-list-a1` / `kbq-tag-list-input-a1`. The tag list reports the id of its input when it has one, so both surface through the form field.
```
lskramarov, line 1130: **The documented id-shape change does not happen in a default app, and the rationale is wrong.**

`@angular/cdk/fesm2022/id-generator.mjs:18-26`:

```js
// Omit the app ID if it's the default `ng`. Since the vast majority of pages have one
// Angular app on them, we can reduce the amount of breakages by not adding it.
if (!counters.hasOwnProperty(prefix)) { counters[prefix] = 0; }
```

and `DEFAULT_APP_ID = 'ng'`. So a normally bootstrapped app still gets `kbq-tag-list-0` / `kbq-tag-list-input-0` — byte-identical to the deleted `nextUniqueId`. The `-a1` shape only appears under Jest, because `@angular/platform-browser/testing` provides `{ provide: APP_ID, useValue: 'a' }`.

So consumers are told to audit selectors and `aria-*` references for a break that won't happen, while the case that *does* change (a custom `APP_ID`, yielding `kbq-tag-list-<appid>0` — app id before the counter, no separator) goes unmentioned.

`data.ts:65-68` additionally claims the app id "keeps two Angular apps on one page from colliding": it's absent by default, and `counters` is a module-level global shared across injectors — the shared counter is what prevents collisions, not the prefix. That line also ships "the id of its input id" (duplicated word) into every consumer's `ng update` console.

Same text in `migration.ru.md:1134`, `tags-signals/README.md:46`, `data.ts:65`, `collection.json:163`, `migrations.json:126`.

packages/components/tags/tag-input.ts
```diff
@@ -150,50 +150,47 @@ export class KbqTagInput implements KbqTagTextControl, OnChanges {
...
      * @docs-private
      */
-    get separators(): KbqTagSeparator[] {
-        return this._separators.filter(
-            (separator) => separator.keyCode === undefined || this._separatorKeyCodes.includes(separator.keyCode)
-        );
-    }
-
-    private _separators: KbqTagSeparator[] = this.defaultOptions.separators || KBQ_TAG_INPUT_DEFAULT_SEPARATORS;
+    readonly separators = computed(() =>
```
lskramarov, line 167: **`computed()` memoizes on the array reference, so an in-place mutation of `separatorKeyCodes` is now ignored.**

Host: `codes = [ENTER];` bound as `[kbqTagInputSeparatorKeyCodes]="codes"`. The consumer runs `this.codes.push(COMMA)`. `bindingUpdated` compares with `Object.is` on the same array reference → the input is never re-written → the signal version never bumps → `separators()` returns its cached filter forever, and typing `,` never ends a tag.

Pre-PR the getter called `this._separatorKeyCodes.includes(...)` on the live array, so the very next keydown honoured the push. Same now for an in-place mutation of `KBQ_TAGS_DEFAULT_OPTIONS.separators`, which `allSeparators` (line 158) captures once behind the memo.

Not a blocker — but the migration guide advertises this input as "newly readable" and says only the read/write syntax changed, which doesn't warn about it.

packages/components/tags/tag-list.component.ts
```diff
@@ -336,28 +335,31 @@ export class KbqTagList
     /** Whether the tags in the list are editable. */
     readonly editable = input(false, { transform: booleanAttribute });
 
-    /** Whether the tags in the list are removable. */
-    // TODO: Skipped for migration because:
-    //  Accessor inputs cannot be migrated as they are too complex.
+    /**
+     * Whether the tags in the list are removable.
+     *
+     * Stays an accessor: the setter pushes the new state onto the projected tags, which have to be
```
lskramarov, line 341: **This rationale is contradicted by `KbqTag.removable`, and the push it defends clobbers a per-tag binding.**

`tag.component.ts:389` already folds the list's state in on every read:

```ts
get removable(): boolean {
    return this._removable() && (this.tagList?.removable ?? true);
}
```

So the tags *do* derive it — `syncTagsRemovableState()` (line 925) is not the mechanism this comment claims, and the AND is applied twice.

The push is also lossy. With `<kbq-tag-list [removable]="listRemovable"><kbq-tag [removable]="false">pinned</kbq-tag>`, toggling `listRemovable` false→true runs `tag.removable = true` and overwrites the pinned tag's own `_removable`. Angular won't re-write `[removable]="false"` because the bound expression's value didn't change, so the pinned tag permanently gains a remove icon.

Two more things about this comment specifically: it's `/** */` on a public `@Input`, so api-extractor drops the `// (undocumented)` marker and `tools/api-gen` renders this migration rationale as the consumer-facing description of `[removable]` on koobiq.io — the same is true of the new `draggable` JSDoc at line 308, which has no user-facing sentence at all. Elsewhere in this same file the identical rationale is correctly given as `//` (lines 203, 228, 247, 294). AGENTS.md: "Use `/** JSDoc */` for user-facing public API documentation and `//` for implementation comments."

packages/components/tags/tag.component.ts
```diff
@@ -359,68 +366,73 @@ export class KbqTag extends KbqColorDirective implements IFocusableOption, OnDes
...
     @Input({ transform: booleanAttribute })
     get disabled(): boolean {
-        return this._disabled || (this.tagList?.disabled ?? false);
+        return this._disabled() || (this.tagList?.disabled ?? false);
     }
 
     set disabled(value: boolean) {
-        this._disabled = value;
+        this._disabled.set(value);
         this.syncDragDisabledState();
```
lskramarov, line 432: **`syncDragDisabledState()` runs only from here and the constructor, so `CdkDrag.disabled` goes stale when the list's `draggable` flips.**

`<kbq-tag-list [draggable]="enabled">` starting with `enabled === false`: each tag's constructor runs `this.drag.disabled = !this.draggable` → `true`. Flipping `enabled` to `true` sets `dropList.disabled = false` and the host binding paints `.kbq-tag_draggable` (line 232) on every tag, but nothing re-runs `syncDragDisabledState()`.

CdkDrag's getter is `this._disabled || !!(this.dropContainer && this.dropContainer.disabled)` (`@angular/cdk/fesm2022/drag-drop.mjs:3502`), so the explicit `_disabled = true` sticks regardless of the container: the tags look draggable and aren't. Tags created *after* the flip construct correctly, so one list ends up with a mix.

Pre-existing, but this PR is the review of the component, and it just made `KbqTagList._draggable` a signal — which is exactly what a `computed`/`effect` derivation would need instead of the two push sites.

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    };
+
+    visit(sourceFile);
+
+    return receivers;
+}
+
+/** Whether a property access on a receiver is within one of the receiver's scopes. */
+function inReceiverScope(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile, receivers: Receiver[]): boolean {
+    const receiverText = node.expression.getText(sourceFile);
```
lskramarov, line 118: **Aliased imports, `!` and parentheses make a read invisible to both the rewrite and the warnings.**

Three shapes, all left untouched with no log line:

- `import { KbqTagInput as TagInput } …; read(t: TagInput) { return t.addOnBlur; }` — `isTypeReference` (line 62) compares `type.typeName.text` against the literal `'KbqTagInput'`.
- `this.tagInput!.addOnBlur` — receiver text is `this.tagInput!`, `r.text` is `this.tagInput`.
- `(tagInput).addOnBlur`.

Neither `warnPatterns` entry fires (one needs `.addOnBlur =`, the other needs `viewChild`), so the consumer ships `addOnBlur` still bound to the `InputSignal` — always truthy, no compile error, no diagnostic.

Comparing receiver text after stripping `!`, parentheses and `as`, and resolving imported aliases, closes all three.

packages/components/tags/tag-input.spec.ts
```diff
@@ -70,13 +76,15 @@ class TestTagInputDistinct {
...
-            <input [kbqTagInputFor]="tagList" (kbqTagInputTokenEnd)="add($event)" />
+            <input [kbqTagInputFor]="tagList" [kbqTagInputAddOnBlur]="addOnBlur" (kbqTagInputTokenEnd)="add($event)" />
         </kbq-tag-list>
     `
 })
 class TestTagInputDefaultSeparators {
     readonly tagInput = viewChild.required(KbqTagInput);
     readonly add = jest.fn();
+
+    addOnBlur = false;
```
lskramarov, line 87: **This flips the directive's `true` default to `false` for every test on this host — including one that now passes for the wrong reason.**

Six tests use `TestTagInputDefaultSeparators` (lines 305, 555, 568, 581, 611, 628). Four never touch `addOnBlur` and survive only because they drive `onKeydown`, never `blur`; the host's name still says "default" while its `addOnBlur` is the opposite of the directive's.

The negative autocomplete test at line 611 is already affected. `blur()` guards on `this.addOnBlur() && (…onInputBlur()(event) ?? true)`, so if the binding breaks or the `addOnBlur = true; detectChanges();` pair is dropped, `expect(add).not.toHaveBeenCalled()` still passes — `addOnBlur()` short-circuits before `onInputBlur` is ever consulted, and the DS-5279 regression the test names is no longer covered.

`false` isn't needed by either consumer: both 611 and 628 set it to `true`.

```suggestion
    addOnBlur = true;
```

packages/components/tags/tag-input.spec.ts
```diff
@@ -935,14 +945,52 @@ describe('KbqTagInput', () => {
...
+
+            fixture.detectChanges();
+
+            const directive = fixture.debugElement.query(By.directive(KbqTagInput)).injector.get(KbqTagInput);
+
+            fixture.componentInstance.separatorKeyCodes = null as unknown as number[];
+            fixture.detectChanges();
+
+            expect(directive.separatorKeyCodes()).toEqual([]);
+            expect(directive.separators().every(({ keyCode }) => keyCode === undefined)).toBe(true);
```
lskramarov, line 970: **Both new assertions are vacuous.**

With `separatorKeyCodes` transformed to `[]`, `separators()` filters `KBQ_TAG_INPUT_DEFAULT_SEPARATORS` — all five entries carry a `keyCode` — down to `[]`, and `[].every(...)` is `true` unconditionally. Invert the computed's predicate to `separator.keyCode !== undefined && …` and this still passes.

```suggestion
            expect(directive.separators()).toEqual([]);
```

Line 980 has the same problem: `addOnBlur = input(true, …)` already defaults to `true`, so `expect(directive.addOnBlur()).toBe(true)` passes even if the valueless `kbqTagInputAddOnBlur` attribute were ignored entirely — only the `distinct` half of "should treat valueless addOnBlur and distinct attributes as true" proves anything. A second host with `kbqTagInputAddOnBlur="false"` asserting `false` would fix it.

And `should generate a unique id` (line 984) compares no two ids, and `/^kbq-tag-list-input-\w+$/` matches the deleted `nextUniqueId` shape just as well — so it locks in nothing about the `_IdGenerator` switch, and overlaps `should have a default id` at line 762.

packages/schematics/src/migrations/tags-signals/data.ts
```diff
@@ -0,0 +1,73 @@
...
+        anchor: TAG_INPUT_ANCHOR,
+        pattern: '\\.\\s*(?:separatorKeyCodes|addOnBlur)\\s*=[^=]',
+        message:
+            'KbqTagInput.separatorKeyCodes and addOnBlur are read-only signal inputs now, so a programmatic ' +
+            'write no longer compiles. Bind [kbqTagInputSeparatorKeyCodes] / [kbqTagInputAddOnBlur] instead. ' +
+            '`separatorKeyCodes` was a setter with no getter, so in exchange it can finally be read.'
+    },
+    {
+        anchor: TAG_INPUT_ANCHOR,
+        pattern: '(?:viewChild|ViewChild|contentChild|ContentChild)[^\\n;]*\\bKbqTagInput\\b',
```
lskramarov, line 53: **This warning fires on the exact code the schematic just fixed correctly, and advises breaking it.**

Given `@ViewChild(KbqTagInput) tagInput: KbqTagInput;` + `this.tagInput.addOnBlur`, the rewrite pass correctly produces `this.tagInput.addOnBlur()` — pinned by `index.spec.ts:67`. Then `[^\n;]*` spans past the `)` into the annotation, the regex matches, and the user is told "reading one is a double call, e.g. `this.tagInput().addOnBlur()`". `this.tagInput` is the instance, not a signal, so following that advice breaks working code the migration had already fixed.

The warning is right only for the signal form `viewChild(KbqTagInput)`, which *is* genuinely skipped by `collectReceivers` (no explicit annotation) — the pattern needs to exclude the decorator spelling.

The sibling pattern at line 45 has the mirror problem: no receiver typing at all, so a consumer's own host field written as `this.addOnBlur = true` and bound via `[kbqTagInputAddOnBlur]` is told its correct code "no longer compiles". The converted specs in this very PR (`testTagInput.addOnBlur = true`) are exactly that shape.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov, line 1038: **The "two waves" framing is false for 6 of the 7 subsections below it.**

Section 18's subsections are Popover, Search expandable, Split button, **Tags**, Title, Toast, Tooltip. This sentence names popover, search-expandable, split-button, title, toast and tooltip as the *first* wave, then says "the second is the one each subsection below belongs to" — so a reader concludes Popover/Title/Toast/Tooltip were re-reviewed in a second pass.

Same defect at `migration.ru.md:1042` ("ко второй относится каждый из подразделов ниже").

Appending `tags` to the existing list and bumping "Ten" to "Eleven" is the smaller, correct edit.

packages/schematics/src/migrations/tags-signals/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    ts.SyntaxKind.ProtectedKeyword,
+    ts.SyntaxKind.ReadonlyKeyword
+]);
+
+/**
+ * Collects the receivers annotated with `typeName`, by explicit annotation only (no cross-package type
+ * resolution): method/function params, class fields (incl. `@ViewChild(KbqTagInput) x: KbqTagInput` and constructor
+ * parameter-properties) and typed locals.
+ */
+function collectReceivers(sourceFile: ts.SourceFile, typeName: string): Receiver[] {
```
lskramarov, line 77: **This is the ~7th verbatim copy of the walker, and it's why five of the findings on this PR exist.**

Across `packages/schematics/src/migrations/`: `applyEdits` + `interface Edit` in **10** copies (8 hash identically; `list-tree-multiple-input` and `tag-slots` have already drifted), `isFunctionLike`/`findAncestor`/`FIELD_MODIFIERS` in **7**, `collectReceivers`/`Receiver` in **6**, `inReceiverScope` in **4** identical copies, and `collectAccessEdits` byte-for-byte identical to `alert-signals/index.ts:163`.

`form-field-signals` had already generalized `collectReceivers` to take a `target`; this PR generalizes it again to take a `typeName`, so two incompatible parameterizations of one function now coexist. Each of the `fix ?? true`, shadowing, receiver-normalization, assignment-shape and template-pass fixes was made in *some* sibling, and none of them reached this copy.

The extraction: `packages/schematics/src/utils/signal-member-migration.ts` exporting `Edit`/`Receiver`/`applyEdits` plus `migrateSignalMemberReads(content, fileName, { typeName, signalMembers })`, and a `runFileMigration({ tree, context, label, references, transform })` for the `Rule` body. `utils/typescript.ts:74` already sets the precedent by hosting `collectInlineTemplateRanges`.

Two smaller things in the same block: `isFunctionLike` (39) and `findAncestor` (48) reimplement `ts.isFunctionLike` and `ts.findAncestor` from the TypeScript API this file already imports — and the local copy is narrower, missing `MethodSignature`/`FunctionType`. And `WRITABLE_MEMBERS` (`data.ts:25`) is a permanently empty `Set`, so the `.set(` branch at 140-146 and the `.set` idempotency guard at 131 are unreachable; the write case collapses to a plain `return`.

packages/schematics/src/migrations/tags-signals/data.ts
artembelik: Регулярку не правил, а убрал: запись и чтение через сигнальный запрос теперь сообщаются по разрешённому приёмнику, так что поле `@ViewChild` и своё поле хоста больше не задеваются.

packages/components/tags/tag-input.spec.ts
artembelik: Тест id заменил не сравнением двух id, а проверкой своего `APP_ID`: при `ng` форма совпадает со старым счётчиком, различие видно только там.

docs/guides/migration.en.md
artembelik: В `collection.json` и `migrations.json` было только «ids come from the CDK `_IdGenerator`» - это верно, их не трогал; заодно старый номер: счётчик начинался с 0, а не с 1.

lskramarov (approved):

NikGurev (approved):

PR #1995, opened 2026-09-04
fix(dl)!: errors following a full review of the component (#DS-5482)

artembelik:
## What

The audit pass over the seven components that were already signal-based. **Six of them needed no change** — `actions-panel`, `dynamic-translation`, `overflow-items`, `resizer`, `skeleton` and `content-panel` are already compliant on signal inputs, `booleanAttribute` / `numberAttribute`, `_IdGenerator`, `protected` for template members, `readonly` on injections, and `@docs-private`. This PR is the one component where the audit found something.

### Five inputs never got a coercion transform

```ts
readonly verticalBreakpoint = input(400, { transform: numberAttribute });
readonly minWidth = input<number | undefined>();          // no transform
readonly wide = input(false);                              // no transform
readonly resizable = input(false, { transform: booleanAttribute });
readonly vertical = input<boolean | null>(null);           // no transform
```

So **`<kbq-dl wide>` and `<kbq-dl vertical>` did nothing** — a valueless attribute passes the empty string, which is falsy — while `<kbq-dl resizable>` on the line above worked. That inconsistency inside one class is the whole finding.

### `vertical` is tri-state, so `booleanAttribute` alone was wrong

`null` means "decide from `verticalBreakpoint`", and `booleanAttribute(null)` is `false`. It uses a transform that passes `null` through and coerces everything else, so the auto mode survives and `<kbq-dl vertical>` means true.

### The numeric ones

`minWidth`, `dtMinWidth` and `ddMinWidth` report `number | undefined` — what an unbound description list always held. A static attribute used to reach the layout arithmetic as a string, which coerced in the `<=` comparison but not in `Math.max`.

## Migration

`dl-attribute-coercion` runs from `ng update @koobiq/components@20`. Warn-only: whether markup relied on a valueless `wide` being ignored is a decision the call site owns.

It reads templates in both `.html` files and inline `.ts` templates, since `<kbq-dl>` is written in either.

## Testing

- `dl.component.spec.ts`: 34 → 39 tests. The valueless `wide` and `vertical` attributes, `null` surviving as the `vertical` default, the coerced numeric attributes, and the unbound optional widths.
- `dl-attribute-coercion/index.spec.ts`: 6 tests, including one that a bound `[wide]` is not reported.
- Full `packages/components` (5166 tests) and `packages/schematics` (488 tests) suites pass.
- `check-api` is in sync.

BREAKING CHANGE


lskramarov (commented):
## Code review — `max` effort

15 findings, most severe first. Ten independent finder angles plus a gap sweep; each finding below was confirmed by at least two of them or by running the code.

**Blocking:** the `check-public-api-any` gate fails as submitted (`dl: 2 → 7`), so the API workflow is red before anything else is considered.

The three that matter most for consumers:

1. `optionalNumberAttribute` yields `NaN` rather than `undefined`, and `NaN` survives every `??` fallback — it reaches `grid-template-columns`, `aria-value*` and the `dtWidth` model. This is a regression against the pre-PR `Math.max(0, '') === 0`, and it is the behavior the README and the schematic message already promise.
2. The schematic reports the shapes that did not change (`<kbq-dl wide>`, `dtMinWidth="120"`) and stays silent on the ones that did (`vertical="false"`, `wide="false"`, `dtMinWidth` valueless / empty / interpolated). Its advice for a valueless `vertical` — remove it — does not preserve behavior either.
3. The *bound* forms change meaning as well — a bound `undefined` on `vertical` switches a permanently-horizontal list to auto, and every falsy non-boolean on `wide` inverts — and that class is documented nowhere and reported by nothing.

The rest are in the inline comments. Nothing here is style-only; the two conventions nits I found (change-history comments at `dl.component.ts:231`/`:244` and `dl.component.spec.ts:644`, and the RU attribution line at `migration.ru.md:1072` breaking the active-voice shape its five siblings use) were cut for space.

packages/components/dl/dl.component.ts
```diff
@@ -25,6 +28,10 @@ import { KBQ_WINDOW, kbqInjectA11yLocaleConfiguration, kbqInjectNativeElement }
 import { KbqResizable, KbqResizer, KbqResizerDirection, KbqResizerSizeChangeEvent } from '@koobiq/components/resizer';
 import { debounceTime, startWith } from 'rxjs/operators';
 
+/** Coerces an optional numeric input, keeping `undefined` distinguishable from `0`. */
+const optionalNumberAttribute = (value: unknown): number | undefined =>
+    value == null ? undefined : numberAttribute(value);
```
lskramarov: **`optionalNumberAttribute` returns `NaN`, not `undefined`** — and `??` cannot catch `NaN`.

Angular's `numberAttribute(value, fallbackValue = NaN)` falls back to `NaN` for anything non-numeric, and the empty string is non-numeric. So `<kbq-dl resizable dtMinWidth>` (or `dtMinWidth="{{ cfg.min }}"` resolving to `''`) gives `dtMinWidth() === NaN`, and from there:

- `resolvedDtMinWidth = this.dtMinWidth() ?? this.measuredDtWidth()` keeps `NaN` (`??` only catches nullish), so `normalizedDtMinWidth = Math.max(0, NaN)` is `NaN`;
- the host writes `--kbq-description-list-dt-min-width: NaNpx`, which makes `minmax(var(--kbq-description-list-dt-min-width), 1fr)` in `dl.scss:49` invalid at computed-value time — the whole `grid-template-columns` declaration drops;
- the separator renders `aria-valuemin="NaN"` / `aria-valuemax="NaN"` on `role="separator"`;
- `setDtWidth` computes `Math.min(maxDtWidth, Math.max(NaN, width)) === NaN`, and `NaN !== this.dtWidth()` is *always* true, so `dtWidth.set(NaN)` latches and `dtWidthChange` emits `NaN`;
- `handleResizeDblClick`'s `this.dtWidth() === this.normalizedDtMinWidth()` is never true (`NaN === NaN`), so the restore-default branch is unreachable, and `updateResizeCursor`'s `if (max <= min)` short-circuit is false for `NaN`, so a separator that cannot move still advertises `col-resize`.

This is a regression: before this PR the same markup reached `Math.max(0, '') === 0`. It is also what the README table and the schematic message already promise ("a non-numeric one reads as undefined") — the transform just does not implement it.

```suggestion
/** Coerces an optional numeric input, keeping `undefined` distinguishable from `0`. */
const optionalNumberAttribute = (value: unknown): number | undefined => {
    if (value == null) return undefined;

    const coerced = numberAttribute(value);

    return Number.isFinite(coerced) ? coerced : undefined;
};
```

`coerceNumberProperty(value, undefined)` from `@angular/cdk/coercion` does the same thing and is already used in `core/common-behaviors/tabindex.ts` and `core/formatters/number/formatter.ts`.

packages/schematics/src/migrations/dl-attribute-coercion/data.ts
```diff
@@ -0,0 +1,53 @@
...
+export interface WarnPattern {
+    /** Only evaluated for files that also render the element. */
+    pattern: string;
+    message: string;
+}
+
+export const warnPatterns: WarnPattern[] = [
+    {
+        // A valueless `wide` or `vertical`: the attribute name with no `=` and no `[` prefix.
+        pattern: '<kbq-dl[^>]*\\s(?:wide|vertical)(?![\\w-]*\\s*=)',
```
lskramarov: **The lookahead excludes every `=` form, which is where the behavior actually inverts.** `(?![\w-]*\s*=)` skips any `wide=` / `vertical=`, so `<kbq-dl vertical="false" wide="false">` is never reported — yet that is the largest visual break in this PR:

- before: the untransformed inputs held the truthy string `'false'`, so `isVertical()` was truthy and `[class.kbq-dl_wide]` applied — the list rendered **wide and vertical**, and `resizerVisible()` was false;
- after: `booleanAttribute('false')` is `false` — the list is **never** vertical (the `vertical() === null` branch is skipped, so the breakpoint never gets a say) and the resizer appears.

`<kbq-dl vertical="">` / `<kbq-dl wide="">` flip the other way (empty string was falsy, `booleanAttribute('')` is `true`) and are excluded by the same lookahead, even though they are byte-for-byte equivalent to the valueless form the pattern *does* report.

Worth noting the two shapes this pattern does catch are the ones a `strictTemplates` consumer could never have written, since the old write types were `boolean` / `boolean | null`.

The pattern over-reports in the other direction too: `[^>]*` runs straight through quoted values, so `wide` / `vertical` as a *word* anywhere in the tag matches — `<kbq-dl class="my-list vertical">`, `<kbq-dl title="Very wide list">`, `<kbq-dl [ngClass]="{ wide: isWide }">`, and `<kbq-dl-other wide>` (the element name is not anchored) all print "Remove it if the markup was relying on it being ignored" for an attribute that is not there. The same `[^>]*` also *stops* at a `>` inside a quoted value, so a genuine `<kbq-dl [verticalBreakpoint]="w > 500 ? 600 : 400" wide>` is skipped.

packages/components/dl/dl.component.ts
```diff
@@ -137,7 +144,10 @@ export class KbqDlComponent {
     readonly horizontalAlign = input<KbqDlAlign>('start');
 
     /** Forces the vertical layout; `null` lets the list decide based on `verticalBreakpoint`. */
-    readonly vertical = input<boolean | null>(null);
+    readonly vertical = input<boolean | null, unknown>(null, {
```
lskramarov, line 186: **The bound forms change meaning too, and nothing warns about them.** The migration guide and the schematic both talk exclusively about valueless *attributes*; `index.spec.ts:73` ("leaves a bound wide alone") locks in that bindings are never reported. But adding a transform changes what a binding means:

- `[vertical]="row.vertical"` with an optional field left `vertical() === undefined` pre-PR. `updateLayout`'s guard is `if (this.vertical() === null)`, which is **false** for `undefined`, so `autoVertical` was never written and the list stayed permanently horizontal. The transform now maps `undefined → null`, the guard passes, and the same markup starts flipping to vertical below `verticalBreakpoint`.
- `booleanAttribute` returns true for anything that is neither nullish nor the string `false`, so on `wide` every falsy non-boolean inverts: `[wide]="items.length"` with `0`, or `[wide]="label"` with an empty string, was falsy and is `true` now — `kbq-dl_wide` appears where it did not.

This is the widest-reach change in the PR (a binding needs no `strictTemplates` escape hatch to exist), and it is the one with no coverage in the guide, the schematic, or the tests. At minimum it belongs in `migration.{en,ru}.md` next to the valueless-attribute paragraph.

tools/public_api_guard/components/dl.api.md
```diff
@@ -26,8 +26,8 @@ export class KbqDlComponent {
     constructor();
     // (undocumented)
     protected get currentDtWidth(): number;
-    readonly ddMinWidth: _angular_core.InputSignal<number | undefined>;
-    readonly dtMinWidth: _angular_core.InputSignal<number | undefined>;
+    readonly ddMinWidth: _angular_core.InputSignalWithTransform<number | undefined, unknown>;
```
lskramarov, line 29: **This fails the `check-public-api-any` gate.** `.github/workflows/api.yml:20` runs it, and `tools/check-public-api-any/baseline.json` still records `"dl": 2` while the regenerated golden file now carries seven counted `unknown` tokens (`ddMinWidth`, `dtMinWidth`, `minWidth`, `resizable`, `vertical`, `verticalBreakpoint`, `wide`). Running the tool's own counter over this file gives `7`, so it prints `dl: 2 → 7` and exits 1.

Two ways out, and the second is the better one:

1. `yarn run approve-public-api-any` to record the new counts.
2. Narrow the write types instead of taking `unknown`. `unknown` also switches off `strictTemplates` on all five bindings — `[dtMinWidth]="'oops'"` and `[wide]="{}"` compile now where they did not before. `alert` already does this: `InputSignalWithTransform<"error" | "warning" | "success" | "info", "error" | "warning" | "success" | "info" | null | undefined>`. `input<number | undefined, number | string | null | undefined>(...)` still accepts `<kbq-dl dtMinWidth="120">` while keeping the checker, and keeps the ratchet green.

packages/schematics/src/migrations/dl-attribute-coercion/data.ts
```diff
@@ -0,0 +1,53 @@
...
+        message:
+            'A valueless `wide` or `vertical` attribute on <kbq-dl> used to pass the empty string, which is ' +
+            'falsy, so it did nothing. Both are coerced now and the attribute means true. Remove it if the ' +
+            'markup was relying on it being ignored.'
```
lskramarov: **"Remove it if the markup was relying on it being ignored" does not preserve behavior for `vertical`.** The advice is right for `wide`, which really was inert. `vertical` was not: pre-PR `<kbq-dl vertical>` gave `vertical() === ''`, which is not `null`, so `updateLayout`'s `if (this.vertical() === null)` guard never ran and `autoVertical` stayed `null` forever. The list was **pinned horizontal** and never switched at the breakpoint.

Deleting the attribute as advised hands the decision back to `verticalBreakpoint` (default 400), so a narrow list that used to stay horizontal now flips to the vertical layout. The behavior-preserving rewrite is `[vertical]="false"`.

The same wording is in `README.md:30` and `docs/guides/migration.en.md:1062` ("`<kbq-dl wide>` and `<kbq-dl vertical>` used to do nothing").

packages/components/dl/dl.component.ts
```diff
@@ -217,10 +228,34 @@ export class KbqDlComponent {
...
+                });
+
+            // The breakpoint comparison was only ever re-run on resize, so changing any of its three
+            // inputs left the layout on the answer computed for the previous ones — including `vertical`
+            // going back to null, which hands the decision back to a stale `autoVertical`.
+            effect(
+                () => {
+                    this.vertical();
+                    this.verticalBreakpoint();
+                    this.minWidth();
```
lskramarov: **The dependency list is wrong in both directions.** `updateLayout()` does two things, and the effect subscribes to the inputs of the first while re-running the second.

*Too narrow* — `dtMinWidth` / `ddMinWidth` are missing, yet `updateLayout()` re-clamps `dtWidth` through `setDtWidth` against `normalizedDtMinWidth()` and `maxDtWidth`, both derived from them. With `resizable`, `[vertical]="false"`, host 600, `dtMinWidth` 100 and `[(dtWidth)]="150"`, raising `dtMinWidth` to 400 leaves `dtWidth() === 150` while the host renders `--kbq-description-list-dt-min-width: 400px` and the separator reports `aria-valuenow="400"`. The model value, the CSS variable and the ARIA value disagree until the next resize — verbatim the staleness the comment above says this effect fixes.

*Too broad* — when `vertical` is set explicitly the breakpoint branch is skipped entirely, so `verticalBreakpoint` and `minWidth` are provably inert; a change to either still runs the full `updateLayout()`, including the `dtWidth` re-clamp against live geometry, which pre-PR only a resize could reach.

packages/schematics/src/migrations/dl-attribute-coercion/data.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+        pattern: '<kbq-dl[^>]*\\s(?:minWidth|dtMinWidth|ddMinWidth)\\s*=\\s*"[^"{]',
+        message:
+            '`minWidth`, `dtMinWidth` and `ddMinWidth` on <kbq-dl> are numeric inputs now. A static attribute ' +
+            'used to reach the layout arithmetic as a string, which happened to coerce in a comparison but ' +
+            'not in `Math.max`. The value is a number now; a non-numeric one reads as undefined.'
```
lskramarov: **Both claims in this sentence are false**, and it ships to consumers as the `ng update` warning (plus `README.md:31`, `docs/guides/migration.en.md:1066` and `migration.ru.md:1070`).

1. *"a non-numeric one reads as undefined"* — it reads as `NaN`. A consumer who guards with `?? fallback` or `=== undefined` will find the guard never fires, and `NaN` reaches the grid (see the comment on `dl.component.ts:32`).
2. *"coerced in a comparison but not in `Math.max`"* — `Math.max` applies ToNumber to every argument: `Math.max(0, '120') === 120`. Tracing `<kbq-dl dtMinWidth="120">` on the pre-PR code, `normalizedDtMinWidth` was already `120`, `maxDtWidth`'s subtraction coerced, and `600 <= '700'` was `true` — every path produced the right number. The stated motivation for the breaking change describes a bug that did not exist.

packages/schematics/src/migrations/dl-attribute-coercion/data.ts
```diff
@@ -0,0 +1,53 @@
...
+    {
+        // A valueless `wide` or `vertical`: the attribute name with no `=` and no `[` prefix.
+        pattern: '<kbq-dl[^>]*\\s(?:wide|vertical)(?![\\w-]*\\s*=)',
+        message:
+            'A valueless `wide` or `vertical` attribute on <kbq-dl> used to pass the empty string, which is ' +
+            'falsy, so it did nothing. Both are coerced now and the attribute means true. Remove it if the ' +
+            'markup was relying on it being ignored.'
+    },
+    {
+        pattern: '<kbq-dl[^>]*\\s(?:minWidth|dtMinWidth|ddMinWidth)\\s*=\\s*"[^"{]',
```
lskramarov: **This pattern reports the call sites that did not change and misses the ones that did.**

`<kbq-dl dtMinWidth="120">` matches — and behaves identically before and after (`Math.max(0, '120') === 120`).

Verified misses, all of which resolve to `''` and therefore to `NaN` under the new transform:

- `<kbq-dl dtMinWidth>` — no `=` at all, so neither pattern matches it;
- `<kbq-dl dtMinWidth="">` — `[^"{]` needs at least one character;
- `<kbq-dl dtMinWidth="{{ cfg.min }}">` — excluded by `[^"{]` on purpose, but this is exactly the shape that flips from `0` to `NaN` when the expression is `undefined`;
- `<kbq-dl dtMinWidth='120'>` — single quotes are legal HTML and are not covered.

packages/components/dl/dl.component.ts
```diff
@@ -122,10 +129,10 @@ export class KbqDlComponent {
     readonly dtWidth = model<number | null>(null);
```
lskramarov: **`dtWidth` is the sixth input with no coercion, and the only one whose value escapes to consumers.** The PR's premise is "the five inputs that never got a coercion transform" — `dtWidth` is not among them, but it has the same hole:

`<kbq-dl resizable dtWidth="120" dtMinWidth="120">` leaves `typeof dtWidth() === 'string'` on a signal declared `number | null`. That string is what a `[(dtWidth)]` parent reads back, what `[style.--kbq-description-list-dt-width.px]` renders, and what `handleResizeDblClick` compares — `'120' === 120` is `false`, so the first double-click takes the wrong branch. It self-corrects only after a drag, because `setDtWidth` launders the value through `Math.min`/`Math.max`.

`model()` takes no `transform`, so this needs a different fix from the other five (a `computed` view over the model, or coercing on the way in). Either way it is worth deciding explicitly rather than leaving the stated sweep silently incomplete — the migration guide tells consumers the numeric inputs are numbers now, and this one still is not.

docs/guides/migration.en.md
```diff
@@ -1035,7 +1035,7 @@ for each option it deselected and reporting the shortened value to the form cont
 
 ### 18. Component review (20.3.0)
 
-Ten components went through a full review in 20.3.0: notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
+Components went through a full review in 20.3.0, in two waves. The first covered notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select; the second is the one each subsection below belongs to. Each review closed the members that were never part of the component's contract, moved inputs to signals where that was the point of it, and fixed the behavior it uncovered along the way. Only the changes that reach a consumer are listed here.
```
lskramarov: **The reword contradicts itself.** It names notification-center, popover, search-expandable, select, split-button, title, toast, tooltip, tree and tree-select as the first wave, then says "the second is the one each subsection below belongs to" — but the subsections below are Description list, **Popover**, **Search expandable**, **Split button**, **Title**, **Toast**, **Tooltip**. Six of the seven are first-wave components named in the same sentence; only Description list is second-wave.

`docs/guides/migration.ru.md:1042` carries the identical error.

Keeping the original sentence shape and appending `dl` to the component list stays correct as more reviews land in 20.3.0, and needs no rewording on the next one.

packages/components/dl/dl.component.spec.ts
```diff
@@ -626,4 +628,123 @@ describe(KbqDlComponent.name, () => {
...
+
+        // The comparison used to run only on resize, so a new breakpoint left the old answer in place.
+        fixture.componentRef.setInput('verticalBreakpoint', 700);
+        fixture.detectChanges();
+
+        expect(getDlElement(fixture).classList).toContain('kbq-dl_vertical');
+        flush();
+    }));
+
+    it('should hand the decision back to the breakpoint when vertical returns to null', fakeAsync(() => {
```
lskramarov, line 710: **This test passes on `main` too, so it does not guard the effect.** The stub width is fixed at 600 and `verticalBreakpoint` is set to 700 *before* `tick(100)`, so the single pre-PR `updateLayout()` (from the debounced `startWith(null)`) already sets `autoVertical = 600 <= 700 = true`.

Toggling `vertical` to `false` never writes `autoVertical` (`updateLayout` only touches it when `vertical() === null`), so going back to `null` yields `isVertical() = null ?? true = true` with or without the effect. The stale value and the fresh value happen to be identical.

To make it discriminating, change what the stale answer would be: keep the default breakpoint (400) so the first measurement gives `autoVertical = false`, set `vertical` to `false`, *then* `setInput('verticalBreakpoint', 700)`, then `vertical` back to `null`. Only the effect recomputes `autoVertical` to `true` at that point.

The sibling test at line 631 is a real guard — it is this one that is vacuous.

packages/components/dl/dl.component.spec.ts
```diff
@@ -626,4 +628,123 @@ describe(KbqDlComponent.name, () => {
...
+        const fixture = createComponent(DlWithValuelessAttributes);
+
+        fixture.detectChanges();
+
+        const dl = fixture.debugElement.query(By.directive(KbqDlComponent)).nativeElement as HTMLElement;
+
+        expect(dl.classList).toContain('kbq-dl_wide');
+    });
+
+    it('should keep null as the vertical default', () => {
```
lskramarov: **Both default tests are vacuous — Angular never applies a transform to an input's declared default.** `createInputSignal` does `node.value = initialValue; node.transformFn = options?.transform;`, so the initial value bypasses `transformFn` entirely.

`expect(fixture.componentInstance.vertical()).toBeNull()` here, and the three `toBeUndefined()` assertions at line 719, pass identically with all five transforms deleted from the component. They assert `input()`'s stored default, not the coercion — and the comment below ("`null` … must survive the transform") states a claim the body cannot observe.

Surviving the transform means a *bound* `null`: `fixture.componentRef.setInput('vertical', null)` after a non-null value, which the test at line 652 already does. What is missing instead is a host for `<kbq-dl dtMinWidth="abc">` / `<kbq-dl dtMinWidth>` and for `[wide]="'false'"` — the cases the README and the schematic message promise a behavior for.

packages/schematics/src/migrations/dl-attribute-coercion/index.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        return appTree.exists(`${root}/app.html`) ? `${root}/app.html` : `${root}/app.component.html`;
+    }
+
+    async function run(): Promise<Tree> {
+        const [first] = projects.keys();
+
+        return runner.runSchematic(SCHEMATIC_NAME, { project: first } satisfies Schema, appTree);
+    }
+
```
lskramarov, line 48: **All six tests go through `firstHtmlPath()`, so the `.ts` half of the schematic is untested.** `index.ts:21` states the reason both extensions are visited: *"Both `.ts` and `.html` are visited, because the element is written in templates of either kind."* Nothing here writes an inline `template:` — and nothing exercises the `content.includes(DL_PACKAGE)` branch of `referencesDl` either.

Drop `'.ts'` from `EXTENSIONS` (index.ts:8), or delete the package half of the gate, and this suite still passes green while every consumer who writes `<kbq-dl>` in an inline template silently stops being reported.

One test writing a `.ts` file whose inline template contains `<kbq-dl wide>` would close it. (The sibling `popover-leave-delay/index.spec.ts` has the inverse gap — only `.ts` — so copying it forward inherited the shape but flipped the untested half.)

packages/components/dl/dl.component.ts
```diff
@@ -217,10 +228,34 @@ export class KbqDlComponent {
...
-                .subscribe(() => this.updateLayout());
+                .subscribe(() => {
+                    layoutMeasured = true;
+                    this.updateLayout();
+                });
+
+            // The breakpoint comparison was only ever re-run on resize, so changing any of its three
+            // inputs left the layout on the answer computed for the previous ones — including `vertical`
+            // going back to null, which hands the decision back to a stale `autoVertical`.
+            effect(
```
lskramarov, line 293: **Four moving parts here re-derive one boolean.** A mutable closure flag whose only writer is a subscription callback, an effect that needs an explicit `Injector` because `afterNextRender`'s callback is not an injection context, a hand-maintained dependency list, and an `untracked()` that is load-bearing only because `updateLayout()` both reads `isVertical()` and writes `autoVertical`.

The root cause is that `autoVertical` is *derived* state ("is the host narrower than the breakpoint?") held imperatively, and its `null` conflates "never measured" with "`vertical` was explicit, so I skipped the write" — `layoutMeasured` is a second variable invented to disambiguate the first. Make the measurement the state and the decision derived:

```ts
private readonly hostWidth = signal<number | null>(null); // written only by the resize subscription

private readonly autoVertical = computed(() => {
    const width = this.hostWidth();

    return width === null ? null : width <= (this.minWidth() ?? this.verticalBreakpoint());
});
```

`isVertical` at line 167 is unchanged and still falls back to `false` before the first measurement, so the flag's stated job is met structurally. That deletes the flag, the effect, the `untracked` call, the `Injector` field and the `if (this.vertical() === null)` branch in `updateLayout`, and it cannot drift out of sync when a fourth input joins the decision. It also covers the case the flag-based effect cannot express: a list that starts with `vertical="false"` and never resizes.

One more thing on the comment at line 231: `afterNextRender` runs after layout, and `measureDtWidth()` two lines above already calls `getBoundingClientRect()` and trusts the result — so "a host that has no box yet" is not what the flag is guarding. What it actually preserves is the 100 ms `debounceTime` deferral of the first layout decision, which the new spec has to `tick(100)` around.

packages/components/dl/dl.component.ts
```diff
@@ -137,7 +144,10 @@ export class KbqDlComponent {
     readonly horizontalAlign = input<KbqDlAlign>('start');
 
     /** Forces the vertical layout; `null` lets the list decide based on `verticalBreakpoint`. */
-    readonly vertical = input<boolean | null>(null);
+    readonly vertical = input<boolean | null, unknown>(null, {
+        // Not `booleanAttribute`: it would fold `null` — the "decide for me" state — into `false`.
+        transform: (value) => (value == null ? null : booleanAttribute(value))
```
lskramarov: **Fourth hand-rolled nullable-`booleanAttribute` in the library, third distinct null semantic.**

- `popover.component.ts:322` — `(value: unknown) => (value == null ? undefined : booleanAttribute(value))`
- `navbar-brand.component.ts:119` and `navbar-item.component.ts:530` — `(value: unknown) => (value === undefined ? undefined : booleanAttribute(value))` (so `null` folds to `false`)
- here — `(value) => (value == null ? null : booleanAttribute(value))`

Three different answers to "what does an unbound or null binding mean", which a reader has to diff by eye, and the new `optionalNumberAttribute` adds a fourth private copy for numbers (`code-block.ts:203` and `inline-edit.ts:206` work around the same missing helper and inherit the `NaN` hole). `packages/components/core/utils/` is re-exported through `core/public-api.ts` and is the obvious home for one `kbqNullableBooleanAttribute` and one optional-number helper, so an edge case gets fixed in one place instead of five.

Minor: the three existing copies annotate `(value: unknown)`; this one omits it.

packages/components/dl/dl.component.ts
artembelik: Логика из suggestion, но хелпер вынесен в `core/utils` как `kbqOptionalNumberAttribute` - по соседнему замечанию про четыре локальные копии.

packages/schematics/src/migrations/dl-attribute-coercion/data.ts
artembelik: Регулярка убрана целиком: схематик обходит AST шаблона, поэтому формы с `=` попадают в отчёт, а `class="my-list vertical"` и `<kbq-dl-other wide>` уже нет.

tools/public_api_guard/components/dl.api.md
artembelik: Взял `approve-public-api-any`: в золотых файлах 159 входов с `unknown` вторым параметром против 6 суженных, так что сужение развело бы `dl` со всем остальным репозиторием.

packages/components/dl/dl.component.ts
artembelik: Флаг, ветка `vertical() === null` и список зависимостей ушли, `autoVertical` стал `computed` над `hostWidth`; эффект и `Injector` остались под пересчёт зажима `dtWidth` из соседнего замечания.

packages/components/dl/dl.component.ts
artembelik: Слишком широкая часть ушла вместе с эффектом раскладки, слишком узкая закрыта отдельным эффектом на `dtMinWidth`/`ddMinWidth`, который заново зажимает `dtWidth`.

packages/components/dl/dl.component.ts
artembelik: Оба хелпера добавлены в `core/utils`, но три существующие копии не тронуты: у них три разные семантики `null`, и перевод на общий хелпер поменял бы поведение popover и navbar.

docs/guides/migration.en.md
artembelik: Фразу переписали на `main` до этой ветки, сюда исправление пришло мержем 2f62f2bdd - правил не я.

packages/components/dl/dl.component.ts
artembelik: Вариант с `computed`-видом не взял: наружу через `[(dtWidth)]` потребитель всё равно получал бы строку. Сделал aliased-вход с `transform` плюс `linkedSignal` и явный `dtWidthChange`, как у `softWrap` в code-block - `model()` не принимает `transform`.

lskramarov (approved):

PR #2003, opened 2026-09-07
fix(actions-panel)!: render the panel inside overlayContainer (#DS-5475)

artembelik:
## Problem

`overlayContainer` is documented as "the element the actions panel is displayed in", but it only anchored the positioning — the overlay itself stayed in the application-wide CDK container, because CDK resolves the overlay host from the single `OverlayContainer` its `Dialog` was injected with.

Every panel on a page therefore shared one container. On `/components/actions-panel/overview` that meant all six open panels lived in `body > .cdk-overlay-container`, and the docs app's `FullscreenOverlayContainer` — which relocates that container into `document.fullscreenElement` — dragged all of them into whichever example was expanded to fullscreen.

## Solution

A panel opened with `overlayContainer` now gets a `Dialog` of its own from a child `EnvironmentInjector`, backed by an `OverlayContainer` scoped to that element, and renders inside it.

- The child `Dialog` reaches the root one through `skipSelf` and shares its registry of open dialogs, so `closeAll()` and `afterAllClosed` keep seeing the panel.
- Positioning moves from `flexibleConnectedTo` to a global strategy resolved against the scoped container, which makes `syncOverlayMaxWidth` and its `ResizeObserver` unnecessary — the width now follows from `.cdk-overlay-pane`'s `max-width: 100%`.
- The scoped container gets a wrapper element of its own so that CDK's `aria-hidden` bookkeeping finds no siblings to hide inside the host. Before this, opening a panel marked the whole application `aria-hidden="true"`.
- A `static` host is promoted to `position: relative` while a panel is open. The promotion is reference-counted per host, because replacing an open panel creates the new overlay before the old one finishes closing.

## Breaking changes

- A panel opened with `overlayContainer` is rendered inside that element instead of `document.body`, so it is clipped by it and no longer floats above the page.
- `maxWidth` is no longer ignored when `overlayContainer` is set.
- Such a panel bypasses a globally provided `OverlayContainer`; the two settings both answer "where is the panel rendered", and the per-panel one wins.

## Also in this PR

- New `actions-panel-global` example showing the default behaviour — a panel pinned to the bottom of the viewport. Every existing example passed `overlayContainer`, so the default was not demonstrated anywhere.
- The "Managing OverlayContainer" docs section: four code snippets in it did not compile (`import { Inject }` while using `inject()`, a missing `KbqActionsPanelRef` import, `{ self: true }` without `providers: [KbqActionsPanel]`, and an `@NgModule({...})` with no class after the decorator).

## Testing

- Jest: 54 passing, including a scoped panel and a global one side by side under a custom global `OverlayContainer`, `Dialog.closeAll()` reaching a scoped panel, and the promotion surviving an overlapping replacement.
- Playwright in Docker: 8 passing. Five are new and assertion-based; the three committed screenshot baselines were regenerated with `tools/e2e/run.js` because the panel now aligns to the container's content box rather than its border box — a 1px shift.
- Public API unchanged (`check-api` clean); the scoped container is internal and not exported from the entry point.

## Known limitation

A panel opened *without* `overlayContainer` still causes CDK `Dialog` to mark every sibling of the global container `aria-hidden="true"`. The actions panel is non-modal, so this is wrong, but fixing it means working around a private CDK method and belongs in a separate change.


packages/components/actions-panel/e2e.playwright-spec.ts
```diff
@@ -33,6 +39,48 @@ test.describe('KbqActionsPanel', () => {
...
+
+        test('pins the panel to the bottom of the custom container', async ({ page }) => {
+            await page.goto('/E2eActionsPanelWithOverlayContainer');
+            const locator = getComponent(page);
+
+            await getOpenButton(locator).click();
+
+            const containerBox = (await getOverlayContainer(locator).boundingBox())!;
+            const panelBox = (await getOverlayContainer(locator).locator(actionsPanelOverlay).boundingBox())!;
+
```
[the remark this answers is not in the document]
artembelik: Fixed in a90c722e7 — both tests now await `toBeVisible()` and then a `boundingBoxAtRest()` helper that polls until the box stops changing. Worth noting the sharper failure mode: the panel slides up on open, so the risk was less a null box than a mid-animation one, where it still sits below its container and the containment assertion would fail.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

NikGurev (approved):
Такой функционал будет полезен для DS-2124

packages/components/actions-panel/actions-panel.ts
```diff
@@ -139,86 +150,86 @@ export class KbqActionsPanel implements OnDestroy {
...
-
-        return true;
+    /** Injector that makes CDK render the panel's overlay inside `host` instead of `document.body`. */
+    private createScopedInjector(host: ElementRef<HTMLElement>): EnvironmentInjector {
+        return createEnvironmentInjector(
+            [
+                // A `Dialog` of its own is what carries the scoped container into `createOverlayRef` — the root one
+                // resolves the application-wide container. It still reaches the root `Dialog` through `skipSelf` and
+                // shares its registry of open dialogs, so `closeAll()` and `afterAllClosed` keep seeing this panel.
+                Dialog,
```
lskramarov: **The scoped `Dialog` shares `openDialogs` with the root one but not `_ariaHiddenElements`, so `aria-hidden` is never taken off the application.**

In `@angular/cdk` 20.2.14 `Dialog` delegates the registry to the parent (`get openDialogs() { return this._parentDialog ? this._parentDialog.openDialogs : this._openDialogsAtThisLevel }`) but keeps `_ariaHiddenElements = new Map()` as a plain per-instance field. `_removeOpenDialog()` restores the marks from `this._ariaHiddenElements` when the *shared* list becomes empty, so whichever instance closes the last dialog is the one that has to own the map.

Sequence that breaks it:

1. a panel opens **without** `overlayContainer` — the root `Dialog` sees an empty registry and marks every sibling of `body > .cdk-overlay-container` with `aria-hidden="true"`, recording them in **its** map;
2. a panel opens **with** `overlayContainer` — registry length is 1, so nothing is hidden (correct);
3. the global panel closes — the **root** `_removeOpenDialog` sees length 1 and skips the restore;
4. the scoped panel closes — the **scoped** `_removeOpenDialog` sees length 0 and restores its own map, which is empty.

The root map is never drained, so the whole application stays `aria-hidden="true"`. It does not heal either: the next root dialog re-records the current value (`"true"`) as the *previous* value and restores it forever.

Verified with a Jest repro (two `KbqActionsPanel` hosts, `NoopAnimationsModule`), counting `body > [aria-hidden="true"]`:

```
PR head    a90c722e7   after global open: 1   after global close: 1   after scoped close: 1   FAIL
merge-base e6532c45f   after global open: 1   after global close: 1   after scoped close: 0   pass
```

The trigger is not limited to two actions panels: any CDK-`Dialog`-based overlay feeds the same root registry (Angular Material’s `MatDialog` injects the root `Dialog`), so "modal open → scoped actions panel inside it → close the modal first" reproduces it in a consumer app.

The `openDialogs` sharing that `closeAll()` / `afterAllClosed` rely on is exactly what desynchronises the aria bookkeeping, so a fix has to reunite the two — e.g. by having a scoped `Dialog` subclass adopt the root’s map:

```ts
@Injectable()
class KbqActionsPanelScopedDialog extends Dialog {
    // The registry of open dialogs is shared with the root `Dialog`, so the aria bookkeeping has to be too:
    // whichever instance closes the last dialog is the one that restores the marks.
    protected override _ariaHiddenElements = inject(Dialog, { skipSelf: true })['_ariaHiddenElements'];
}
```

Whatever shape it takes, a regression test for the global-first / scoped-last order would be worth adding: the current suite covers both panels open side by side, but never closes them in that order.

packages/components/actions-panel/actions-panel-scoped-overlay-container.ts
```diff
@@ -0,0 +1,149 @@
...
+     * removes every other container carrying one, which is bookkeeping for the single application-wide container. A
+     * scoped container is not that container, and joining the scheme would make it delete — and be deleted by — the
+     * global one as soon as both exist.
+     */
+    protected override _createContainer(): void {
+        const wrapper = this._document.createElement('div');
+
+        wrapper.classList.add(KBQ_ACTIONS_PANEL_SCOPED_OVERLAY_CONTAINER_SELECTOR);
+        wrapper.style.position = 'absolute';
+        wrapper.style.inset = '0';
```
lskramarov, line 75: **A scrollable `overlayContainer` scrolls the panel out of view instead of keeping it pinned.**

An absolutely positioned child of an `overflow: auto` element resolves against the padding box at scroll origin and then scrolls with the content, so `inset: 0` here pins the panel to the bottom of the *content*, not of the *scrollport*.

Measured in headless Chromium against the exact DOM and CSS this container builds (host `position: relative; overflow: auto; height: 200px`, 900px of content, CDK + Koobiq structural overlay rules):

```
scrollTop=0     hostBottom=222.0  paneBottom=221.0  visible=true
scrollTop=300   hostBottom=222.0  paneBottom=-79.0  visible=false
```

The strategy this replaces (`flexibleConnectedTo(overlayContainer)` + `RepositionScrollStrategy` in the viewport-fixed container) kept the panel on the container’s visible bottom edge while the container scrolled. The use case the docs single out — "for example, inside a [Sidebar](/en/components/sidebar)" — is precisely a container that scrolls, so this is worth either supporting (a `position: sticky` pane, or sizing the scoped container against the host’s client rect the way the old strategy did) or calling out as unsupported.

packages/components/actions-panel/actions-panel.en.md
```diff
@@ -75,38 +80,59 @@ If you are using a `TemplateRef` for your actions panel content, the `data` and
...
     openActionsPanel() {
-        const actionsPanelRef = this.actionsPanel.open(YourActionsPanelComponent, {
+        this.actionsPanel.open(YourActionsPanelComponent, {
             overlayContainer: this.customContainer
         });
     }
 }
 ```
 
+The panel is rendered inside that element, pinned to its bottom center, so it is clipped by the element and follows it
```
lskramarov: **Both halves of this sentence overstate what the new positioning does.**

*"follows it when the element scrolls"* — when the element is itself a scroll container the panel does not follow it, it scrolls away with the content. Measured on the DOM this builds: pane bottom 221 (flush with the host bottom at 222) at `scrollTop: 0`, pane bottom -79 after scrolling 300px inside the host.

*"it is clipped by the element"* — only when the element sets `overflow: hidden`. At the default `overflow: visible` nothing clips, and the `translateY(100%)` entrance/exit animation renders the panel below the container, over whatever follows it on the page. Every actions-panel docs example happens to clip (`actions-panel-adaptive` gets it from the parent’s `example-actions-panel { overflow: hidden }`), which hides this on the docs site but not from consumers.

The same sentence is in `actions-panel.ru.md:116` («следует за ним при прокрутке», «обрезается границами элемента»).

packages/components/actions-panel/actions-panel.spec.ts
```diff
@@ -522,29 +525,195 @@ describe(KbqActionsPanelModule.name, () => {
...
+        const hostElement = componentInstance.elementRef.nativeElement;
+
+        componentInstance.openFromTemplate({ overlayContainer: { nativeElement: hostElement } });
+
+        expect(hostElement.querySelector('.cdk-overlay-pane')).not.toBeNull();
+        expect(getOverlayContainerElement().querySelector('.cdk-overlay-pane')).toBeNull();
+    });
+
+    it('should not reach for the global custom OverlayContainer at all when overlayContainer is provided', () => {
+        const getContainerElement = jest.spyOn(FakeOverlayContainer.prototype, 'getContainerElement');
```
lskramarov, line 624: **The spy is restored only on the happy path.**

`getContainerElement.mockRestore()` is the last statement of the test, so if `expect(getContainerElement).not.toHaveBeenCalled()` ever fails, the throw skips it and the spy stays installed on `FakeOverlayContainer.prototype` for the rest of the file. `jest.config.js` sets `clearMocks: true` but not `restoreMocks`, and clearing only wipes recorded calls — it does not put the original method back.

The two tests that follow both use `FakeOverlayContainer`, so one real failure turns into a cascade. A `try` / `finally`, or moving the restore into an `afterEach`, keeps the failure local.

packages/components/actions-panel/actions-panel-scoped-overlay-container.ts
```diff
@@ -0,0 +1,149 @@
...
+ * application. Providing this container in a child injector — together with a `Dialog` of its own — is what makes a
+ * single overlay render inside a specific element while the rest of the application keeps using the global container.
+ *
+ * The container is scoped to one host and is not reusable: it is expected to be created and destroyed alongside the
+ * overlay it hosts.
+ *
+ * @docs-private
+ */
+@Injectable()
+export class KbqActionsPanelScopedOverlayContainer extends OverlayContainer {
```
lskramarov, line 45: **Consider `packages/components/core/overlay/` for this class.**

That directory already owns the repository’s other `OverlayContainer` subclass — `KbqShadowDomOverlayContainer` in `shadow-dom-overlay-container.ts`, which solves the same shape of problem (relocate the container somewhere other than `document.body`, then deal with the fallout) — alongside the shared `panel-width`, `panel-height` and scroll-strategy helpers.

Nothing in this container is actions-panel-specific except the token name and the `aria-hidden` rationale: it is the generic answer to "give this one overlay a container of its own". Leaving it in a component entry point means the next component that wants a scoped overlay either re-derives the wrapper / containing-block promotion / `platform`-attribute reasoning, or imports across entry points from `@koobiq/components/actions-panel`.

packages/components/actions-panel/actions-panel-config.ts
```diff
@@ -50,7 +50,13 @@ export class KbqActionsPanelConfig<D = unknown> {
     closeOnNavigation?: boolean = true;
 
     /**
-     * Element in which the actions panel overlay will be connected to.
+     * Element the actions panel overlay is rendered into, instead of the application-wide overlay container. The panel
```
lskramarov, line 53: **Breaking change without an `ng update` migration.**

`AGENTS.md` (repo root) requires that a breaking change to a component "ships with a migration there, or at least a warning that points at the manual steps", and this PR is titled `fix(actions-panel)!` with three documented breaking changes: the panel renders inside (and is clipped by, and scrolls with) `overlayContainer`, `maxWidth` is no longer ignored, and a globally provided `OverlayContainer` is bypassed.

None of it is codemoddable, but that is what the "or at least a warning" clause covers — a migration under `packages/schematics/src/migrations` that detects `overlayContainer:` in `KbqActionsPanel.open()` calls and prints the manual steps would give upgrading consumers the signal they otherwise only get from the changelog.

packages/components/actions-panel/actions-panel-scoped-overlay-container.ts
```diff
@@ -0,0 +1,149 @@
...
+        container.style.position = 'absolute';
+
+        // The container gets a wrapper of its own rather than being appended to the host directly, because CDK's
+        // `Dialog` marks every sibling of the overlay container with `aria-hidden` while a dialog is open. Inside the
+        // host those siblings are the very content the actions panel acts on, so the container is kept as an only
+        // child and there is nothing left for CDK to hide.
+        wrapper.appendChild(container);
+
+        this.promoteHostToContainingBlock();
+        this.host.appendChild(wrapper);
```
lskramarov, line 93: **Worth documenting alongside the `position` promotion: the host gains a child while the panel is open.**

The `static` → `relative` promotion is called out in the `overlayContainer` JSDoc and in both docs pages, but this line has the same "we mutate your element" character and is not mentioned anywhere. For as long as a panel is open the consumer’s element has one extra child node, so:

- `.list:empty { … }` stops matching;
- `.container > *:last-child { margin-bottom: 0 }` selects the invisible wrapper instead of the real last item;
- `:nth-last-child()` rules shift by one;
- `element.children.length` / `childElementCount` read one higher.

The package’s own tests cannot catch this because the e2e host is an empty `<div>`. A sentence next to the promotion note in `actions-panel-config.ts` and the two `.md` files would be enough.

lskramarov (commented):
## Code review (max effort)

7 findings, most severe first. The first is a confirmed regression against the merge-base and the only blocker in my view; the second is a confirmed behaviour change for scrollable containers.

Verification done locally against `a90c722e7`: `actions-panel.spec.ts` 54/54 passing, `check-e2e-types` clean, `prettier` / `eslint` / `cspell` clean on the touched files, and `build:docs-examples-module` reproduces the committed `example-module.ts` exactly.

packages/components/actions-panel/actions-panel.ts
artembelik: Перенял карту присваиванием в конструкторе: `protected override` на приватном члене базового класса не проходит проверку типов. Добавил регрессионный тест на порядок закрытия.

packages/components/actions-panel/actions-panel-scoped-overlay-container.ts
artembelik: Не делаем в этом PR: поддержку прокручиваемого контейнера вынесем в отдельную задачу.

packages/components/actions-panel/actions-panel-scoped-overlay-container.ts
artembelik: Оставляем в actions-panel: перенесём в core, когда scoped-контейнер понадобится другому компоненту.

lskramarov (approved):

PR #2006, opened 2026-09-08
feat(list): "select all" row in kbq-list-selection (#DS-4036)

lskramarov:
## Summary

`kbq-list-selection` had only the keyboard half of "select all": `Ctrl`/`Cmd` + `A`, `selectAllToggle`, `onSelectAll` and the shared `toggleSelectAll` / `getSelectAllState` helpers from `@koobiq/components/core`. There was no visible master checkbox, no `indeterminate` state, and `getSelectAllState` went unused.

This adds the row, bringing the list to parity with `kbq-select` and `kbq-tree-selection`. Both entry points — the row's click and the shortcut — funnel through a single `toggleSelectAll()`, so the checkbox and the keyboard can never disagree.

The attribute is written `selectAll`, exactly as on the two sibling components. The class member behind it is `selectAllEnabled` with `alias: 'selectAll'`, because `selectAll()` is already an existing public method; this keeps the template API identical across the family with no breaking change and no migration schematic. There is precedent for the alias in `KbqAccordion.valueInput`.

## List of notable changes:

- **added** the `selectAll` attribute to `kbq-list-selection`, rendering a tri-state master checkbox above the options. The label comes from the existing `select.selectAll` locale key, making the list a third reader of that section (JSDoc in `core/locales/{select,types}.ts` updated accordingly — no new strings, all five locales already carry it).
- **added** an internal `selectAllRow` marker input on `KbqListOption`, mirroring `KbqTreeOption`. The row is a real option, so it inherits the styling, focus handling and theming for free — no new SCSS is needed, matching `kbq-select__select-all` and `kbq-tree-option_select-all`.
- **added** `navigableOptions`, a separate query the key manager runs on with the row at its head. The public `options` content query stays free of synthetic entries, so `getSelectedOptionValues`, the drop-indicator geometry and every consumer tally keep meaning "the options the consumer projected".
- **added** `allOptionsSelected`, `selectAllState` and `toggleSelectAll()` to the public API, plus a `refreshSelectAllState()` hook (`@docs-private`) the options use to keep the row in step.
- **updated** the Ctrl/Cmd + A handler to delegate to `toggleSelectAll()`, removing the second, divergent copy of the batch logic.
- **added** a `list-select-all` docs example, an `E2eListSelectAllStates` scenario with new `03-{light,dark}.png` baselines, a `selectAll` playground in the dev app, and documentation in both `list.en.md` and `list.ru.md`.
- **fixed**, in passing, two dead `(onSelectAll)` bindings in the dev app that sat on single-selection lists where the shortcut never fires.

### Combinations the row is deliberately refused in

Keyboard and single selection, `horizontal`, an empty list, and inside a `cdk-virtual-scroll-viewport`. Under a virtual scroller the list only ever holds the options currently rendered, so a master checkbox built on them would report "everything selected" after touching a fraction of the data — the same reason `kbq-select` refuses the row under `withVirtualScroll`. That combination also logs a dev-mode warning. Detection uses `contentChild(CdkVirtualForOf)`, matching how `KbqSelect` does it, so it works under SSR and follows a viewport that appears later.

1. **The drag-and-drop interaction.** `KbqListOption` carries `hostDirectives: [CdkDrag]`, and `CdkDrag` registers itself with the enclosing `CdkDropList` from its own constructor. Left alone, the row would become drag item #0 and shift `previousIndex` in every `dropped` event by one, while `currentIndex` — derived from `options` — would not move. The row is therefore unregistered via the public `CdkDropList.removeItem()` and its `CdkDrag` explicitly disabled in `ngOnInit`; the constructor cannot do it, because a signal input still holds its default there. There is a regression test asserting the row stays out of `getSortedItems()`.

2. **The index-space migration.** The key manager now reports positions in `navigableOptions`, so `toggleFocusedOption`, `selectActiveOptions`, `listenToOptionsFocus`, `removeOptionFromList`, `isValidIndex`, `copyActiveOption` and `followMovedOption` were moved to that space. `updateTabIndex` and `focus` deliberately stayed on `options` — the row only renders when options exist, so the counts are always zero together, and `navigableOptions` is rebuilt a pass later. This is the part most worth a careful read; `kbq-select` solved the same problem the same way.

3. **`selectionChange` stays silent for a batch.** The event carries a single option and has no shape a batch could take, so a batch is reported through `onSelectAll` plus one value report to the form control — the contract Ctrl/Cmd + A has always had. Flagging in case the team wants `KbqListSelectionChange` to grow an `options` array later.

4. **Change-detection coupling.** `markForCheck()` from an option only marks its ancestors for traversal, not for re-evaluating their own bindings, so the row would paint a stale checkbox. The list marks itself on both routes that can change the batch state — the selection model and an option's `disabled`.

### Known gap, not addressed here

`toggleSelectAll` does not consult `canDeselectLast()`, so with `noUnselectLast` two activations can empty the selection. The same hole exists in `KbqTreeSelection.selectAllOptions`, so it is a family-wide gap rather than a regression — but this change makes it reachable with one click. Closing it needs a decision on which option survives a batch deselect, which felt out of scope here.

### Follow-up worth considering

The select-all block now exists in three near-identical copies (`select`, `tree-selection`, `list`): `selectAllState`, `allOptionsSelected`, `selectAllAdapter`, the template row and the "prepend the row to the key manager's list" mechanism. Rule of three is met; extracting the shared piece into `core/selection` would be a good separate PR across the three components.

## Verification

- 200 unit tests for the list (34 new), 5028 across all components
- 40 Playwright tests in Docker, baselines regenerated via `e2e:docker:update-snapshots`
- `check-api` + `approve-api components/list`; the `any` ratchet moved 7 → 9, which is exactly the two `InputSignalWithTransform<boolean, unknown>` signatures that `booleanAttribute` inputs always emit
- `eslint`, `stylelint`, `prettier`, `cspell`, `check-e2e-types`, and AOT builds of `components` and `dev-e2e`


rmnturov:
Такие тонны описания в доках — это перебор. Или прям все нужно? (у меня сильный скепсис)
Я пошел сокращать

lskramarov:
> Такие тонны описания в доках — это перебор. Или прям все нужно? (у меня сильный скепсис) Я пошел сокращать

@rmnturov хз, сейчас вроде бы это единственное место где описано поведение по "Выбрать все" со всеми нюансами.

Мне кажется сокращать нужно с самого примера, не понятно зачем там поиск, а он очень сильно раздувает пример и вызывает вопросики по поведению.

rmnturov:
> > Такие тонны описания в доках — это перебор. Или прям все нужно? (у меня сильный скепсис) Я пошел сокращать
> 
> @rmnturov хз, сейчас вроде бы это единственное место где описано поведение по "Выбрать все" со всеми нюансами.
> 
> Мне кажется сокращать нужно с самого примера, не понятно зачем там поиск, а он очень сильно раздувает пример и вызывает вопросики по поведению.

поиск иллюстрирует работу мастер-чекбокса при фильтрации

packages/components/list/list.ru.md
```diff
@@ -62,6 +62,59 @@
...
+отмеченный, когда выбраны все. Нажатие в промежуточном состоянии выбирает оставшиеся опции, а не снимает
+выделение. Подпись берётся из локали (`select.selectAll`).
+
+Отключённые опции игнорируются: их не выбирают и не снимают, и состояние чекбокса отражает только те опции,
+которые пользователь действительно может переключить. Вся пачка один раз отдаёт новое значение в форму и
+сообщает о себе событием `onSelectAll` с опциями, на которые смогла подействовать. `selectionChange` на пачку
+не приходит: событие несёт одну опцию, и у пачки нет подходящей для него формы — тот же контракт, что был у
+`Ctrl`/`Cmd` + `A` с самого начала.
+
+Строка участвует в клавиатурной навигации как первый элемент списка: до неё доводит `Home` или стрелка вверх,
```
NikGurev: Если выбрать все и потом зайти клавиатурной навигацией в список, то выберется второй элемент, а не строка "Выбрать все"

packages/components/list/list.ru.md
```diff
@@ -62,6 +62,59 @@
...
+`Space` и `Enter` переключают её, а переход в список по `Tab`, когда ничего не выбрано, приводит на неё, а не
+на первую опцию. Это команда, а не значение, поэтому она не участвует в том, что действует на значения: набор
+букв её пропускает, `Ctrl`/`Cmd` + `C` с неё ничего не копирует, и на ней нельзя закрепить `Shift`-диапазон.
+В перетаскиваемом списке она отображается, но не поднимается за курсором.
+
+Не поддерживается вместе с `multiple="keyboard"`, одиночным выбором, `horizontal`, пустым списком и
+`cdk-virtual-scroll-viewport` — во всех этих случаях строка не отображается. Под виртуальным скроллом список
```
NikGurev: по-хорошему бы потом продумать или дать пример, как выстраивать "выбрать все" с virtual scroll
Отдельной задачей

lskramarov: не факт что это нужно... в list виртуальный скролл полностью под контролем разработчика (в отличии от селекта или дерева) и в случае реализации выбрать все просто делается кнопка со своей логикой

packages/components/list/list.ru.md
lskramarov: Наблюдение верное, но само поведение менять не стал: правило входа общее для семейства. `kbq-list-selection`, `kbq-tree-selection` и `kbq-select` при входе ставят фокус на первую выбранную опцию и только при пустом выборе — на первый элемент навигации, которым и оказывается строка «Выбрать все». Так же устроен паттерн listbox в ARIA APG, а строка объявлена как `role="option"` с трёхпозиционным `aria-checked`, то есть для скринридера она действительно первая опция листбокса. Поменять это в одном списке значило бы разойтись с деревом и селектом.

Сломана была формулировка. Оговорка «когда ничего не выбрано» стояла двадцатым словом фразы, которая начиналась безусловным «участвует в клавиатурной навигации как первый элемент списка». И само правило входа не описано нигде: в разделе «Клавиатура» есть `Home`, `End`, `PageUp`, стрелки и набор букв, но не точка входа, поэтому единственное упоминание жило в абзаце про `selectAll` и читалось как свойство строки.

Поправил в 288b91def: правило входа теперь описано один раз в «Клавиатуре», а в абзаце про строку случай назван прямо — «после „выбрать всё“ вход в список приводит на первую опцию, а не на строку». Обе точки входа закрыл тестами; проверил мутацией, что каждый падает от поломки своей ветки `focus()`.

_Generated with Claude Code_

NikGurev (approved):

PR #2007, opened 2026-09-08
fix(file-upload,ellipsis-center): proper file item overflow (#DS-5271)

NikGurev:
Uploading a file with a long name didn't shorten the name — the row stretched instead and a horizontal scrollbar appeared.

The row was growing to fit the whole name rather than shrinking. That caused the scrollbar, and it also made the shortening logic measure the stretched width, so it trimmed the text to a size that was already off screen.

**Changes**

- **File upload** — the row can now shrink, so the name is shortened in the middle, keeps its extension, and the scrollbar is gone.
- **Ellipsis-center** — the tooltip and the shortening logic were overwriting each other's on/off state; now separate. `minVisibleLength` was a public setting that did nothing — it now works.
- **Breadcrumbs examples (3)** — same underlying problem, quietly broken. Fixed in the examples only, component untouched.
- **Dev app** — `yarn run dev:file-upload` didn't build; fixed.

**Testing** — new browser test and 4 unit tests, all confirmed to fail without the fix. Docker screenshot tests green, no reference images changed.


packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -107,29 +130,48 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
...
         });
 
         this.renderer.appendChild(this.elementRef.nativeElement, dataTextStart);
         this.renderer.appendChild(this.elementRef.nativeElement, dataTextEnd);
     }
+
+    private syncTooltipDisabled(): void {
+        // Assigns the protected field rather than going through the base setter, which would fold the
+        // derived value back into the bookkeeping that tracks what the consumer actually asked for.
+        this._disabled = this.disabled;
```
lskramarov: **correctness — dead write, and `forDisabledComponent` silently stops working**

`this._disabled = this.disabled;` has no effect any more. `_disabled` had exactly one reader — the base `get disabled()` — and this class now shadows it (`pop-up-trigger.ts:377` reaches it through the getter too). The comment above the line describes bookkeeping that no longer happens.

The knock-on is real: `tooltip.component.ts:691` enables a tooltip for a disabled wrapped control by writing `this._disabled = !disabled`, guarded by `explicitlyDisabled === undefined`. The override never touches the base's private `explicitlyDisabled`, so that guard stays open forever while the write itself is now ignored — `<span kbqEllipsisCenter [forDisabledComponent]="ctrl">` used to get a tooltip and no longer does.

Guard on `this.disabled` instead of the field, and decide explicitly what `forDisabledComponent` should mean here.

packages/components/file-upload/multiple-file-upload.component.scss
```diff
@@ -235,9 +242,15 @@
...
+            min-width: 0;
 
             .kbq-file-item__text {
-                width: 90%;
+                // Takes whatever is left between the icon and the size column. `overflow: hidden` clips the
+                // non-shrinkable tail span, sized from the directive's estimated `charWidth`, instead of
+                // letting an underestimate push into the size column.
+                flex: 1 1 auto;
+                min-width: 0;
+                overflow: hidden;
```
lskramarov: **correctness — the tail can be clipped, taking the extension with it**

The tail span is `flex: 1 0 auto` (shrink 0) and its length comes from an estimate: `lastCharsLength = clientWidth / 2 / charWidth` with a 7px default. When real glyphs are wider than that — Cyrillic, CJK, uppercase — the tail needs more than the cell has, `start` shrinks to 0, and `overflow: hidden` now swallows the overflow instead of letting it push the row.

At ~14px/char in a 150px cell: `lastCharsLength = 11` → tail ≈ 154px > 150px → the user sees a name with no extension and no ellipsis. Before this PR the same case was ugly but visible; now it is silent.

The new `Math.min`/`Math.max` clamp only guards `sliceIndex <= 0`, not this. The deeper fix is to measure the tail (or let it shrink from the left, `direction: rtl` + `text-overflow`) rather than estimating from `charWidth`.

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -107,29 +130,48 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
...
         this.renderer.appendChild(dataTextEnd, this.renderer.createText(end));
         setTimeout(() => {
-            this.disabled = this.elementRef.nativeElement.clientWidth > dataTextStart.scrollWidth;
+            this.truncated = this.elementRef.nativeElement.clientWidth < dataTextStart.scrollWidth;
 
-            if (this.disabled) {
-                start = '';
-                end = this._kbqEllipsisCenter;
-            } else {
+            if (this.truncated && this._kbqEllipsisCenter.length >= this.minVisibleLength()) {
```
lskramarov: **behaviour change — `minVisibleLength` was inert since the initial commit**

`minVisibleLength` has been declared-but-unused since `bebf3bbef`, so today *any* overflowing text gets the middle ellipsis. Making the input effective silently changes every consumer that leaves the default of 50: a 30-character label that does not fit now falls into the `else` branch and is end-truncated, i.e. the directive stops doing the thing it is named for.

Both in-repo call sites pass `10` and the breadcrumbs example passes `15`, so only external consumers are hit — which is exactly the case AGENTS.md covers: *"A breaking change to a component ships with a migration there, or at least a warning that points at the manual steps."* This wants a `!` in the PR title / a changelog note, and ideally a second look at the default.

Separately, the name now means the opposite of how it reads: "shortest text worth splitting", not "minimum number of characters kept visible". The JSDoc says so, but the input name will keep misleading.

packages/components/file-upload/multiple-file-upload.component.html
```diff
@@ -51,6 +51,7 @@
                                 <span
                                     class="kbq-file-item__text"
                                     [debounceInterval]="0"
```
lskramarov: **correctness — container-only resizes never re-split, and the stale tail is now clipped**

`KbqEllipsisCenterDirective` recomputes only on `(window:resize)`. `.kbq-file-upload__list` is `overflow-y: auto`: add files until the vertical scrollbar appears and every existing row loses ~15px, but `@for (… ; track file)` keeps the DOM, `[kbqEllipsisCenter]` does not change, and no window resize fires — so no `refresh()` runs. The names stay split for the old width, and with the `overflow: hidden` added in this PR the tail is now cut silently instead of overflowing visibly. Same for a splitter drag or a sidebar collapse.

`KbqTitleDirective` already solves exactly this with `inject(SharedResizeObserver).observe(this.parent)` (`title.directive.ts:231`), and its comment spells out why `window:resize` is not enough. Reusing it here would also make this per-consumer `[debounceInterval]="0"` unnecessary.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-head-items/breadcrumbs-truncate-head-items-example.ts
```diff
@@ -38,6 +38,13 @@ import { KbqBreadcrumbsModule } from '@koobiq/components/breadcrumbs';
                     white-space: nowrap;
                     text-overflow: ellipsis;
                 }
+
+                /* The wrapper above clips, but the node that actually holds the text is this one, and as an
+                   inline-block it shrink-to-fits to the full text width regardless. What overflows is then an
+                   atomic inline box rather than text, which text-overflow cannot put an ellipsis on. */
+                .kbq-button-text {
```
lskramarov: **altitude — this patches a symptom the example itself creates (same in the tail and center examples)**

`_button-base.scss:48-60` already ships the working contract: `.kbq-button-wrapper { display: flex; min-width: 0 }` and `.kbq-button-text { @include list.kbq-truncate-line(); display: inline-block; min-width: 0 }`, with `%kbq-button-clamped-to-container` giving the button `max-width: 100%`.

The example overrides the wrapper to `display: inline-block`, which removes the flex context that made `min-width: 0` effective — that override *is* the bug. Dropping it (together with the `overflow`/`white-space`/`text-overflow` trio the button already provides) restores truncation in one deletion, instead of layering a third rule on top of a broken one in three files with three copies of the same comment.

If `.kbq-button-text` genuinely needs `max-width: 100%` for non-flex parents, that belongs in `_button-base.scss` next to its existing `min-width: 0`, so the next consumer who overrides `.kbq-button-wrapper` does not rediscover this.

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -57,8 +62,26 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
...
+    private truncated = false;
+
     private resizeSubscription = Subscription.EMPTY;
 
+    /** Hint is suppressed by the consumer, or unnecessary because the whole text is already visible. */
+    override get disabled(): boolean {
+        return this.consumerDisabled || !this.truncated;
+    }
+
+    override set disabled(value: boolean) {
```
lskramarov: **reuse — this re-implements arbitration the base class already has**

`KbqTooltipTrigger` keeps `explicitlyDisabled` for precisely this purpose ("makes the input win over the state `forDisabledComponent` derives, instead of the two fighting over `_disabled`"), with `_disabled` as the effective value. `consumerDisabled` + `truncated` is a second copy of that idea inside the same object, and the shadowing getter disables the first one.

`KbqTitleDirective` has the identical need and solves it worse — it writes `this.disabled = !this.isOverflown` (`title.directive.ts:207/234/276`), so a consumer's `kbqTooltipDisabled` there is overwritten on the next hover.

A protected hook on the base (something like `setDerivedDisabled(value)` folding into `explicitlyDisabled ?? derived`) fixes both directives once and keeps `disabled` meaning one thing across the tooltip family. As it stands, reading `directive.disabled` back after writing `false` returns `true` whenever the text fits, which is a surprising public API.

packages/components/file-upload/e2e.playwright-spec.ts
```diff
@@ -48,6 +48,24 @@ test.describe('KbqFileUploadModule', () => {
...
+            const list = item.locator('.kbq-file-upload__list');
+
+            // `KbqEllipsisCenterDirective` splits the text in a macrotask, so the layout only settles a frame
+            // after the list itself is attached.
+            await expect.poll(() => list.evaluate((el) => el.scrollWidth <= el.clientWidth)).toBe(true);
+
+            // Without this the assertion above would also pass on a name that never rendered at all.
+            const start = item.locator('.kbq-ellipsis-center_data-text-start');
+
+            await expect.poll(() => start.evaluate((el) => el.scrollWidth > el.clientWidth)).toBe(true);
```
lskramarov: **test-coverage — neither assertion checks that the name was actually split**

`start.scrollWidth > start.clientWidth` is equally true when no split happened and the whole name sits in the start span. Combined with the first assertion (which `min-width: 0` alone satisfies), this stays green if the middle-ellipsis logic regresses — bump `minVisibleLength` above the name length, or break `sliceIndex` so the tail comes out empty, and nothing here fails.

The PR's actual promise is *"shortened in the middle, keeps its extension"*. Assert that too:

```ts
const end = item.locator('.kbq-ellipsis-center_data-text-end');

await expect(end).toHaveText(/\.pdf$/);
```

lskramarov (commented):
## Code review — `xhigh`

Reviewed `b3ed2bac` (`bugfix/DS-5271`) against `main`. The layout fix itself looks right: `min-width: 0` up the flex chain is the correct root cause for the stretching row, and the new e2e case pins it at the narrowest realistic width.

14 findings inline, most severe first:

1. `syncTooltipDisabled()` writes a field nothing reads, and `forDisabledComponent` silently stops working on this directive.
2. The non-shrinkable tail can now be clipped by the new `overflow: hidden`, losing the extension the feature exists to preserve.
3. `minVisibleLength` was inert since the initial commit; making it effective changes behaviour for every consumer on the default of 50.
4. Container-only resizes still never re-split — and the stale tail is now clipped instead of visibly overflowing.
5. The breadcrumbs `max-width: 100%` patches a symptom of the examples' own `display: inline-block` override.
6. `consumerDisabled`/`truncated` re-implements the base class's `explicitlyDisabled` arbitration.
7. The new Playwright case never asserts that the name was actually split.
8. `refresh()`'s `setTimeout` is still never cancelled, and it now calls `hide()`.
9. The `sliceIndex` clamp returns `0` when the text length is `<= 1`.
10. The single-file variant keeps the pre-fix layout and has no coverage.
11. No unit test reaches the new clamp.
12. The center example duplicates the `.kbq-ellipsis-center` block globally.
13. Dead members left in the two example files.
14. The new e2e comment's min-width justification is wrong for a selected list.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -107,29 +130,48 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
         this.renderer.appendChild(dataTextStart, this.renderer.createText(this._kbqEllipsisCenter));
         this.renderer.appendChild(dataTextEnd, this.renderer.createText(end));
         setTimeout(() => {
```
lskramarov: **correctness — the pending macrotask is never cancelled**

`refresh()` schedules this timer and nothing clears it. `ngOnDestroy` unsubscribes `resizeSubscription` and runs `super.ngOnDestroy()`, but a queued callback still fires afterwards — and it now does more than write text: `syncTooltipDisabled()` → `this.hide()` → `this.ngZone.run(...)` on a torn-down trigger, plus a `clientWidth` read on a detached node.

Reachable from the very component this PR fixes: remove a row from `@for (file of files; track file)` while a refresh is in flight.

Keeping the handle and clearing it in `ngOnDestroy` — and at the top of the next `refresh()` — also removes the second problem here: a stale callback measures the spans the newer `refresh()` already detached (`scrollWidth === 0`), flipping `truncated` to `false` and firing a spurious `hide()` before the good one lands.

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -107,29 +130,48 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
...
+            if (this.truncated && this._kbqEllipsisCenter.length >= this.minVisibleLength()) {
                 const averageCharWidth = this.charWidth();
                 const lastCharsLength = Math.round(this.elementRef.nativeElement.clientWidth / 2 / averageCharWidth);
-                const sliceIndex: number = Math.round(this._kbqEllipsisCenter.length - lastCharsLength);
+                // Clamped so an underestimated `charWidth` (e.g. wider glyphs than the 7px default assumes)
+                // cannot push the whole name into `end`, which has no `text-overflow` and does not shrink.
+                const sliceIndex = Math.min(
+                    Math.max(1, Math.round(this._kbqEllipsisCenter.length - lastCharsLength)),
+                    this._kbqEllipsisCenter.length - 1
+                );
```
lskramarov: **correctness — the clamp does not deliver what its comment promises**

`Math.min(Math.max(1, …), length - 1)` inverts when `length <= 1`: the upper bound wins, `sliceIndex` is `0`, `start` is empty and the whole text lands in the non-shrinkable `end` span — exactly the state the clamp exists to prevent. Reachable with `[minVisibleLength]="1"` (or `0`) and a one-character label that does not fit.

```suggestion
                const sliceIndex = Math.max(
                    1,
                    Math.min(
                        Math.round(this._kbqEllipsisCenter.length - lastCharsLength),
                        this._kbqEllipsisCenter.length - 1
                    )
                );
```

packages/components/file-upload/file-upload.spec.ts
```diff
@@ -874,10 +874,17 @@ describe(KbqSingleFileUploadComponent.name, () => {
...
     describe('with ellipsis in the center', () => {
+        afterEach(() => jest.restoreAllMocks());
+
         it('should add tooltip and ellipsis in the center for a file with a long name', fakeAsync(() => {
             component.disabled = false;
             fixture.detectChanges();
 
+            // jsdom lays nothing out, so `KbqEllipsisCenterDirective` would measure the name as fitting and
+            // suppress its hint. Both sides of its fit test have to be stubbed for the name to count as long.
+            jest.spyOn(Element.prototype, 'clientWidth', 'get').mockReturnValue(100);
```
lskramarov, line 888: **test-coverage — the single-file variant keeps the pre-fix layout and is not covered**

These stubs make the tooltip assertion work, but jsdom lays nothing out, so the test says nothing about layout — and the single variant did not get the fix. `single-file-upload.component.scss:124` still reads:

```scss
.kbq-file-item__text {
    width: 120px;
    flex-grow: 1;
}
```

no `min-width: 0`, no `overflow: hidden`, inside a `display: flex` wrapper whose items keep the default `min-width: auto` — structurally the same hazard the PR fixes one level up, with the same `[kbqEllipsisCenter]` + `[minVisibleLength]="10"` (`single-file-upload.component.html:27`).

Either confirm the single variant is fine at 320px and say so in the PR, or extend the fix and the new Playwright case to cover it.

packages/components/ellipsis-center/ellipsis-center.directive.spec.ts
```diff
@@ -63,23 +92,71 @@ describe(KbqEllipsisCenterDirective.name, () => {
...
 
         componentInstance.text = '123456789012345678901234567890123456789012345678901234567890';
         fixture.detectChanges();
 
-        const nativeElement: HTMLElement = directiveDebugElement.nativeElement;
+        const { start, end } = refreshAt(fixture, 150, 420);
+
+        expect(end).not.toBe('');
+        expect(start + end).toBe(componentInstance.text);
+        expect(start.length + end.length).toBeLessThanOrEqual(componentInstance.text.length);
```
lskramarov, line 98: **test-coverage — the new clamp is untested**

The `Math.min`/`Math.max` clamp is the only genuinely new arithmetic in `refresh()`, and every case added here exercises either the `minVisibleLength` gate or the `disabled` split. With `charWidth = 7`, `clientWidth = 150` and a 60-character string, `lastCharsLength` is 11 — `sliceIndex` never approaches either bound, so the whole clamp could be deleted without turning a test red.

It is reachable from this fixture: set `charWidth` to `1` (or widen the host relative to the text) so `lastCharsLength >= length`, then assert `start` is non-empty and `end` is not the whole string.

A boundary case for `length === minVisibleLength()` is missing too — `>=` vs `>` is currently untested in either direction.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-center-items/breadcrumbs-truncate-center-items-example.ts
```diff
@@ -52,6 +52,13 @@ import { KbqToolTipModule } from '@koobiq/components/tooltip';
                     white-space: nowrap;
                     text-overflow: ellipsis;
                 }
+
+                /* The wrapper above clips, but the node that actually holds the text is this one, and as an
+                   inline-block it shrink-to-fits to the full text width regardless. kbqEllipsisCenter would
+                   then measure that inflated width and conclude the text fits, leaving it unsplit. */
+                .kbq-button-text {
```
lskramarov: **reuse — the `.kbq-ellipsis-center` block below is a third copy, and it leaks**

Lines 65-81 duplicate `file-upload.scss:21-36` verbatim. The directive ships no styles of its own, so every consumer copies the same six declarations, and any drift between the copies changes what the directive does per host.

Because this component is `ViewEncapsulation.None` and the selector is not prefixed, this copy also applies to any `kbqEllipsisCenter` rendered elsewhere on the same docs page — including the file-upload examples, whose own copy it then competes with.

The two spans are created by the directive, so their layout contract belongs in an `ellipsis-center.scss` forwarded from `packages/components/_index.scss`. That is also where this PR's `overflow`/`min-width` work would live once instead of twice.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-tail-items/breadcrumbs-truncate-tail-items-example.ts
```diff
@@ -40,6 +40,13 @@ import { PopUpPlacements } from '@koobiq/components/core';
                     text-overflow: ellipsis;
                     direction: rtl;
                 }
+
+                /* The wrapper above clips, but the node that actually holds the text is this one, and as an
+                   inline-block it shrink-to-fits to the full text width regardless. What overflows is then an
+                   atomic inline box rather than text, which text-overflow cannot put an ellipsis on. */
+                .kbq-button-text {
```
lskramarov, line 33: **cleanup — dead members in the two example files this PR edits**

- Here: `protected readonly PopUpPlacements = PopUpPlacements;` (line 62) and its `@koobiq/components/core` import are unreferenced by the template.
- `breadcrumbs-truncate-center-items-example.ts:87`: `breadcrumbs = ['branch', 'Users', 'Report dated 28.08.2018']` is unused — the template hardcodes the items, so editing that array looks like it changes the example and does not.

These render on the docs site as the canonical way to truncate a breadcrumb, so dead members get copied into consumer code.

packages/components/file-upload/e2e.ts
```diff
@@ -103,6 +103,19 @@ class CustomErrorStateMatcher implements ErrorStateMatcher {
                 }
             </table>
         </div>
+
+        <!--
+            Deliberately outside both tables: a td under table-layout auto grows to max-content, which hands
+            the row all the width it asks for and hides the very overflow this case exists to catch. 320px is
+            --kbq-file-upload-size-multiple-big-container-min-width, the narrowest the component renders at.
```
lskramarov: **comment accuracy — the stated constraint does not hold here**

`--kbq-file-upload-size-multiple-big-container-min-width` is applied under `&:not(.kbq-selected)` (`multiple-file-upload.component.scss:178`). This fixture always has a file, so the component carries no min-width at all in this block — 320px is pinned purely by the wrapper `div`, and the component would happily render narrower.

AGENTS.md: *"Update or remove comments that become outdated because of your changes."* As written, the next person narrowing this fixture will trust a floor that does not exist.

```suggestion
            the row all the width it asks for and hides the very overflow this case exists to catch. 320px is
            the design minimum the empty state renders at
            (--kbq-file-upload-size-multiple-big-container-min-width); once a file is present the component
            sets no minimum of its own, so the wrapper below is what pins the width.
```

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -57,8 +62,26 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
...
+
     private resizeSubscription = Subscription.EMPTY;
 
+    /** Hint is suppressed by the consumer, or unnecessary because the whole text is already visible. */
+    override get disabled(): boolean {
+        return this.consumerDisabled || !this.truncated;
+    }
+
+    override set disabled(value: boolean) {
+        this.consumerDisabled = coerceBooleanProperty(value);
```
artembelik: value это boolean судя по типам, зачем булеан дополнительно приводить к boolean типу?

artembelik:
пример бы добавить/обновить, чтобы это было видно

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Согласен, будет ошибка, если попытаться записать не boolean. Копипаст из tooltip

image

NikGurev:
> пример бы добавить/обновить, чтобы это было видно

e2e проверяет это, только без скриншота. Предлагаешь скриншот сделать?

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-center-items/breadcrumbs-truncate-center-items-example.ts
NikGurev: сделаю отдельным PR, не относится к задаче

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-head-items/breadcrumbs-truncate-head-items-example.ts
NikGurev: сделаю отдельным PR, не относится к задаче

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: сделаю отдельным PR, не относится к задаче

artembelik:
> Предлагаешь скриншот сделать?

да, необязательно всего компонента, достаточно только элемент с ellipsis

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-center-items/breadcrumbs-truncate-center-items-example.ts
NikGurev: Сделано в #2018

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: исправлено в #2018

NikGurev: Fixed in #2018. The directive no longer overrides `disabled`, so the base class is back in control and `forDisabledComponent` works again.

packages/components/file-upload/multiple-file-upload.component.scss
NikGurev: Fixed. The split point now comes from the width the text actually rendered at, instead of the 7px guess, so wide letters can no longer push the extension out of view. `charWidth` is kept as a lower bound. The new browser test uses a Cyrillic name and checks the extension is still there and the tail is not cut off.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Fixed by changing the default to `0`. The input works now, but nothing changes for anyone who leaves it alone, so there is no breaking change and no migration.

packages/components/file-upload/multiple-file-upload.component.html
NikGurev: Fixed. The directive now watches its own element with the shared observer, the same way `kbq-title` does, so a row that only got narrower re-splits the name. `[debounceInterval]="0"` is gone, and there is a new browser test that resizes only the container.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-head-items/breadcrumbs-truncate-head-items-example.ts
NikGurev: Fixed. All three examples now set a width and nothing else — the button already handles truncation, and the wrapper override was what broke it.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Fixed in #2018. The base class now keeps both values and decides which one wins in one place; the directive only reports whether the text was truncated. `kbq-title` uses the same thing, so it no longer overwrites what the consumer asked for.

packages/components/file-upload/e2e.playwright-spec.ts
NikGurev: Fixed. The check now also asserts the extension survived and the tail is not clipped, and the same helper covers the single-file variant.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Fixed. The timer handle is kept and cleared both on destroy and at the start of the next refresh.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Fixed, took your suggestion.

packages/components/file-upload/file-upload.spec.ts
NikGurev: Fixed. The single-file variant now gets the same layout fix, and the new browser test covers it at 320px as well. The unit test now says outright that it proves nothing about layout and points at the browser test.

packages/components/ellipsis-center/ellipsis-center.directive.spec.ts
NikGurev: Fixed. Added tests for the clamp, for an underestimated `charWidth`, and for both sides of the `minVisibleLength` boundary.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-center-items/breadcrumbs-truncate-center-items-example.ts
NikGurev: Fixed in #2018 — the stylesheet now lives next to the directive and loads itself, so both copies are gone. In this PR the example's copy is at least scoped to the example, so it stops reaching the file upload examples on the same page.

packages/docs-examples/components/breadcrumbs/breadcrumbs-truncate-tail-items/breadcrumbs-truncate-tail-items-example.ts
NikGurev: Fixed, both removed.

packages/components/file-upload/e2e.ts
NikGurev: Fixed, took your wording.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Вся связанная логика уехала в KbqTooltipTrigger в #2018, оно было там до этого PR. 

В рамках репозитория будет ругаться, для `strictTemplates: true`

А вот за пределами ругаться не будет и сломает логику: 

`kbqTooltipDisabled="false"` -> вернет truthy.

artembelik (approved):

lskramarov (approved):

PR #2008, opened 2026-09-08
fix(modal): add modal caption as directive (#DS-5519)

NikGurev:
## Summary

Added missing directive, additional tests and updated example

Layout mirrors the existing modal html layout


packages/components/modal/modal.scss
```diff
@@ -87,11 +87,34 @@
...
+        padding: 0 var(--kbq-modal-size-header-padding-right) var(--kbq-modal-size-header-padding-vertical)
+            var(--kbq-modal-size-header-padding-left);
+    }
+
     & + .kbq-modal-body {
         padding-top: var(--kbq-modal-size-content-padding-top);
     }
 }
 
+.kbq-modal-header-caption {
```
lskramarov: **The caption only gets padding when a `.kbq-modal-header` is its immediate previous sibling.**

The only rule that gives `.kbq-modal-header-caption` any padding is `.kbq-modal-header + .kbq-modal-header-caption` (line 95). The standalone `.kbq-modal-header-caption` block below sets nothing but `display: block`.

Nothing enforces that a title is present: `KbqModalCaption` injects `KbqModalComponent`, not `KbqModalTitle`, and the content of a `kbqComponent` modal is author-written. So this compiles and renders with the caption text flush against the modal edges (0px instead of `--kbq-modal-size-header-padding-left` / `-right` / `-vertical`):

```html
<kbq-modal-caption>Caption</kbq-modal-caption>
<kbq-modal-body>...</kbq-modal-body>
```

Suggested shape: give `.kbq-modal-header-caption` the full header padding by default and zero out only `padding-top` under `.kbq-modal-header + &`, instead of hanging the whole padding off the sibling rule.

packages/components/modal/modal.scss
```diff
@@ -87,11 +87,34 @@
...
+    display: block;
+
+    & + .kbq-modal-body {
+        padding-top: var(--kbq-modal-size-content-padding-top);
+    }
+}
+
+// The shadow is cast by the bottom edge of the header, so a caption takes it over from the title.
+.kbq-modal-overflow-shadow-top:not(:has(+ .kbq-modal-header-caption)) {
```
lskramarov: **The shadow hand-off is order-dependent and silently produces two shadows when the order is reversed.**

`KbqModalTitle` and `KbqModalCaption` both add `.kbq-modal-overflow-shadow-top`, and this rule relies on `:not(:has(+ .kbq-modal-header-caption))` to cancel the title's copy. That only works when the caption is the *immediately following* element sibling.

With the order reversed — which nothing in the API prevents:

```html
<kbq-modal-caption>Caption</kbq-modal-caption>
<kbq-modal-title>Title</kbq-modal-title>
<kbq-modal-body>...</kbq-modal-body>
```

the caption is not followed by a caption, so it draws a shadow *above* the title, and the header is not followed by a caption either, so it draws a second one — two stacked shadows inside the header block, plus the caption has no padding (see the comment on line 105).

The same happens if anything else is placed between the two (a wrapper element, a component that renders an element host).

packages/components/modal/modal.directive.ts
```diff
@@ -45,6 +45,29 @@ export class KbqModalTitle {
     protected readonly a11yLocaleConfiguration = kbqInjectA11yLocaleConfiguration();
 }
 
+/**
+ * Caption of a manually composed modal (`kbqComponent`). Placed next to `KbqModalTitle`, it
+ * continues the header with additional context below the title. The caption is clamped to two
+ * lines, so the host carries the header paddings while the text is clamped inside it.
+ */
+@Component({
```
lskramarov, line 56: **Altitude: the caption is a sibling of the header here, but a child of it in the built-in modal — that mismatch is what forces the three compensating CSS rules.**

`modal.component.html` (`tplContentDefault`) renders the same two pieces as:

```html
<div class="kbq-modal-header">
  <div class="kbq-modal-header-content">
    <div class="kbq-modal-title" kbq-title>…</div>
    <div class="kbq-modal-caption" kbq-title>…</div>
  </div>
  <button class="kbq-modal-close">…</button>
</div>
```

Composing them as siblings instead means the PR has to re-derive that layout with three rules that only hold for one exact DOM order: `padding-bottom: 0` on the header, `padding` re-applied on the caption, and `:not(:has(+ …))` to cancel the header's shadow.

Projecting the caption into the header (e.g. `KbqModalTitle` content-projecting an optional `KbqModalCaption`, or a `kbq-modal-header` wrapper directive) removes all three rules, removes the ordering fragility flagged above, and makes the custom composition produce the *same* DOM as `KbqModalService.create`, so consumer CSS and the `.kbq-modal-header-content` hook behave identically in both modes.

packages/components/modal/modal.directive.ts
```diff
@@ -35,7 +35,7 @@ import { KbqModalComponent } from './modal.component';
     host: {
         class: 'kbq-modal-header',
         '[class.kbq-modal-header_closable]': 'modal.kbqClosable',
-        '[style.box-shadow]': 'modal.bodyOverflow().top ? "var(--kbq-shadow-overflow-normal-bottom)" : null'
+        '[class.kbq-modal-overflow-shadow-top]': 'modal.bodyOverflow().top'
```
lskramarov: **Two mechanisms for one concept now live in this file.**

The title (and the new caption) publish the shadow through a CSS class resolved in `modal.scss`, while `KbqModalFooter` (line 99) still writes it inline:

```ts
'[style.box-shadow]': 'modal.bodyOverflow().bottom ? "var(--kbq-shadow-overflow-normal-top)" : null'
```

They now behave differently for anyone overriding them: the footer's inline style needs `!important` to beat, the header's class rule (specificity 0-2-0) does not. Worth converting the footer to a matching `.kbq-modal-overflow-shadow-bottom` class in the same PR, so the modal has one story.

packages/components/modal/modal.directive.ts
```diff
@@ -35,7 +35,7 @@ import { KbqModalComponent } from './modal.component';
     host: {
         class: 'kbq-modal-header',
         '[class.kbq-modal-header_closable]': 'modal.kbqClosable',
```
lskramarov: **Dead host binding (pre-existing, but this `host` object is being edited here).**

`kbq-modal-header_closable` is not referenced by a single stylesheet, template or spec in the repo:

```
$ grep -rn "kbq-modal-header_closable" --include=*.scss --include=*.css --include=*.html --include=*.ts .
./packages/components/modal/modal.directive.ts:37
```

It costs a binding evaluation per change-detection pass and a class toggle in the DOM for nothing. Since the neighbouring line is already changing, this is a cheap one to drop.

packages/components/modal/e2e.playwright-spec.ts
```diff
@@ -125,40 +125,45 @@ test.describe('KbqModalModule', () => {
...
-                el.scrollTop = Math.floor((el.scrollHeight - el.clientHeight) / 2);
-            });
+            await openModal(page, 'e2eOpenModal');
+            await scrollBody(page, 'middle');
 
             await expect.poll(() => e2eHasOverflowShadow(page.locator('.kbq-modal-header'))).toBeTruthy();
             await expect.poll(() => e2eHasOverflowShadow(page.locator('.kbq-modal-footer'))).toBeTruthy();
         });
+
+        test('should show the header shadow on the caption after scrolling down', async ({ page }) => {
```
lskramarov: **The visual part of this change has no visual regression coverage.**

What the PR actually changes on screen is layout: `padding-bottom: 0` on the header, the header's vertical padding moved onto the caption, and the shadow relocated from the header's bottom edge to the caption's. This test asserts only two computed-style booleans, so a wrong padding token, a doubled gap, or a caption that ends up flush against the body would all pass.

Every other modal state in this file has a baseline (`01-light.png` … `03-light.png`). Adding `await expect(page.locator('.kbq-modal-container')).toHaveScreenshot('04-light.png')` (plus the `e2eEnableDarkTheme` dark variant) after the modal settles would cover the layout math the CSS rules encode. Baselines have to be produced with `yarn run e2e:docker:update-snapshots` or `/approve-snapshots`.

packages/components/modal/modal.spec.ts
```diff
@@ -601,6 +601,45 @@ describe('KbqModal', () => {
...
+
+            expect(header.textContent).toContain('Title');
+            expect(caption.textContent).toContain('Caption');
+            expect(header.nextElementSibling).toBe(caption);
+            expect(caption.querySelector('.kbq-modal-caption')).toBeTruthy();
+
+            flush();
+        }));
+
+        it('should mark both the title and the caption with the top overflow shadow class', fakeAsync(() => {
```
lskramarov: **This asserts the state the CSS deliberately neutralizes, so it gives false confidence.**

`.kbq-modal-header` does carry `kbq-modal-overflow-shadow-top`, but `modal.scss:114` cancels it via `:not(:has(+ .kbq-modal-header-caption))` precisely so the title renders *no* shadow when a caption is present. The test therefore locks in an intermediate DOM state and would keep passing if the suppression rule were deleted.

Also, the test name says "the title" while the assertion queries `.kbq-modal-header`.

jsdom does not apply the stylesheet, so the only thing that can verify the real behaviour is the Playwright test — which makes the missing screenshot baseline (see `e2e.playwright-spec.ts:161`) the load-bearing gap here.

packages/components/modal/modal.spec.ts
```diff
@@ -601,6 +601,45 @@ describe('KbqModal', () => {
...
+
+            const header = document.querySelector('.kbq-modal-header')!;
+            const caption = document.querySelector('.kbq-modal-header-caption')!;
+
+            expect(header.textContent).toContain('Title');
+            expect(caption.textContent).toContain('Caption');
+            expect(header.nextElementSibling).toBe(caption);
+            expect(caption.querySelector('.kbq-modal-caption')).toBeTruthy();
+
+            flush();
```
lskramarov: **These two tests leave the modal open, unlike every other modal test in this file.**

The `created by service` block closes explicitly, with a comment explaining why:

```ts
afterEach(fakeAsync(() => {
    // wait all openModals to be closed to clean up the ModalManager as it is globally static
    modalService.closeAll();
    fixture.detectChanges();
    tick(ANIMATION_DURATION * 2);
}));
```

Here the overlay stays attached and `flush()` runs while `KbqModalBody`'s `KbqScrollbarViewport` is live — the same viewport whose self-rescheduling `requestAnimationFrame` loop (`scrollbar.ts:70-72`) is called out at line 544 of this file as the reason `flush()` cannot drain the queue there. It passes today, but it is one scrollbar change away from a "flush failed after reaching the limit" failure, and `document.querySelector` in these tests assumes exactly one modal is in the document.

Closing the modal (or `discardPeriodicTasks()`) before the end of each test would match the rest of the file.

lskramarov (commented):
## Code review — `fix(modal): add modal caption as directive (#DS-5519)`

Reviewed the full diff (12 files) at extra-high effort. The layout math is right for the intended order — header bottom padding moves onto the caption, the title keeps its `3xs` gap, and the shadow moves to the caption — and the docs are updated in both languages. All CI is green.

The findings below are, in order: two order-dependence bugs in the new SCSS, the structural mismatch that causes them, and then cleanup/coverage/convention items. Nothing here blocks on its own, but findings 1-3 share one root cause and are worth resolving together.

| # | File | Issue |
| - | ---- | ----- |
| 1 | `modal.scss:105` | Caption has no padding unless a header is its immediate previous sibling |
| 2 | `modal.scss:114` | Reversed title/caption order yields two stacked shadows |
| 3 | `modal.directive.ts:53` | Caption is a sibling of the header; the built-in modal nests it inside |
| 4 | `modal.directive.ts:38` | Class-based shadow on title/caption vs inline style on the footer |
| 5 | `modal.directive.ts:37` | `kbq-modal-header_closable` is a dead binding |
| 6 | `e2e.playwright-spec.ts:161` | No screenshot baseline for the new layout |
| 7 | `modal.spec.ts:623` | Asserts the class the CSS deliberately cancels |
| 8 | `modal.spec.ts:620` | New tests leave the modal open, unlike the rest of the file |
| 9 | `e2e.ts:104` | `E2eModalFullCustomCaptionContent` duplicates its sibling |
| 10 | `modal.directive.ts:68` | Missing `readonly`; injects the host’s concrete class |
| 11 | `baseline.json:30` | `any` ratchet raised 22 → 24 to accommodate that field |
| 12 | `components-dev/modal/module.ts:16` | Import path inconsistent with the lines above |
| 13 | `e2e.playwright-spec.ts:133` | Helper extracted for one `describe` only |
| 14 | `modal.en.md:131` | Docs understate the ordering requirement |

_Generated with Claude Code_

packages/components/modal/e2e.ts
```diff
@@ -82,10 +82,34 @@ export class E2eModalFullCustomContent {
...
+            }
+        </kbq-modal-body>
+
+        <div kbq-modal-footer>
+            <button kbq-button>Ok</button>
+        </div>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class E2eModalFullCustomCaptionContent {
```
lskramarov: **`E2eModalFullCustomCaptionContent` is `E2eModalFullCustomContent` plus one line.**

Selector, imports, body loop, footer, `items` and `changeDetection` are identical; the only difference is the `<kbq-modal-caption>` element. Two copies of the same scenario drift apart as soon as one is touched.

An `input()`-driven flag on the existing component, or passing the caption text through `KBQ_MODAL_DATA`, keeps one component:

```ts
@if (caption()) {
    <kbq-modal-caption>{{ caption() }}</kbq-modal-caption>
}
```

packages/components/modal/modal.directive.ts
```diff
@@ -45,6 +45,29 @@ export class KbqModalTitle {
...
+        </div>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'kbq-modal-header-caption',
+        '[class.kbq-modal-overflow-shadow-top]': 'modal.bodyOverflow().top'
+    }
+})
+export class KbqModalCaption {
+    protected modal = inject(KbqModalComponent);
```
lskramarov: **Two CLAUDE.md rules on this line.**

`AGENTS.md` (loaded by `CLAUDE.md`):

> Prefer `readonly` where appropriate (e.g., signals, injections)

> When a projected child depends on its host component, inject a narrow `InjectionToken` scoped to the members it uses, not the host's concrete class.

`KbqModalCaption` is exactly a projected child depending on its host, and it needs one member (`bodyOverflow`). Injecting `KbqModalComponent` ties the new public class to the whole modal component and, because the field is `protected`, exports `KbqModalComponent<any, any>` into the public API surface.

`private readonly` fixes the `readonly` rule and keeps the field out of the API guard entirely (host bindings are not access-checked — `typeCheckHostBindings` is not set in `tsconfig.json`):

```suggestion
    private readonly modal = inject(KbqModalComponent);
```

The token extraction is the larger change; the neighbouring `KbqModalTitle`/`KbqModalFooter` have the same pattern, so it is reasonable to do it for all three at once rather than only here.

tools/check-public-api-any/baseline.json
```diff
@@ -27,7 +27,7 @@
     "link": 7,
     "list": 7,
     "loader-overlay": 1,
-    "modal": 22,
+    "modal": 24,
```
lskramarov: **The `any` ratchet is being moved in the wrong direction to accommodate new code.**

The +2 comes entirely from `protected modal: KbqModalComponent<any, any>;` on the new `KbqModalCaption` (see `tools/public_api_guard/components/modal.api.md`). The guard is documented as failing in both directions, so raising the baseline permanently blesses two new `any`s on a class that only ever reads `modal.bodyOverflow()`.

Making the field `private readonly` (see the comment on `modal.directive.ts:68`) removes it from the emitted `.d.ts` and lets this line stay at `22`.

packages/components-dev/modal/module.ts
```diff
@@ -13,6 +13,7 @@ import { ModalOverviewExample } from 'packages/docs-examples/components/modal/mo
 import { ModalScrollExample } from 'packages/docs-examples/components/modal/modal-scroll/modal-scroll-example';
 import { ModalSizesExample } from 'packages/docs-examples/components/modal/modal-sizes/modal-sizes-example';
 import { ModalTemplateExample } from 'packages/docs-examples/components/modal/modal-template/modal-template-example';
+import { ModalFullCustomExample } from '../../docs-examples/components/modal';
```
lskramarov: **Import path is inconsistent with the eight lines directly above it.**

Lines 8-15 all use the deep `packages/docs-examples/components/modal/<name>/<name>-example` form; this one reaches into the barrel with a relative path, which also drags `ModalExamplesModule` and all nine examples into the module graph.

```ts
import { ModalFullCustomExample } from 'packages/docs-examples/components/modal/modal-full-custom/modal-full-custom-example';
```

Note that `prettier`'s `organize-imports` will re-sort it between the `modal-component` and `modal-multiple` lines, so move it there rather than leaving it at the end.

packages/components/modal/e2e.playwright-spec.ts
```diff
@@ -125,40 +125,45 @@ test.describe('KbqModalModule', () => {
...
 
     test.describe('overflow shadow (full custom content)', () => {
-        test('should show footer shadow on init when body content overflows', async ({ page }) => {
+        const openModal = async (page: Page, testId: string) => {
             await page.setViewportSize({ width: 400, height: 350 });
             await page.goto('/E2eModalFullCustom');
-            await page.getByTestId('e2eOpenModal').click();
+            await page.getByTestId(testId).click();
             await page.locator('.kbq-modal-container').waitFor({ state: 'visible' });
+        };
```
lskramarov: **The extraction stops one `describe` short.**

`openModal` is exactly the boilerplate that the sibling `overflow shadow` describe (lines 89-125) still repeats verbatim three times — same four lines, same viewport, only the route differs — and `scrollBody` covers the two `evaluate` blocks it also duplicates.

Lifting both helpers to the outer `describe` and parameterising the route removes ~20 lines and keeps the two suites from drifting.

Minor: `scrollBody(page, 'middle')` uses a magic string inside a `number | 'middle'` union for a single call site; a second helper (`scrollBodyToMiddle`) reads better than the sentinel.

packages/components/modal/modal.en.md
```diff
@@ -128,7 +128,7 @@ Opening one modal dialog from another is incorrect. But if it happens (as an eco
 
 #### Fully custom content
 
-When a modal is opened via `KbqModalService.open` with a `kbqComponent`, use the `kbq-modal-title`, `kbq-modal-body`, and `kbq-modal-footer` directives to compose the dialog. They keep the standard layout and scroll shadows, so a long body still casts a shadow onto the header and footer, exactly as in a modal created via `KbqModalService.create`.
+When a modal is opened via `KbqModalService.open` with a `kbqComponent`, use the `kbq-modal-title`, `kbq-modal-caption`, `kbq-modal-body`, and `kbq-modal-footer` directives to compose the dialog. `kbq-modal-caption` is placed next to `kbq-modal-title` and continues the dialog header. They keep the standard layout and scroll shadows, so a long body still casts a shadow onto the header and footer, exactly as in a modal created via `KbqModalService.create`.
```
lskramarov: **"is placed next to `kbq-modal-title`" understates a hard requirement.**

The CSS in `modal.scss` needs the caption to be the *immediately following element sibling* of the title: without that, it loses all its padding, and the title keeps a shadow the caption also draws (see the comments on `modal.scss:105` and `:114`). "Next to" reads as a layout hint, not a constraint an author can violate.

Worth stating explicitly that `kbq-modal-caption` must directly follow `kbq-modal-title` — or, better, removing the constraint per the comment on `modal.directive.ts:53`. The same sentence in `modal.ru.md:131` ("располагается рядом с") has the same problem.

packages/components/modal/modal.scss
NikGurev: Поправил. Подзаголовок теперь выводится внутри шапки, рядом с заголовком, а не отдельным блоком под ней — отступы он получает от самой шапки. Правило «соседнего элемента» стало не нужно, стили модалки вернулись к тому виду, что был до этой задачи.

packages/components/modal/modal.scss
NikGurev: Поправил. Тень при прокрутке снова рисует только шапка. Подзаголовок лежит внутри неё, поэтому двух теней и зависимости от порядка элементов больше нет.

packages/components/modal/modal.directive.ts
NikGurev: Согласен, сделал именно так. Подзаголовок теперь размещается внутри kbq-modal-title и попадает в ту же обёртку .kbq-modal-header-content, что и в окне, созданном через KbqModalService.create. Вёрстка стала одинаковой в обоих случаях, а три дополнительных правила в стилях удалось убрать совсем.

packages/components/modal/modal.directive.ts
NikGurev: Поправил, но в обратную сторону: вернул шапке обычный стиль через box-shadow — так же, как у футера и у директив KbqOverflowShadowTop/Bottom. Отдельный класс нужен был только чтобы передать тень подзаголовку, теперь в этом нет необходимости. Способ остался один на весь компонент.

packages/components/modal/modal.directive.ts
NikGurev: Убрал.

packages/components/modal/e2e.playwright-spec.ts
NikGurev: Добавил скриншотные тесты для светлой и тёмной темы (04-light.png / 04-dark.png), эталоны сгенерировал в Docker. Прогнал их отдельно, без повторных попыток — проходят стабильно.

packages/components/modal/modal.spec.ts
NikGurev: Поправил. Класса больше нет, поэтому тест переписан: теперь он проверяет, что подзаголовок действительно оказался в шапке под заголовком, а тень появляется у шапки. Внешний вид закрыт скриншотами.

packages/components/modal/modal.spec.ts
NikGurev: Поправил: оба теста теперь закрывают окно в конце, как и остальные тесты в файле.

packages/components/modal/e2e.ts
NikGurev: Объединил в один компонент — текст подзаголовка передаётся через KBQ_MODAL_DATA, второй компонент удалил.

packages/components/modal/modal.directive.ts
NikGurev: Подзаголовку больше вообще не нужен доступ к окну: тень рисует шапка, поэтому обращение к KbqModalComponent убрал целиком. У KbqModalTitle добавил readonly.

tools/check-public-api-any/baseline.json
NikGurev: Вернул 22. Благодаря правке выше новых any в публичном API не появилось.

packages/components-dev/modal/module.ts
NikGurev: Поправил на полный путь, как в соседних строках.

packages/components/modal/e2e.playwright-spec.ts
NikGurev: Вынес обе вспомогательные функции на верхний уровень файла, теперь ими пользуются оба блока тестов. Вместо строки 'middle' сделал отдельную функцию scrollBodyToMiddle.

packages/components/modal/modal.en.md
NikGurev: Переписал в обоих языках: теперь прямо сказано, что kbq-modal-caption нужно размещать внутри kbq-modal-title.

packages/components-dev/modal/module.ts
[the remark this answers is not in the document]
NikGurev: Поправил на полный путь, как в соседних строках.

lskramarov (approved):

PR #2009, opened 2026-09-08
fix: prerender every documentation example under SSR (#DS-5467)

artembelik:
## Summary

`packages/components-dev/ssr/routes.ts` excluded 97 examples from the SSR dev app, which is the only place CI exercises prerendering (`yarn run ssr:build`). Three defects were behind every exclusion except AG Grid; all three are fixed here, so the app now prerenders all 568 non-AG-Grid examples instead of 490.

The stated reasons for the exclusions were mostly symptoms of one root cause. `filter-bar.types.ts` listed the concrete pipe components in `defaultFilterBarPipes`, while every pipe imports that same module — an import cycle. Entered through the types module, as the prerender bundle enters it, the cycle evaluated a pipe before `KbqBasePipe` existed and threw `TypeError: Class extends value undefined` at chunk load. That killed the prerender worker and, with it, every route batched into that worker — which is why `accordion`, `username` and `tree-select` were also on the list.

Only the AG Grid entries remain excluded: the library has no server-side rendering of its own.

## List of notable changes:

- **added** `packages/components/filter-bar/filter-bar-pipes.ts` holding `defaultFilterBarPipes`, `KBQ_FILTER_BAR_PIPES` and `kbqFilterBarPipesProvider`, so `filter-bar.types.ts` keeps no runtime import inside the entry point and the cycle is gone. All three symbols stay exported from `@koobiq/components/filter-bar`; `check-api` confirms the public surface is unchanged.
- **updated** `kbqGetElementHeight` in `@koobiq/components/core` to call `getClientRects` defensively. Angular 20 ships a server DOM that does not implement the method at all, so the call threw instead of reporting "no box" as its doc comment promised. This is what `KbqTreeSelection.getHeight()` hit under SSR.
- **updated** the checkbox indeterminate example: it wrapped `kbq-checkbox` (which renders `div`s) in `<p>`. The HTML parser closes the paragraph at the first `<div>`, so the browser's DOM never matched the server's and hydration failed with NG0500 — the child checkboxes rendered without their boxes on every prerendered page, koobiq.io included. Rewritten on signals along the way: `ChangeDetectorRef` and three `detectChanges()` calls are gone, the parent's state is now `computed` from a single `signal`, and the content follows the security vocabulary the other examples use.
- **removed** the four TODO exclusions from `SSR_EXCLUDED_EXAMPLE_IDS` / `SSR_EXCLUDED_IMPORT_PATHS`. The `popover-scrolling-and-layering` entry turned out to be stale: the `**` fallback route already covers the `/examples/popover` iframe, and the route prerenders and hydrates cleanly.

- **The pipe registry move.** `KBQ_FILTER_BAR_PIPES` and friends changed file but not entry point, so consumers importing from `@koobiq/components/filter-bar` are unaffected and the golden API files did not move. Worth confirming that splitting the file is the shape you want, versus keeping the registry in `filter-bar.types.ts` and breaking the cycle some other way.
- **The checkbox example's look.** Replacing `<p>` with `layout-column` on the host drops the paragraph's default margins, and the children now carry a `margin-left` so the parent/child relation stays readable. This changes how the example renders on the docs site — deliberately, since the current rendering there is broken.
- **What is still excluded and why.** The `KbqSidepanelService` workaround in `packages/components-dev/ssr/config.ts` is untouched: I rebuilt without it and 377 routes still fail with NG0201, so it is not a symptom of the import cycle and remains an open part of #DS-5467.

Verification, beyond the CI gates: I served the production build and loaded all 568 prerendered routes through Playwright, collecting console and page errors. Zero routes report an error; before this change, `checkbox-indeterminate` was the only one that did once the build got far enough to produce pages at all.

### DIFF
--- a/packages/components-dev/ssr/routes.ts
+++ b/packages/components-dev/ssr/routes.ts
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
 import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
 
 const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
-    // TODO: NG0500: During hydration Angular expected a comment node but found <p>. (#DS-5467)
-    'checkbox-indeterminate',
-    // TODO: KbqTreeSelection.getHeight() calls getClientRects(), absent under SSR. (#DS-5467)
-    'tree-select-select-all',
-    'tree-select-select-all-label',
     // AG Grid does not support server-side rendering.
-    'content-panel-with-grid',
-    // TODO: NG04002: Cannot match any routes. URL Segment: 'examples/popover'. (#DS-5467)
-    'popover-scrolling-and-layering'
+    'content-panel-with-grid'
 ]);
 const SSR_EXCLUDED_IMPORT_PATHS = new Set([
-    // TODO: Restore after fixing SSR errors in accordion examples. (#DS-5467)
-    'components/accordion',
     // AG Grid does not support server-side rendering.
-    'components/ag-grid',
-    // TODO: Restore after breaking the circular dependency in filter-bar. (#DS-5467)
-    'components/filter-bar',
-    // TODO: Restore with filter-bar; this barrel exports UsernameFilterBarOptionExample. (#DS-5467)
-    'components/username'
+    'components/ag-grid'
 ]);
 
 /**
--- a/packages/components-dev/ssr/routes.ts
+++ b/packages/components-dev/ssr/routes.ts
@@ -4,7 +4,12 @@ import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-mod
 
 const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
     // AG Grid does not support server-side rendering.
-    'content-panel-with-grid'
+    'content-panel-with-grid',
+    // Both examples are a bare `<iframe src="/examples/<name>">`, a URL that only the docs app routes.
+    // Here they fall through to `**` and render an unrelated example, so prerendering them proves
+    // nothing about the popover or the select. Restore once this app serves those routes itself.
+    'popover-scrolling-and-layering',
+    'select-scrolling-and-layering'
 ]);
 const SSR_EXCLUDED_IMPORT_PATHS = new Set([
     // AG Grid does not support server-side rendering.
--- a/packages/components-dev/ssr/routes.ts
+++ b/packages/components-dev/ssr/routes.ts
@@ -2,6 +2,10 @@ import { Type } from '@angular/core';
 import { Routes } from '@angular/router';
 import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
 
+// TODO: Removing an entry here is not proof that the example survives SSR: `ssr:build` only prerenders,
+// while a hydration mismatch is thrown by the browser, after its parser has restructured the server
+// markup. A browser run over `devSsrExampleIds` would make this list a gate rather than a record.
+// (#DS-5539)
 const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
     // AG Grid does not support server-side rendering.
     'content-panel-with-grid',
--- a/packages/components/core/utils/dom.ts
+++ b/packages/components/core/utils/dom.ts
@@ -36,10 +36,11 @@ export const kbqFocusAndReveal = (element: HTMLElement, skipReveal = false): voi
 /**
  * Rendered height of an element, or `0` when it has no box.
  *
- * `getClientRects()` returns an empty list on the server and for elements that are not laid out, so
- * the first rect is read defensively rather than through `getBoundingClientRect()`, whose zeroes are
- * indistinguishable from a genuinely collapsed element.
+ * `getClientRects()` returns an empty list for elements that are not laid out and is missing entirely
+ * from the server DOM, so both the method and the first rect are read defensively rather than going
+ * through `getBoundingClientRect()`, whose zeroes are indistinguishable from a genuinely collapsed
+ * element.
  */
 export const kbqGetElementHeight = (element: Element): number => {
-    return element.getClientRects()[0]?.height ?? 0;
+    return element.getClientRects?.()[0]?.height ?? 0;
 };
--- a/packages/components/core/utils/dom.ts
+++ b/packages/components/core/utils/dom.ts
@@ -42,5 +42,5 @@ export const kbqFocusAndReveal = (element: HTMLElement, skipReveal = false): voi
  * element.
  */
 export const kbqGetElementHeight = (element: Element): number => {
-    return element.getClientRects?.()[0]?.height ?? 0;
+    return element.getClientRects?.()?.[0]?.height ?? 0;
 };
--- a/packages/components/filter-bar/filter-bar-pipes.ts
+++ b/packages/components/filter-bar/filter-bar-pipes.ts
@@ -24,9 +24,14 @@ export const defaultFilterBarPipes: [KbqPipeType, Type<KbqBasePipe<unknown>>][] 
     [KbqPipeTypes.Datetime, KbqPipeDatetimeComponent]
 ];
 
-/** Injection Token for providing pipes in filter-bar */
+/**
+ * Injection Token for providing pipes in filter-bar. Defaults to {@link defaultFilterBarPipes} so the
+ * standalone `KbqFilterBar` works without `KbqFilterBarModule`; override it with
+ * {@link kbqFilterBarPipesProvider} or a provider of your own.
+ */
 export const KBQ_FILTER_BAR_PIPES = new InjectionToken<Map<KbqPipeType, Type<KbqBasePipe<unknown>>>>(
-    'kbq-filter-bar-pipes'
+    'kbq-filter-bar-pipes',
+    { providedIn: 'root', factory: () => new Map(defaultFilterBarPipes) }
 );
 
 /** Utility provider for `KBQ_FILTER_BAR_PIPES`. */
--- a/packages/components/filter-bar/filter-bar-pipes.ts
+++ b/packages/components/filter-bar/filter-bar-pipes.ts
@@ -1,0 +1,38 @@
+import { InjectionToken, Provider, Type } from '@angular/core';
+import { KbqPipeType, KbqPipeTypes } from './filter-bar.types';
+import type { KbqBasePipe } from './pipes/base-pipe';
+import { KbqPipeDateComponent } from './pipes/pipe-date';
+import { KbqPipeDatetimeComponent } from './pipes/pipe-datetime';
+import { KbqPipeInputComponent } from './pipes/pipe-input';
+import { KbqPipeMultiSelectComponent } from './pipes/pipe-multi-select';
+import { KbqPipeMultiTreeSelectComponent } from './pipes/pipe-multi-tree-select';
+import { KbqPipeReadonlyComponent } from './pipes/pipe-readonly';
+import { KbqPipeSelectComponent } from './pipes/pipe-select';
+import { KbqPipeTextComponent } from './pipes/pipe-text';
+import { KbqPipeTreeSelectComponent } from './pipes/pipe-tree-select';
+
+/** list of pipes available out of the box. */
+export const defaultFilterBarPipes: [KbqPipeType, Type<KbqBasePipe<unknown>>][] = [
+    [KbqPipeTypes.ReadOnly, KbqPipeReadonlyComponent],
+    [KbqPipeTypes.Text, KbqPipeTextComponent],
+    [KbqPipeTypes.Input, KbqPipeInputComponent],
+    [KbqPipeTypes.Select, KbqPipeSelectComponent],
+    [KbqPipeTypes.TreeSelect, KbqPipeTreeSelectComponent],
+    [KbqPipeTypes.MultiSelect, KbqPipeMultiSelectComponent],
+    [KbqPipeTypes.MultiTreeSelect, KbqPipeMultiTreeSelectComponent],
+    [KbqPipeTypes.Date, KbqPipeDateComponent],
+    [KbqPipeTypes.Datetime, KbqPipeDatetimeComponent]
+];
+
+/** Injection Token for providing pipes in filter-bar */
+export const KBQ_FILTER_BAR_PIPES = new InjectionToken<Map<KbqPipeType, Type<KbqBasePipe<unknown>>>>(
+    'kbq-filter-bar-pipes'
+);
+
+/** Utility provider for `KBQ_FILTER_BAR_PIPES`. */
+export const kbqFilterBarPipesProvider = (): Provider => {
+    return {
+        provide: KBQ_FILTER_BAR_PIPES,
+        useValue: new Map<KbqPipeType, Type<KbqBasePipe<unknown>>>(defaultFilterBarPipes)
+    };
+};
--- a/packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
+++ b/packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
@@ -1,81 +1,57 @@
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, inject } from '@angular/core';
+import { ChangeDetectionStrategy, Component, computed, model } from '@angular/core';
 import { KbqCheckboxModule } from '@koobiq/components/checkbox';
-
-interface ICheckbox {
-    name: string;
-    checked: boolean;
-}
 
 /**
  * @title Checkbox indeterminate
  */
 @Component({
     selector: 'checkbox-indeterminate-example',
-    imports: [
-        KbqCheckboxModule
-    ],
+    imports: [KbqCheckboxModule],
     template: `
-        <div class="kbq-text-big">
-            <kbq-checkbox [checked]="parentChecked" [indeterminate]="parentIndeterminate" (change)="toggleChecked()">
-                All fruits
+        <kbq-checkbox [checked]="allSelected()" [indeterminate]="someSelected()" (change)="toggleAll()">
+            All event types
+        </kbq-checkbox>
+
+        @for (eventType of eventTypes; track eventType) {
+            <kbq-checkbox
+                class="example-checkbox-indeterminate__child"
+                [checked]="selected().has(eventType)"
+                (change)="toggle(eventType)"
+            >
+                {{ eventType }}
             </kbq-checkbox>
-            @for (fruit of fruits; track fruit) {
-                <p>
-                    <kbq-checkbox [checked]="fruit.checked" (change)="updateCheckboxes($index)">
-                        {{ fruit.name }}
-                    </kbq-checkbox>
-                </p>
-            }
-        </div>
+        }
     `,
-    changeDetection: ChangeDetectionStrategy.OnPush
+    styles: `
+        .example-checkbox-indeterminate__child {
+            margin-left: var(--kbq-size-l);
+        }
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'layout-column'
+    }
 })
 export class CheckboxIndeterminateExample {
-    private ref = inject(ChangeDetectorRef);
+    protected readonly eventTypes = ['Malware', 'Phishing', 'Ransomware'];
+    protected readonly selected = model(new Set([this.eventTypes[0]]));
 
-    parentIndeterminate = true;
-    parentChecked = true;
+    protected readonly allSelected = computed(() => this.selected().size === this.eventTypes.length);
+    protected readonly someSelected = computed(() => this.selected().size > 0 && !this.allSelected());
 
-    fruits: ICheckbox[] = [
-        { name: 'Apples', checked: true },
-        { name: 'Bananas', checked: false },
-        { name: 'Grapes', checked: false }
-    ];
-
-    updateCheckboxes(index: number) {
-        this.toggleFruitChecked(index);
-        this.updateIndeterminate();
-        this.ref.detectChanges();
+    protected toggleAll(): void {
+        this.selected.set(this.allSelected() ? new Set() : new Set(this.eventTypes));
     }
 
-    toggleFruitChecked(index: number) {
-        this.fruits[index].checked = !this.fruits[index].checked;
-    }
+    protected toggle(eventType: string): void {
+        this.selected.update((selected) => {
+            const next = new Set(selected);
 
-    toggleChecked() {
-        this.parentChecked = !this.parentChecked;
-
-        for (const fruit of this.fruits) {
-            fruit.checked = this.parentChecked;
-        }
-
-        this.parentIndeterminate = false;
-        this.ref.detectChanges();
-    }
-
-    updateIndeterminate() {
-        let checked: number = 0;
-        let unchecked: number = 0;
-        const length = this.fruits.length;
-
-        this.fruits.forEach((fruit) => {
-            if (fruit.checked) {
-                checked++;
-            } else {
-                unchecked++;
+            if (!next.delete(eventType)) {
+                next.add(eventType);
             }
+
+            return next;
         });
-        this.parentIndeterminate = checked !== length && unchecked !== length;
-        this.parentChecked = this.parentIndeterminate || length === checked;
     }
 }
--- a/packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
+++ b/packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, computed, model } from '@angular/core';
+import { ChangeDetectionStrategy, Component, computed, signal } from '@angular/core';
 import { KbqCheckboxModule } from '@koobiq/components/checkbox';
 
 /**
@@ -24,17 +24,18 @@ import { KbqCheckboxModule } from '@koobiq/components/checkbox';
     `,
     styles: `
         .example-checkbox-indeterminate__child {
-            margin-left: var(--kbq-size-l);
+            margin-top: var(--kbq-size-l);
+            margin-left: calc(var(--kbq-size-l) + var(--kbq-size-s));
         }
     `,
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
-        class: 'layout-column'
+        class: 'layout-column layout-align-start-start'
     }
 })
 export class CheckboxIndeterminateExample {
     protected readonly eventTypes = ['Malware', 'Phishing', 'Ransomware'];
-    protected readonly selected = model(new Set([this.eventTypes[0]]));
+    protected readonly selected = signal(new Set([this.eventTypes[0]]));
 
     protected readonly allSelected = computed(() => this.selected().size === this.eventTypes.length);
     protected readonly someSelected = computed(() => this.selected().size > 0 && !this.allSelected());
--- a/packages/components/accordion/accordion-content.directive.ts
+++ b/packages/components/accordion/accordion-content.directive.ts
@@ -1,3 +1,4 @@
+import { Platform } from '@angular/cdk/platform';
 import {
     afterNextRender,
     AfterRenderRef,
@@ -34,6 +35,7 @@ import { KbqAccordionItem } from './accordion-item';
 })
 export class KbqAccordionContentDirective implements AfterViewInit {
     private readonly renderer: Renderer2 = inject(Renderer2);
+    private readonly platform = inject(Platform);
 
     /** @docs-private */
     protected readonly nativeElement = kbqInjectNativeElement();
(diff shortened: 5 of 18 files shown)


packages/components/core/utils/dom.ts
```diff
@@ -36,10 +36,11 @@ export const kbqFocusAndReveal = (element: HTMLElement, skipReveal = false): voi
...
  *
- * `getClientRects()` returns an empty list on the server and for elements that are not laid out, so
- * the first rect is read defensively rather than through `getBoundingClientRect()`, whose zeroes are
- * indistinguishable from a genuinely collapsed element.
+ * `getClientRects()` returns an empty list for elements that are not laid out and is missing entirely
+ * from the server DOM, so both the method and the first rect are read defensively rather than going
+ * through `getBoundingClientRect()`, whose zeroes are indistinguishable from a genuinely collapsed
+ * element.
  */
 export const kbqGetElementHeight = (element: Element): number => {
```
lskramarov, line 44: **Returning `0` instead of throwing gives `KbqTreeSelection` a `NaN` page size.**

`tree-selection.component.ts:926` is the caller this change was made for:

```ts
updateScrollSize(): void {
    if (!this.renderedOptions.first) { return; }
    this.keyManager.withScrollSize(Math.floor(this.getHeight() / this.renderedOptions.first.getHeight()));
}
```

Both operands now return `0` on the server (`@angular/platform-server` defines `getClientRects` zero times in every bundle), so `Math.floor(0 / 0)` is `NaN` and `ListKeyManager.scrollSize` becomes `NaN`. The path is definitely reached: the constructor `effect()` at `tree-selection.component.ts:582` calls `updateRenderedOptions()` with no `platform.isBrowser` check, and that ends in `updateScrollSize()` at `:1341` — the removed TODO named this exact call, which also proves `keyManager` was already assigned. `setNextPageItemActive(delta = this.scrollSize)` then computes `activeIndex + NaN`, `items[NaN]` is `undefined`, and PageUp/PageDown silently move focus nowhere. Same hole in the browser via `tree-select.component.ts:1348`, which calls `updateScrollSize()` on overlay attach before the panel is laid out.

`KbqListSelection` already solves this, twice — `list-selection.component.ts:597` (`if (!this.platform.isBrowser) return;`) and `:662-668`:

```ts
const optionHeight = this.options.first.getHeight();

// `getHeight()` is 0 whenever the option is not laid out (SSR, jsdom, `display: none`);
// dividing by it would hand the key manager a `NaN` page size.
if (!optionHeight) { return; }
```

The tree has neither. Worth porting the zero-check (and, if you want the deeper fix, gating the constructor effect on `platform.isBrowser` — the component already injects `Platform` at `:195` and uses it at `:700`).

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
```diff
@@ -1,81 +1,57 @@
...
     `,
-    changeDetection: ChangeDetectionStrategy.OnPush
+    styles: `
+        .example-checkbox-indeterminate__child {
+            margin-left: var(--kbq-size-l);
+        }
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'layout-column'
```
lskramarov: **`layout-column` has no `gap`, so the four checkboxes now render flush against each other.**

`_layout.scss:138-149` defines it as `display: flex; flex-direction: column; min-width: 0; min-height: 0` — nothing else. The old markup got its rhythm from the `<p>` wrappers' UA `margin-block: 1em` (nothing in the repo resets `p`).

The only fallback would be the label's own vertical padding, and that is `0`: `checkbox.scss:37,39` read `var(--kbq-checkbox-size-normal-padding-top)` / `-padding-bottom` with no fallback value, and those two custom properties are defined **nowhere** — grep over the worktree and over `@koobiq/design-tokens` returns only those two reference lines. Both declarations are therefore invalid at computed-value time and fall back to the initial `0`. (Separate pre-existing bug worth its own ticket: every `kbq-checkbox` in the library has zero vertical padding for this reason.)

Net result on `/components/checkbox/overview`: the rows and their 16px hit targets touch. `pseudo-checkbox-example.ts:13-16` in the sibling folder shows the intended shape — `layout-column` plus `layout-margin-top-l` on each child.

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
```diff
@@ -1,81 +1,57 @@
...
+    }
```
lskramarov, line 34: **The flex host stretches each `kbq-checkbox` full-width, so the hover highlight fires far outside the clickable label.**

`.kbq-checkbox` is `display: inline-block; max-width: 100%` (`checkbox.scss:54-58`). As a flex item of a column container with the default `align-items: stretch` and no `width` / `align-self`, its used width becomes the whole container width.

The hover rule sits on that host: `_checkbox-theme.scss:89-91` applies `kbq-checkbox-color()` under `.kbq-checkbox`, and `:25-31` is `&:hover .kbq-checkbox__frame { border-color: …states-hover-border; background-color: …states-hover-background }`. The clickable region, meanwhile, is `.kbq-checkbox__layout` — an `inline-flex` label shrink-wrapped to its text (`checkbox.scss:31`). So on the docs site, moving the pointer anywhere across the empty right-hand side of the example paints the checkbox as hovered while a click there does nothing.

Before the change each `kbq-checkbox` was an `inline-block` in normal flow and shrink-wrapped, so hover and hit area coincided. `align-items: flex-start` on the host restores that.

packages/components/core/utils/dom.ts
```diff
@@ -36,10 +36,11 @@ export const kbqFocusAndReveal = (element: HTMLElement, skipReveal = false): voi
...
-    return element.getClientRects()[0]?.height ?? 0;
+    return element.getClientRects?.()[0]?.height ?? 0;
```
lskramarov: **The `?.` guards the method but not a nullish return — and the repo already has a spec for the case this misses.**

Optional chaining short-circuits only on a nullish base; once the call happens, `[0]` is a plain index read. Verified by execution:

| input | result |
| --- | --- |
| `{}` | `0` |
| `{ getClientRects: () => [] }` | `0` |
| `{ getClientRects: () => undefined }` | `TypeError: Cannot read properties of undefined (reading '0')` |

`KbqListSelection.getHeight()` (`list-selection.component.ts:803`, and `:1664`) is written the other way round — `getClientRects()?.[0]?.height ?? 0` — and `list-selection.component.spec.ts:1651-1657` pins exactly the third row:

```ts
jest.spyOn(listElement, 'getClientRects').mockReturnValue(undefined as unknown as DOMRectList);
expect(list.getHeight()).toBe(0);
```

So the codebase already treats a nullish return as worth testing, and after this change the shared helper is the only place that does not handle it. One character covers all three:

```suggestion
    return element.getClientRects?.()?.[0]?.height ?? 0;
```

lskramarov (commented):
Max-effort review of the SSR prerender fix. 15 findings, most severe first. The filter-bar cycle surgery itself is sound (`filter-bar.types.ts` is now a leaf, no new cycle is introduced, the API golden files genuinely need no regeneration, and the two `import type` switches are safe) — the findings cluster in three places instead: what the `dom.ts` guard turns the SSR crash into, what the rewritten checkbox example renders, and which exclusions were removed without a matching fix.

packages/components-dev/ssr/routes.ts
```diff
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
...
-    // TODO: NG0500: During hydration Angular expected a comment node but found <p>. (#DS-5467)
-    'checkbox-indeterminate',
-    // TODO: KbqTreeSelection.getHeight() calls getClientRects(), absent under SSR. (#DS-5467)
-    'tree-select-select-all',
-    'tree-select-select-all-label',
     // AG Grid does not support server-side rendering.
-    'content-panel-with-grid',
-    // TODO: NG04002: Cannot match any routes. URL Segment: 'examples/popover'. (#DS-5467)
-    'popover-scrolling-and-layering'
+    'content-panel-with-grid'
```
lskramarov: **`popover-scrolling-and-layering` is un-excluded with no fix — the `**` fallback the description credits already existed at the base commit.**

`git show bcca78370:packages/components-dev/ssr/routes.ts` contains `{ path: '**', redirectTo: devSsrExampleIds[0] }` verbatim, so it was already in place when the `NG04002: Cannot match any routes. URL Segment: 'examples/popover'` TODO was written. The PR changes neither the route table nor the example, so "the `**` fallback route already covers the `/examples/popover` iframe" means the exclusion was stale rather than that anything was fixed.

What the fallback actually does: the example is nothing but `<iframe src="/examples/popover">` (`popover-scrolling-and-layering-example.ts:9-10`), a URL that exists only in the docs app. Under `ssr:start` the iframe therefore loads `devSsrExampleIds[0]` — an accordion — and nothing about the popover is exercised. That is the opposite of the intent stated 7 lines below ("Fail the build instead of drifting silently").

Note `select-scrolling-and-layering` has the identical `<iframe src="/examples/select">` shape and was never excluded, which is presumably what made this entry look removable. Either add a real `examples/<name>` route so both iframes resolve, or keep both excluded.

packages/components-dev/ssr/routes.ts
```diff
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
...
 const SSR_EXCLUDED_IMPORT_PATHS = new Set([
-    // TODO: Restore after fixing SSR errors in accordion examples. (#DS-5467)
-    'components/accordion',
     // AG Grid does not support server-side rendering.
-    'components/ag-grid',
-    // TODO: Restore after breaking the circular dependency in filter-bar. (#DS-5467)
-    'components/filter-bar',
-    // TODO: Restore with filter-bar; this barrel exports UsernameFilterBarOptionExample. (#DS-5467)
-    'components/username'
+    'components/ag-grid'
```
lskramarov, line 20: **`components/accordion` is un-excluded but the diff touches no accordion file, and the accordion still has a concrete SSR defect.**

The removed line read `// TODO: Restore after fixing SSR errors in accordion examples. (#DS-5467)`. If an earlier commit fixed that failure, it would be worth naming it in the description — otherwise the exclusion is being dropped on the hypothesis that it was collateral damage from the killed prerender worker.

Here is one accordion SSR defect that is definitely still there: `accordion-content.directive.ts:74-76` calls `disableAnimation()` from `ngAfterViewInit`, which runs on the server, and `:89-92` does

```ts
this.savedTransition = this.nativeElement.style.transition;
this.nativeElement.style.transition = 'none';
```

The restore lives only in `afterNextRender` (`:63`), which that file's own comment says never runs on the server. Angular 20's server DOM is domino, whose `element.style` is a `Proxy` whose `set` trap forwards unknown properties to `target.setProperty(dasherize(prop), value)` (verified in `@angular/platform-server/fesm2022/init.mjs`), so the write reflects into the `style` attribute and is serialized into the prerendered HTML. On hydration the node is reused, the client's `ngAfterViewInit` reads `savedTransition = 'none'` back out of it, and `enableAnimation()` writes `'none'` again — `accordion.scss`'s `transition: height 300ms ease-out` is dead permanently. `KbqAccordionTrigger` has the identical pair at `accordion-trigger.ts:71/78/84`.

It does not throw, so `ssr:build` stays green either way.

packages/components/core/utils/dom.ts
```diff
@@ -36,10 +36,11 @@ export const kbqFocusAndReveal = (element: HTMLElement, skipReveal = false): voi
 /**
  * Rendered height of an element, or `0` when it has no box.
  *
- * `getClientRects()` returns an empty list on the server and for elements that are not laid out, so
- * the first rect is read defensively rather than through `getBoundingClientRect()`, whose zeroes are
- * indistinguishable from a genuinely collapsed element.
+ * `getClientRects()` returns an empty list for elements that are not laid out and is missing entirely
```
lskramarov, line 39: **The hardened helper reaches only 2 of the 5 sites that compute this, and one of the others has the same `NaN` bug.**

`grep -rn getClientRects packages/components --include=*.ts` (specs excluded):

| site | expression |
| --- | --- |
| `core/utils/dom.ts:45` | `getClientRects?.()[0]?.height ?? 0` (this PR) |
| `core/option/option.ts:297` | `getClientRects()[0]?.height ?? 0` — the exact pre-PR body, in `core` |
| `list/list-selection.component.ts:803` | `getClientRects()?.[0]?.height ?? 0` |
| `list/list-selection.component.ts:1664` | `getClientRects()?.[0]?.height ?? 0` |
| `select/select.component.ts:1961` | `getClientRects()[0]?.height` — **no `?? 0`** |

Only `tree/tree-option.component.ts:499` and `tree/tree-selection.component.ts:1312` go through `kbqGetElementHeight`.

The select one is a live bug of the same shape as the tree: `getHeightOfOptionsContainer(): number` returns `undefined` whenever there is no rect, and `select.component.ts:1970` then computes `Math.floor(undefined / this.options.first.getHeight())` → `NaN` → `withScrollSize(NaN)`. The declared return type is a lie.

The `?.` in the two list sites is after the call, so it guards a return value a real DOM never produces — not the missing method their JSDoc ("or 0 when it is not laid out (SSR, jsdom, `display: none`)") is about. They are safe today only because `list-selection.component.ts:597` gates the caller; the moment one is reached from a constructor effect — exactly how the tree got there — it is a `TypeError`, not a `0`.

Routing all four through `kbqGetElementHeight` here makes this helper the single place a future guard has to land, which is the point of having it.

packages/components/filter-bar/filter-bar-pipes.ts
```diff
@@ -0,0 +1,38 @@
+import { InjectionToken, Provider, Type } from '@angular/core';
```
lskramarov, line 1: **The pipes cycle is genuinely broken, but two cycles survive in this folder and one of them has an eager edge.**

I verified the good part: `filter-bar.types.ts` is now a leaf (only `@angular/core`, `@koobiq/components/core`, `rxjs` and one `import type`), every `extends KbqBasePipe` resolves through the acyclic `pipes/` subtree, and this new file introduces no new cycle. What remains:

1. `filter-bar.ts:28 ↔ filters.ts:29` — both back-edges deferred (`contentChild(KbqFilters)`, `inject(KbqFilterBar)`), so benign.
2. `filters.ts:30 ↔ filter-bar-button.ts:5` — **not** deferred on one side. `filters.ts:44` lists `KbqFilterBarButton` inside the `@Component({ imports: [...] })` array, which is evaluated at module scope, while `filter-bar-button.ts:19` does `inject(KbqFilters)`. If a bundler ever enters that pair from `filter-bar-button.ts` first — two importers split across chunks, which is how this repo has hit it before — `filters.ts` evaluates while `KbqFilterBarButton` is still in its class-declaration TDZ. Today it is safe only because `public-api.ts:1` is `export * from './filter-bar'`, which happens to force the entry order.

Note `filter-bar-button.ts` already injects `KBQ_FILTER_BAR_HOST` for the bar but the concrete `KbqFilters` for the other side. AGENTS.md: *"When a projected child depends on its host component, inject a narrow `InjectionToken` scoped to the members it uses, not the host's concrete class."* A `KBQ_FILTERS` token would delete both remaining cycles rather than relocate an edge.

Separately: `eslint.config.js` has no `import/no-cycle` (and there is no madge / dependency-cruiser step), so nothing prevents the next one. Given this PR exists because a cycle escaped review, a gate here would be worth more than the file move.

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
```diff
@@ -1,81 +1,57 @@
...
 })
 export class CheckboxIndeterminateExample {
-    private ref = inject(ChangeDetectorRef);
-
-    parentIndeterminate = true;
-    parentChecked = true;
+    protected readonly eventTypes = ['Malware', 'Phishing', 'Ransomware'];
+    protected readonly selected = model(new Set([this.eventTypes[0]]));
```
lskramarov: **`model()` for state that is never two-way bound.**

AGENTS.md, *Components*: "Use `model()` for two-way bound properties with `[(prop)]` syntax instead of pairing `input()` with `output()`"; *State Management*: "Use signals for local component state".

`selected` is only read (`selected().has(eventType)` at :18, `this.selected().size` at :39-40) and written imperatively (:43, :47) — there is no `[(selected)]` anywhere. `model()` still registers a real `selected` input and `selectedChange` output on the component definition, so the example ships public API it does not want, and every `.set()` / `.update()` fires a `selectedChange` emission with no subscriber.

Repo usage is unanimous the other way — `clamped-list-example.ts:51` uses `model(true)` *because* it is bound at `:18`, while `clamped-list-overview-example.ts:55` uses `signal(true)` for the unbound case.

`signal(...)` is a drop-in with the same initializer (the import on line 1 changes too). Since this is a docs example, it is also what readers copy off the site.

packages/components-dev/ssr/routes.ts
```diff
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
 import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
 
 const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
```
lskramarov, line 9: **The list is being treated as a gate, but `ssr:build` cannot observe two of the three bug classes it now covers.**

`.github/workflows/build.yml` runs `yarn run ssr:build` and nothing loads the output in a browser; no Playwright project or spec targets this app. NG0500 is thrown by client-side `validateMatchingNode`, *after* the browser's parser has restructured the server markup — prerendering the same markup succeeds. So re-adding `checkbox-indeterminate` certifies only that it renders on the server, which it always did. The verification described in the PR body ("loaded all 568 prerendered routes through Playwright") is exactly the right check, but it is not committed, so the next example that nests a `kbq-*` component inside a `<p>` regresses on koobiq.io with a green `ssr` job.

Same gap for the new `dom.ts` guard: `grep` over `packages/**/*.spec.ts` finds no reference to `kbqGetElementHeight`, and `packages/components/core/utils/` has no spec file at all. It is also invisible to the compiler (`Element.getClientRects` is non-optional in `lib.dom.d.ts`) and to lint (`eslint.config.js` extends only the non-type-checked `tseslint.configs.recommended`, so `no-unnecessary-condition` cannot fire) — nothing stops someone "simplifying" the `?.` away again.

A small spec that iterates the exported `devSsrExampleIds` against `ssr:start` and fails on any console error would turn this list from a to-do into a gate, and would be the thing that justifies removing entries from it.

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
```diff
@@ -1,81 +1,57 @@
...
 /**
  * @title Checkbox indeterminate
  */
 @Component({
     selector: 'checkbox-indeterminate-example',
-    imports: [
-        KbqCheckboxModule
-    ],
+    imports: [KbqCheckboxModule],
     template: `
```
lskramarov, line 10: **The NG0500 is a `kbq-checkbox` trap, not an example bug — rewriting the example removes the only repro.**

`checkbox.html:1-3`:

```html
<label #label class="kbq-checkbox__layout" [attr.for]="inputId">
    <div
        class="kbq-checkbox__inner-container"
```

The HTML parser auto-closes any open `<p>` at that `<div>`, so the browser's tree differs from the server's. Any consumer following `checkbox.en.md:11-15` ("a group of options and a higher-level checkbox") with a `<p>` — or any other phrasing-only wrapper — under SSR gets NG0500 with a message that points at `<p>` and never at the checkbox.

All three offending elements set their own `display` in `checkbox.scss` — `.kbq-checkbox__inner-container` (`inline-block`), `.kbq-checkbox__frame` (`flex`), `.kbq-checkbox__text-container` (`flex`) — so changing the three `<div>`s to `<span>`s is visually a no-op and removes the constraint for everyone.

If that is out of scope here, a sentence in `checkbox.{en,ru}.md` would at least record it: after this PR nothing in the repo exercises the shape, so a future change to `checkbox.html` cannot regress it visibly either.

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
```diff
@@ -1,81 +1,57 @@
...
-                    </kbq-checkbox>
-                </p>
-            }
-        </div>
+        }
     `,
-    changeDetection: ChangeDetectionStrategy.OnPush
+    styles: `
+        .example-checkbox-indeterminate__child {
+            margin-left: var(--kbq-size-l);
```
lskramarov: **The indent is one frame width, so the children land 8px short of the parent's label — and the rule duplicates a shipped utility.**

`checkbox-tokens.scss` sets `--kbq-checkbox-size-normal-width: var(--kbq-size-l)`, i.e. this margin equals exactly the frame width. But `.kbq-checkbox__layout` offsets its text by `--padding: calc(var(--kbq-checkbox-size-normal-width) + var(--kbq-checkbox-size-normal-horizontal-content-padding))` = 16px + 8px = 24px (`checkbox.scss:24,41`). So the parent's label starts at x=24 while a child's frame starts at x=16 — the child box straddles the parent's text origin instead of aligning under it, which reads as a misalignment rather than a nesting level:

```suggestion
            margin-left: calc(var(--kbq-size-l) + var(--kbq-size-s));
```

Separately, if 16px *were* the intended value, the whole `styles` block is redundant: the `layout-indents()` mixin already emits `.layout-margin-left-l { margin-left: var(--kbq-size-l) !important; }` (`_layout.scss:461-500`, from the `'l'` entry at `:468`), and two examples in this very folder use those utilities instead of a local rule — `pseudo-checkbox-example.ts:15` and `checkbox-overview-example.html:2`.

packages/components/filter-bar/filter-bar-pipes.ts
```diff
@@ -0,0 +1,38 @@
...
+    [KbqPipeTypes.Select, KbqPipeSelectComponent],
+    [KbqPipeTypes.TreeSelect, KbqPipeTreeSelectComponent],
+    [KbqPipeTypes.MultiSelect, KbqPipeMultiSelectComponent],
+    [KbqPipeTypes.MultiTreeSelect, KbqPipeMultiTreeSelectComponent],
+    [KbqPipeTypes.Date, KbqPipeDateComponent],
+    [KbqPipeTypes.Datetime, KbqPipeDatetimeComponent]
+];
+
+/** Injection Token for providing pipes in filter-bar */
+export const KBQ_FILTER_BAR_PIPES = new InjectionToken<Map<KbqPipeType, Type<KbqBasePipe<unknown>>>>(
```
lskramarov, line 32: **Worth giving the token a `factory` default while it is moving.**

As declared, `KBQ_FILTER_BAR_PIPES` is resolvable only through `KbqFilterBarModule` — `pipe.directive.ts:11` injects it non-optionally (`private pipes = inject(KBQ_FILTER_BAR_PIPES);`) and the only provider in the repo is `filter-bar.module.ts:27`. `KbqPipeDirective` and `KbqFilterBar` are standalone and exported from `public-api.ts`, so a consumer who imports the standalone pieces without the NgModule — the direction AGENTS.md pushes toward ("Always use standalone components over NgModules") — hits `NG0201: No provider for InjectionToken kbq-filter-bar-pipes` at first render, with nothing pointing at the NgModule as the only source.

The repo's dominant shape is the token owning its own default, with the `kbq*Provider` factory reserved for overrides — `actions-panel-config.ts:6-9`, `breadcrumbs.ts:55-64` (the pair side by side), `core/locales/a11y.ts:9-12`.

```ts
export const KBQ_FILTER_BAR_PIPES = new InjectionToken<Map<KbqPipeType, Type<KbqBasePipe<unknown>>>>(
    'kbq-filter-bar-pipes',
    { providedIn: 'root', factory: () => new Map(defaultFilterBarPipes) }
);
```

That makes the module's `providers` entry deletable; `kbqFilterBarPipesProvider` can stay exported for source compatibility. Pre-existing rather than introduced here — flagging it because this is the moment the declaration is being touched.

packages/components-dev/ssr/routes.ts
```diff
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
...
-    // TODO: Restore after fixing SSR errors in accordion examples. (#DS-5467)
-    'components/accordion',
     // AG Grid does not support server-side rendering.
-    'components/ag-grid',
-    // TODO: Restore after breaking the circular dependency in filter-bar. (#DS-5467)
-    'components/filter-bar',
-    // TODO: Restore with filter-bar; this barrel exports UsernameFilterBarOptionExample. (#DS-5467)
-    'components/username'
+    'components/ag-grid'
 ]);
```
lskramarov, line 21: **Shrinking the exclusion set silently moves `devSsrExampleIds[0]`.**

`EXAMPLE_COMPONENTS` starts at `"accordion-content"` with `"importPath": "components/accordion"` (`example-module.ts:29-43`). With `components/accordion` excluded, index 0 was the first non-accordion example; removing the exclusion makes it `accordion-content`.

That value is both the landing route and every catch-all destination:

```ts
{ path: '', redirectTo: devSsrExampleIds[0], pathMatch: 'full' },
{ path: '**', redirectTo: devSsrExampleIds[0] }
```

So `/` and the popover example's iframe both now render an accordion, as a side effect of an exclusion-list edit. Naming the intended landing example explicitly would make that a decision rather than a consequence of catalogue ordering.

packages/components-dev/ssr/routes.ts
```diff
@@ -3,25 +3,12 @@ import { Routes } from '@angular/router';
 import { EXAMPLE_COMPONENTS, loadExample } from '../../docs-examples/example-module';
 
 const SSR_EXCLUDED_EXAMPLE_IDS = new Set([
-    // TODO: NG0500: During hydration Angular expected a comment node but found <p>. (#DS-5467)
-    'checkbox-indeterminate',
-    // TODO: KbqTreeSelection.getHeight() calls getClientRects(), absent under SSR. (#DS-5467)
-    'tree-select-select-all',
-    'tree-select-select-all-label',
     // AG Grid does not support server-side rendering.
```
lskramarov, line 10: **The id exclusion does not keep AG Grid out of the prerender worker.**

`loadExample` imports the whole barrel, and `content-panel-overview` is not excluded: `example-module.ts:1369-1380` gives it `"importPath": "components/content-panel"`, and `packages/docs-examples/components/content-panel/index.ts:3` imports `ContentPanelWithGridExample`, whose module body (`content-panel-with-grid-example.ts:11-24`) imports `ag-grid-angular` / `ag-grid-community` and runs `ModuleRegistry.registerModules([AllCommunityModule]);` at evaluation time.

So "AG Grid does not support server-side rendering" is honoured only for the `components/ag-grid` barrel, which is never loaded anyway — AG Grid is still imported and initialized in the prerender worker via the overview example. It survives today, but the day an ag-grid release touches a browser global at import time, the failing route is `content-panel-overview` and this entry will not help.

The docblock just below presents the two sets as interchangeable mechanisms; they are not — an id exclusion skips a route, an import-path exclusion is what keeps a barrel out. Not introduced by this PR, but it is the surviving half of the comment.

lskramarov (approved):

NikGurev (approved):

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
artembelik: Вертикальный отступ добавил в то же локальное правило, что и margin-left, - утилита потребовала бы второй класс там, где правило уже есть.

packages/components/core/utils/dom.ts
artembelik: Все четыре места переведены на kbqGetElementHeight; в KbqSelect.updateScrollSize добавил ещё guard на нулевую высоту опции - без него честный 0 из хелпера всё равно давал бы NaN.

packages/components-dev/ssr/routes.ts
artembelik: Вернул в исключения оба примера с iframe, popover и select; маршрут examples/<name> в этом приложении заводить не стали.

packages/components-dev/ssr/routes.ts
artembelik: disableAnimation() теперь не вызывается на сервере в KbqAccordionContentDirective и KbqAccordionTrigger - transition: none в пререндеренный HTML не попадает.

packages/components/filter-bar/filter-bar-pipes.ts
artembelik: Ввёл KBQ_FILTERS, цикл filters ↔ filter-bar-button ушёл; правило import/no-cycle - вне этого PR.

packages/components-dev/ssr/routes.ts
artembelik: Спека в этот PR не входит - вынес в DS-5539, в routes.ts оставил TODO со ссылкой.

packages/components/filter-bar/filter-bar-pipes.ts
artembelik: Фабрику добавил; kbqFilterBarPipesProvider() в модуле оставил - он ничего не ломает и остаётся точкой переопределения.

packages/docs-examples/components/checkbox/checkbox-indeterminate/checkbox-indeterminate-example.ts
artembelik: Правку checkbox.html выношу в отдельную задачу - в этом PR меняем только пример.

packages/components-dev/ssr/routes.ts
artembelik: Оставляем как есть: это dev-приложение, селектор примеров в шапке работает, а выбор посадочного примера так же произволен, как алфавитный порядок.

packages/components-dev/ssr/routes.ts
artembelik: Сегодня влияния нет - инициализация ag-grid браузерных глобалов не трогает; сужать покрытие ради гипотетического риска не стали.

PR #2012, opened 2026-09-08
test(e2e): run playwright at zero retries and keep a trace per failure

artembelik:
## Summary

Playwright retried a failing test twice on CI, so a test that failed and then passed was reported as flaky and did not fail the run — the suite could be reliably green while being measurably unstable. The default is now **0 retries everywhere**, CI included: an unstable test fails the run and gets named. `PLAYWRIGHT_RETRIES` stays as the escape hatch for a run that has to be nursed through a known flake.

That change alone would have made failures harder to diagnose, since `trace: 'on-first-retry'` never fires without a retry. Traces are therefore retained on failure — minus the screencast, which is the expensive half twice over: it takes CPU from every test (and this suite's known flakes are the ones that only appear under load, per [`docs/e2e-flakiness.md`](docs/e2e-flakiness.md)), and it dominates the size of the uploaded report on the run where a Chromium bump invalidates every baseline at once. What is kept — DOM snapshots, network, the action log — is what actually explains a screenshot flake; the pixels are already attached as `-actual.png`, `-expected.png` and `-diff.png`.

The staged rollout this repo's own audit recommended (hold CI at 2, soak at 0 nightly, then flip) was skipped in favour of flipping directly, on the strength of its ×5 verification run: 2810 results, 0 failures. That deviation is now recorded in `docs/e2e-flakiness.md` rather than left implicit.

## List of notable changes:

- **updated** `playwright.config.ts` — `retries` defaults to 0 regardless of `CI`; `trace` is `{ mode: 'retain-on-failure', screenshots: false }`
- **updated** `.github/workflows/e2e-approve-snapshots.yml` — asks for `PLAYWRIGHT_RETRIES: 2` explicitly, because that job regenerates baselines rather than measuring stability: a flake there reports nothing anyone acts on and costs a maintainer another `/approve-snapshots` plus another full image build
- **updated** `.github/workflows/e2e.yml`, `tools/e2e/Dockerfile`, `tools/e2e/docker-compose.yml` — comments that claimed `CI=true` means two retries, and that `PLAYWRIGHT_RETRIES=0` is the way to see flakiness, are now the opposite of the truth
- **updated** documentation — `docs/guides/06-testing.md` (Retries section rewritten, plus the stale "retried twice" claim in the Docker section and where to find a trace), `docs/e2e-flakiness.md`, `AGENTS.md`

1. **Whether the docs smoke should follow.** `playwright.docs.config.ts` spreads `...baseConfig`, so `e2e:docs` inherits 0 retries too — but it was never part of the flakiness audit, which measured only `packages/components`. Hydration and SSR smoke specs are the classic shape of a timing flake, so this is the job most likely to start turning PRs red. Dialling it back is one line (`PLAYWRIGHT_RETRIES: 2` in that job's `env`) if we would rather not find out the hard way.
2. **Whether `screenshots: false` gives up too much.** The trace viewer loses its filmstrip; the DOM snapshots and the separate diff PNGs are the argument that nothing important goes with it.
3. **The retry cushion put back on `/approve-snapshots`** — deliberate asymmetry with `e2e.yml`, worth agreeing on.

Not verified locally: the suite was not run for this change (no local Docker run in this session), so the real cost of per-test trace recording on a 4-vCPU runner will first show up in this PR's own E2E job.

### DIFF
--- a/.github/workflows/e2e-approve-snapshots.yml
+++ b/.github/workflows/e2e-approve-snapshots.yml
@@ -48,6 +48,10 @@ jobs:
           # As in e2e.yml: the compose default is tuned for developer machines, the runner wants
           # its own core count.
           PLAYWRIGHT_WORKERS: 100%
+          # Unlike e2e.yml, this job regenerates baselines rather than measuring stability. A flake
+          # here reports nothing anyone acts on and costs a maintainer another /approve-snapshots
+          # plus another full image build, so buy the cushion back that the 0 default removes.
+          PLAYWRIGHT_RETRIES: 2
       - uses: stefanzweifel/git-auto-commit-action@4a55954c782fc1ea30b9056cd3e7a2b40ca8887d # v7.2.0
         id: commit-and-push
         with:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -5,13 +5,13 @@ on:
     branches:
       - main
   pull_request:
-  # Retries keep the runner green through a flake, which is the right default for a pull request and
-  # the wrong one for finding out what is unstable. This is how the suite gets run at zero retries
-  # without blocking anyone; see docs/guides/06-testing.md.
+  # Retries are 0 everywhere, so a flake fails the run and gets named rather than kept green. This
+  # input is the escape hatch: a dispatched run can put them back to nurse the suite through a known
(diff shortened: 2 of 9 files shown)


artembelik:
влить после https://github.com/koobiq/angular-components/pull/2014

lskramarov (approved):

PR #2014, opened 2026-09-08
fix(tabs): align overflow on scale (#DS-5520)

NikGurev:
## Summary

At fractional browser zoom the tab header kept its next arrow and `_overflow-after` mask lit at the end of
the scroll range. `scrollWidth`/`clientWidth` round to whole CSS pixels while `scrollLeft` snaps to the
device-pixel grid, so the two disagree by up to `1 + 1 / devicePixelRatio` px at the real end of the range.

## Changes

- New `kbqGetScrollOverflowTolerance` / `kbqHasScrollOverflow` in `@koobiq/components/core`: the tolerance
  is derived from `devicePixelRatio` rather than being a constant (3px at 50% zoom, 4px at 33%).
- `KbqPaginatedTabHeader` spends it only where a rounding error exists — the pagination mount threshold and
  the far scroll bound. The "before" bound stays an exact `<= 0`.
- `startInertia` finishes a coast on contact with the far bound instead of waiting for a clamp that never
  fires there.
- The same comparison in `tab-label-wrapper`, `overflow-items`, `navbar-item` and `code-block` is left for a
  separate task; the helper's `axis` parameter is shaped for it.

## E2E baselines

`setupTabsUnderlinedTextIconStates` indexed `labels[4]` in a 4-tab group. It threw, and the shared
`afterNextRender` callback let that skip `setupTabsVerticalIconsOnlyStates` as well — its two focus rings
were missing from the old baselines and are back now. The state that line was meant to add moved to the
group's last tab, and the three setups now run separately through a helper that throws on an out-of-range
index.

## Review focus

- The tolerance goes on the mount threshold and the far bound only.
- `startInertia`'s direction-aware boundary check — a flick back inwards from the far end must survive it.
- The mount threshold widens, by at most the tolerance, the range where an underlined header falls into the
  `overflow: visible` branch.


packages/e2e/utils/scroll-position.ts
```diff
@@ -0,0 +1,46 @@
...
+ * `toHaveScreenshot`'s own re-shoot loop does not rescue it either. That stops as soon as two
+ * consecutive frames match, which both the easing tail and the debounce window before the scroll
+ * has even started satisfy while the component is still on its way to the settled position.
+ *
+ * Quiet time rather than a stable-value poll, because those two windows are exactly the ones in
+ * which nothing scrolls yet nothing has settled. `quietMs` has to outlast the longer of them, plus
+ * the change-detection throttle that repaints the pagination arrows and the edge masks after the
+ * last scroll event.
+ */
+export const e2eWaitForSettledScrollPositions = (page: Page, selector: string, quietMs = 250): Promise<void> =>
```
artembelik: используется только в 1 месте, предлагаю сделать утилиту локальной, скорей всего больше нигде не потребуется

artembelik (approved):

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -69,6 +69,13 @@ const FRICTION_PER_MILLISECOND = 0.003;
 /** Audit interval (ms) for the scroll-box `ResizeObserver` — see the `auditTime` usage below. */
 const RESIZE_AUDIT_TIME = 100;
 
+/**
+ * Tolerance (px) for the scroll-bound and overflow checks. `scrollWidth`/`clientWidth` are rounded
+ * to whole CSS pixels while `scrollLeft` snaps to the device-pixel grid, so at fractional zoom the
+ * two disagree by up to a full pixel at the real end of the range — more than a `Math.ceil` absorbs.
+ */
+const SCROLL_BOUND_TOLERANCE = 2;
```
lskramarov: **A fixed 2px tolerance is too small below ~67% browser zoom.**

`scrollWidth`/`clientWidth` round to whole CSS px (up to 1px of error), and `scrollLeft` snaps to the device-pixel grid — where one device pixel is `1 / devicePixelRatio` CSS px. The worst-case disagreement is therefore about `1 + 1 / devicePixelRatio`, not a constant.

At the 0.8 zoom the spec comment measures that is ~1.6–2.25px, so `2` only just covers it. At Chrome's lower zoom steps on a 1× display it does not: 50% → dPR 0.5 → bound 3px; 33% → 4px; 25% → 5px. Below ~67% zoom the next arrow and the `_overflow-after` mask stay lit at the end of the range again — the exact bug this PR fixes.

`KBQ_WINDOW` is already injected as `this.window`, so the bound can be derived rather than guessed:

```ts
private get scrollBoundTolerance(): number {
    return 1 + 1 / (this.window.devicePixelRatio || 1);
}
```

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -697,11 +706,13 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
 
         const container = this.tabListContainer.nativeElement;
         const position = this.logicalScrollPosition;
+        const maxScrollPosition = container.scrollWidth - container.clientWidth;
```
lskramarov, line 714: **`startInertia` computes the same maximum without the tolerance, so a coast never detects the end bound at fractional zoom.**

Line 867 repeats `container.scrollWidth - container.clientWidth` as `maxScrollLeft` and uses it as the clamp for the inertia loop:

```ts
const clamped = Math.max(minBound, Math.min(maxBound, position));

container.scrollLeft = clamped;

const reachedBoundary = clamped !== position;
```

At the very zoom levels this PR is about, that bound overstates the largest offset the browser will hand out by ~1px, so `clamped !== position` never becomes true at the end of the range. A flick towards the end stops decaying instead of stopping on contact: the loop keeps requesting frames and writing an offset the browser silently clamps until friction alone kills the velocity — and `scrollCorrection` stays suppressed for that whole tail (`if (this.dragState || this.inertiaFrameId !== null) return;`).

Whatever tolerance lands here should reach that call site too.

packages/components/tabs/e2e.playwright-spec.ts
```diff
@@ -1,6 +1,53 @@
...
+ * Quiet time rather than a stable-value poll, because those two windows are exactly the ones in
+ * which nothing scrolls yet nothing has settled. `quietMs` has to outlast the longer of them, plus
+ * the change-detection throttle that repaints the pagination arrows and the edge masks after the
+ * last scroll event.
+ */
+const waitForSettledTabScroll = (page: Page, quietMs = 250): Promise<void> =>
+    page.evaluate(
+        ({ scrollportSelector, quietTime }) =>
+            new Promise<void>((resolve) => {
+                const scrollports = Array.from(document.querySelectorAll(scrollportSelector));
```
lskramarov, line 28: **The helper resolves after one quiet window when the selector matches nothing, so the de-flaking is silently reversible.**

An empty `scrollports` list is indistinguishable from "everything already settled": `restart()` is armed unconditionally on line 46, so the promise resolves after one `quietMs` window and the test goes green.

Rename `kbq-tab-header__scroll-container`, or point the spec at a route without headers, and this degrades to a bare 250ms sleep with no failure anywhere — the flake comes back looking fixed. `e2eWaitForSettledScrollbars` in `packages/e2e/utils/scrollbar.ts` documents the opposite rule for exactly this reason: *"Pass `expectedTracks` when the number is known — a count that never arrives fails loudly instead of passing on an empty match."*

Either assert the count in the spec before waiting:

```ts
await expect(page.locator(TAB_SCROLLPORT_SELECTOR)).not.toHaveCount(0);
```

or reject inside the `evaluate` when `scrollports.length === 0`.

packages/components/tabs/e2e.ts
```diff
@@ -240,8 +240,6 @@ export class E2eTabsStates {
 
         labels[2].classList.add('kbq-hover');
         labels[2].classList.add('cdk-keyboard-focused');
```
lskramarov: **The deleted line was throwing, and removing it silently changes what two other groups render.**

`e2eTabsUnderlinedTextIcon` renders `tabs.slice(0, 4)`, so `labels` has 4 entries and `labels[4]` was `undefined` — the removed line raised a `TypeError`. Since all three setups run in one `afterNextRender` callback, that aborted the callback and `setupTabsVerticalIconsOnlyStates()` never ran at all.

Removing the line un-deads it, and that is the *entire* content of the new baselines. Diffing `01-light.png` / `01-dark.png` against `main` gives 2159 / 2161 changed pixels, confined to two 72×72 boxes at x 1660–1731, y 460–531 and y 636–707 — the two `cdk-keyboard-focused` rings on the vertical icons-only group. Nothing in the scrolling groups moved. Worth naming in the PR description; "updated flacky test" does not cover a baseline gaining states for an unrelated group.

Two things are left unaddressed:

- the text+icon group loses the focus-without-hover state the line existed to cover — retargeting to a valid index (1 or 3) would keep it;
- the pattern that hid this stays: unguarded index access across three sequential setups where the first failure silently kills the rest. Asserting `labels.length`, or isolating each setup, would make the next index drift fail loudly instead of quietly rewriting a baseline.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -697,11 +706,13 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
...
 
-        // `Math.ceil` guards against subpixel `scrollWidth`/`clientWidth` rounding producing a
-        // false "still scrollable" reading right at the end.
-        this.disableScrollBefore = position <= 0;
-        this.disableScrollAfter = Math.ceil(position + container.clientWidth) >= container.scrollWidth;
+        // Compared with a tolerance: at the bounds `position` and `maxScrollPosition` differ by up to
+        // a pixel of rounding error, which an exact comparison reports as remaining scroll distance —
+        // leaving the arrow and its edge mask enabled with nowhere left to scroll.
+        this.disableScrollBefore = position <= SCROLL_BOUND_TOLERANCE;
```
lskramarov: **The "before" bound has no rounding to absorb — the tolerance here is a small regression.**

The start of the scroll range is exactly `0` in both directions (`logicalScrollPosition` negates `scrollLeft` for RTL), and neither `scrollWidth` nor `clientWidth` takes part in this comparison. The mechanism the constant's doc describes — *"the two disagree by up to a full pixel at the real end of the range"* — only applies to `disableScrollAfter` on line 715.

What the tolerance does here instead: at a genuine offset of 1–2px the previous arrow gets `.kbq-disabled` (`opacity: 0; visibility: hidden; pointer-events: none` in `_tabs-common.scss`) and `_overflow-before` drops the left mask, while the first tab really is clipped. `position <= 0` was correct, and keeping it also closes the both-bounds window described on line 556.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -544,7 +551,9 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
         }
 
         const container = this.tabListContainer.nativeElement;
-        const isEnabled = container.scrollWidth > container.clientWidth;
+        // Tolerated rather than an exact `>`, so a rounding-sized overflow can't mount arrows that
+        // `updateScrollState` immediately reports as sitting on both bounds at once.
+        const isEnabled = container.scrollWidth - container.clientWidth > SCROLL_BOUND_TOLERANCE;
```
lskramarov: **The comment's guarantee does not hold with a single tolerance.**

Mounting requires overflow `> SCROLL_BOUND_TOLERANCE`, but `updateScrollState` spends the tolerance at *each* end, so "sitting on both bounds at once" is only excluded when overflow `> 2 * SCROLL_BOUND_TOLERANCE`.

With overflow of 3px and the strip at 1.5px, both `position <= 2` and `position >= 3 - 2` hold: the arrows are mounted (`kbq-tab-header__pagination-controls_enabled`, `cursor: grab`) and both are `visibility: hidden`, with 1.5px still scrollable in each direction. The window is `overflow ∈ (T, 2T)` and widens linearly with the constant — so it gets worse, not better, if the tolerance is raised to cover lower zoom levels.

Either compare against `2 * SCROLL_BOUND_TOLERANCE` here, or drop the tolerance from the "before" bound (line 714), which closes the window as well.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -544,7 +551,9 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
         }
 
         const container = this.tabListContainer.nativeElement;
-        const isEnabled = container.scrollWidth > container.clientWidth;
+        // Tolerated rather than an exact `>`, so a rounding-sized overflow can't mount arrows that
```
lskramarov: **Raising the mount threshold also flips a CSS branch on the underlined variants.**

`tab-header.scss` and `_tabs-common.scss` key `overflow` off `showPaginationControls`:

```scss
.kbq-tab-header_underlined:not(.kbq-tab-header__pagination-controls_enabled) { overflow: visible; }
.kbq-tab-header__container { .kbq-tab-header_underlined:not(.kbq-tab-header__pagination-controls_enabled) & { overflow: visible; } }
```

A header with a 1–2px overflow used to be "enabled" and clipped by `overflow-x: auto`; it now falls into the `overflow: visible` branch, so those pixels of the last tab paint outside the header's `overflow: hidden` and outside its `border-radius`. Small, but it is a new escape path at exactly the fractional zoom levels this constant targets — worth a look at an underlined group at 0.8 zoom before merging.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -544,7 +551,9 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
...
+        // `updateScrollState` immediately reports as sitting on both bounds at once.
+        const isEnabled = container.scrollWidth - container.clientWidth > SCROLL_BOUND_TOLERANCE;
 
         if (!isEnabled) {
```
lskramarov, line 563: **This branch now also fires for a genuine 1–2px overflow, forcing `scrollLeft = 0`.**

For non-underlined headers the container keeps `overflow-x: auto`, so a user can still wheel those pixels into view — and the next `updatePagination()` (the `ResizeObserver` audit, a content mutation, an items change) snaps them straight back to 0, with no arrows to get there. The same header was paginated and kept its position before this change.

Nothing dramatic at 2px, but it is the reason to keep the tolerance as tight as the measurement justifies rather than rounding it up.

packages/components/tabs/tab-header.spec.ts
```diff
@@ -255,6 +255,72 @@ describe('KbqTabHeader', () => {
...
+
+                expect(header.disableScrollAfter).toBe(true);
+                expect(
+                    fixture.nativeElement
+                        .querySelector('.kbq-tab-header__pagination_after')
+                        .classList.contains('kbq-disabled')
+                ).toBe(true);
+            });
+
+            it('should keep the previous arrow disabled for a rounding-sized scroll offset', () => {
```
lskramarov: **One-sided assertion: this passes for any tolerance value.**

`updatePagination()` at offset 0 already sets `disableScrollBefore = true`, so the test only shows the flag did not flip — never that the tolerance has an upper edge. It stays green if `SCROLL_BOUND_TOLERANCE` were bumped to 300 (which would disable the previous arrow across the whole range), and it stays green if the field were hardcoded `true`; no other LTR test asserts `disableScrollBefore === false`.

The next-arrow test above pins its bound from both sides — this one should too:

```ts
header.updatePagination();
fixture.detectChanges();

container.scrollLeft = 10;
container.dispatchEvent(new Event('scroll'));
fixture.detectChanges();

expect(header.disableScrollBefore).toBe(false);

container.scrollLeft = 1;
container.dispatchEvent(new Event('scroll'));
fixture.detectChanges();

expect(header.disableScrollBefore).toBe(true);
```

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -69,6 +69,13 @@ const FRICTION_PER_MILLISECOND = 0.003;
 /** Audit interval (ms) for the scroll-box `ResizeObserver` — see the `auditTime` usage below. */
 const RESIZE_AUDIT_TIME = 100;
 
+/**
+ * Tolerance (px) for the scroll-bound and overflow checks. `scrollWidth`/`clientWidth` are rounded
```
lskramarov: **The tolerance is scoped to this file, but the comparison it corrects is repeated verbatim across the library.**

- `packages/components/tabs/tab-label-wrapper.directive.ts:79` — `isOverflown()` drives the tab's ellipsis tooltip. Same component: after this PR a tab header's arrows are correct at 0.8 zoom while a tab that is not actually truncated can still get a tooltip.
- `packages/components/overflow-items/overflow-items.ts:242`
- `packages/components/navbar/navbar-item.component.ts:100`
- `packages/components/code-block/code-block.ts:527`

A `kbqHasScrollOverflow(element, axis)` alongside the other DOM helpers in `packages/components/core/utils/dom.ts` would fix all of them at once and give the constant a single home, instead of leaving one component correct and the rest carrying the original bug.

packages/components/tabs/tab-header.spec.ts
```diff
@@ -421,6 +487,27 @@ describe('KbqTabHeader', () => {
                 expect(header.disableScrollBefore).toBe(false);
                 expect(header.disableScrollAfter).toBe(true);
             });
+
+            it('should disable the next arrow at the scroll end when the box metrics and the scroll offset disagree', () => {
```
lskramarov: **Duplicate test title, and a near-duplicate body.**

The `it` title is verbatim identical to the LTR test on line 280. `npx jest -t 'should disable the next arrow at the scroll end when the box metrics and the scroll offset disagree'` runs both, and a failure report names neither direction — a `… in RTL` suffix (or one direction-parameterised test) fixes that.

The body is also close to a duplicate of `should toggle the pagination arrows from the negative RTL scrollLeft range` (line 466); only the metrics differ. Folding the 1139/318/-820 case into that test would keep the RTL bound coverage in one place.

packages/components/tabs/tab-header.spec.ts
```diff
@@ -255,6 +255,72 @@ describe('KbqTabHeader', () => {
...
+                fixture.detectChanges();
+
+                expect(header.disableScrollAfter).toBe(false);
+
+                container.scrollLeft = 820;
+                container.dispatchEvent(new Event('scroll'));
+                fixture.detectChanges();
+
+                expect(header.disableScrollAfter).toBe(true);
+                expect(
```
lskramarov: **The DOM half of this assertion duplicates existing coverage.**

`should toggle kbq-disabled on the previous/next arrows at each scroll bound without removing them from the DOM` (line 619) already checks the same class on the same element, for both arrows, with the same 400/100 fixture. What is new here is only that `disableScrollAfter` flips at 820 against a metric max of 821 — which line 299 already asserts.

Dropping lines 300-304 keeps this test on the one thing it exists to prove.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Replaced with a derived bound: `kbqGetScrollOverflowTolerance(element)` returns `1 + 1 / devicePixelRatio`, read per call so it follows zoom changes rather than being frozen at load. It now lives in `packages/components/core/utils/dom.ts` instead of as a constant here.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Handled. `startInertia` keeps clamping to the real metric bounds — writing a reduced maximum would visibly pull the strip back at the start of a flick — and finishes the coast on contact instead:

```ts
const reachedBoundary =
    clamped !== position ||
    (isRtl
        ? velocity < 0 && clamped <= minBound + farBoundTolerance
        : velocity > 0 && clamped >= maxBound - farBoundTolerance);
```

The direction check is what keeps a flick back inwards from the far end alive; without it a coast starting inside the tolerance would end on its first frame. Both cases are covered by tests.

packages/components/tabs/e2e.playwright-spec.ts
NikGurev: Added `await expect(page.locator(TAB_SCROLLPORT_SELECTOR)).not.toHaveCount(0);` before the wait, so a renamed class fails on the count instead of degrading the helper into a 250ms sleep.

packages/components/tabs/e2e.ts
NikGurev: All three points addressed.

- The focused state is back, on the group's last (disabled) tab. A fifth tab to carry it pushes the group to 409px inside its 400px grid column — `overflow: visible` on the unpaginated underlined branch — and the ring's right edge gets clipped out of the shot. I checked that in the running app before settling on index 3, where the ring renders through `border-color` and is clearly visible.
- The three setups are now separate `afterNextRender` callbacks, so one failure cannot skip the others.
- They go through a helper that throws `[testid] no .kbq-tab-label at index N, the group has M`.

The baseline delta is named in the PR description.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Agreed — back to an exact `this.disableScrollBefore = position <= 0;`.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Closed by the previous one: the tolerance is now spent on the far bound only, so mounting at `overflow > T` does exclude sitting on both bounds at once.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Checked in the running app — the mechanism is real. On an unpaginated underlined header both the header and the scroll container compute to `overflow: visible` (with `border-radius: 8px` on the on-surface variant), so an overflow inside the tolerance paints the last tab's edge outside the rounded box. It is bounded by the tolerance (1.5px at dPR 2, 2px at dPR 1), and I am leaving it rather than widening this PR past the one component it fixes. Called out under review focus in the description.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Kept as is, on your own conclusion in this thread: the tolerance is now derived from the measurement rather than rounded up, so the range where this fires is exactly the range that is not measurable.

packages/components/tabs/tab-header.spec.ts
NikGurev: Replaced. The previous-arrow test now asserts the flip in both directions, and a separate test pins the next arrow's bound from both sides — offset 297 leaves `disableScrollAfter` false, 299 flips it true, against a metric maximum of 300.

packages/components/tabs/paginated-tab-header.ts
NikGurev: Extracted: `kbqGetScrollOverflowTolerance` and `kbqHasScrollOverflow(element, axis)` now sit next to the other DOM helpers in `packages/components/core/utils/dom.ts`, and `tabs` is their only consumer.

The four other call sites are deliberately left alone — this PR is a bug fix for one component, and a separate task will cover them. The `axis` parameter is shaped for that adoption (two horizontal, one that checks both axes).

packages/components/tabs/tab-header.spec.ts
NikGurev: Folded into `should toggle the pagination arrows from the negative RTL scrollLeft range` — the 1139/318/−820 metrics are now a second stage of that test, so the RTL bound coverage lives in one place and the duplicate title is gone.

packages/components/tabs/tab-header.spec.ts
NikGurev: Dropped — the test is back to just the `disableScrollAfter` flip.

lskramarov (approved):

PR #2018, opened 2026-09-09
chore(tooltip,ellipsis-center,title): load styles with directive and synced disabled (#DS-5271)

NikGurev:
## Summary

addressed review fixes from #2007 that are related to initial fix but not actually connected to it, just an improvement

### DIFF
--- a/packages/components/ellipsis-center/ellipsis-center.directive.ts
+++ b/packages/components/ellipsis-center/ellipsis-center.directive.ts
@@ -1,7 +1,9 @@
 import { SharedResizeObserver } from '@angular/cdk/observers/private';
+import { _CdkPrivateStyleLoader } from '@angular/cdk/private';
 import {
     AfterViewInit,
     ChangeDetectorRef,
+    Component,
     Directive,
     inject,
     Input,
@@ -9,7 +11,8 @@ import {
     NgModule,
     numberAttribute,
     OnDestroy,
-    OnInit
+    OnInit,
+    ViewEncapsulation
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { KbqTooltipTrigger } from '@koobiq/components/tooltip';
@@ -17,12 +20,20 @@ import { Subject } from 'rxjs';
 import { debounceTime } from 'rxjs/operators';
 
 /**
+ * Component used to load the `.kbq-ellipsis-center` styles.
+ */
+@Component({
+    selector: 'ellipsis-center-style-loader',
+    template: '',
+    styleUrl: 'ellipsis-center.scss',
+    encapsulation: ViewEncapsulation.None
+})
+class EllipsisCenterStyleLoader {}
+
+/**
  * Renders text as two spans and moves the ellipsis into the middle, so the tail — a file extension, the
  * digits that tell two reports apart — stays readable when the host is too narrow for the whole string. A
  * tooltip spells out the full text, and is enabled only while something is actually hidden.
- *
- * The host needs the layout contract the consuming component ships (see `.kbq-ellipsis-center` in
- * `file-upload.scss`): the two spans have no styles of their own.
  */
 @Directive({
     selector: '[kbqEllipsisCenter]',
@@ -40,6 +51,9 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
      * none of which resize the window, all of which change where the text has to be split.
      */
     private readonly resizeObserver = inject(SharedResizeObserver);
+
+    /** The two spans this directive renders have no styles of their own until this is loaded. */
+    private readonly styleLoader = inject(_CdkPrivateStyleLoader);
 
     // TODO: Skipped for migration because:
     //  Accessor inputs cannot be migrated as they are too complex.
@@ -94,6 +108,12 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
     private lastMeasuredWidth: number | undefined;
 
     private refreshTimeoutId: ReturnType<typeof setTimeout> | undefined;
+
+    constructor() {
+        super();
+
+        this.styleLoader.load(EllipsisCenterStyleLoader);
+    }
 
     override ngOnInit(): void {
         super.ngOnInit();
--- a/packages/components/ellipsis-center/ellipsis-center.directive.ts
+++ b/packages/components/ellipsis-center/ellipsis-center.directive.ts
@@ -35,8 +35,9 @@ class EllipsisCenterStyleLoader {}
  * digits that tell two reports apart — stays readable when the host is too narrow for the whole string. A
  * tooltip spells out the full text, and is enabled only while something is actually hidden.
  *
- * The tooltip's `forDisabledComponent` has no meaning here: the hint always repeats the text this directive
- * shortened, never an explanation of why a wrapped control is unavailable. Overflow is the only state the
+ * The tooltip's `forDisabledComponent` is not for this host and should not be bound on it: the hint always
+ * repeats the text this directive shortened, never an explanation of why a wrapped control is unavailable,
+ * and the two derive the same state, so they would overwrite each other. Overflow is the only state the
  * directive derives, and `kbqTooltipDisabled` is how a consumer suppresses the hint.
  */
 @Directive({
--- a/packages/components/ellipsis-center/ellipsis-center.directive.ts
+++ b/packages/components/ellipsis-center/ellipsis-center.directive.ts
@@ -34,6 +34,10 @@ class EllipsisCenterStyleLoader {}
  * Renders text as two spans and moves the ellipsis into the middle, so the tail — a file extension, the
  * digits that tell two reports apart — stays readable when the host is too narrow for the whole string. A
  * tooltip spells out the full text, and is enabled only while something is actually hidden.
+ *
+ * The tooltip's `forDisabledComponent` has no meaning here: the hint always repeats the text this directive
+ * shortened, never an explanation of why a wrapped control is unavailable. Overflow is the only state the
+ * directive derives, and `kbqTooltipDisabled` is how a consumer suppresses the hint.
  */
 @Directive({
     selector: '[kbqEllipsisCenter]',
@@ -141,6 +145,14 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
     }
 
     /**
+     * The hint spells out what the split hid, so `kbqTooltipDisabled="false"` cannot conjure one for a name
+     * that fits — unlike the base, where an explicit value wins outright.
+     * @docs-private */
+    protected override foldDisabled(): boolean {
+        return this.explicitlyDisabled === true || (this.derivedDisabled ?? false);
+    }
+
+    /**
      * Updates the displayed text with center ellipsis truncation based on container width.
      * Recreates start/end span elements, measures available space, and adjusts text accordingly.
      * @docs-private
@@ -215,7 +227,7 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
             dataTextStart.innerText = start;
             dataTextEnd.innerText = end;
 
-            this.disabled = !truncated;
+            this.setDerivedDisabled(!truncated);
             this.cdr.markForCheck();
         });
 
--- a/packages/components/ellipsis-center/ellipsis-center.scss
+++ b/packages/components/ellipsis-center/ellipsis-center.scss
@@ -1,0 +1,31 @@
+// Layout contract for the two spans `KbqEllipsisCenterDirective` renders. Loaded once by the directive
+// itself through `_CdkPrivateStyleLoader`, so a host no longer copies these declarations — every copy was a
+// chance for the directive to behave differently depending on where it was used.
+.kbq-ellipsis-center {
+    position: relative;
+    display: flex;
+
+    // The split point is measured against the host's own `clientWidth`, so the host must not grow to fit the
+    // text it is about to shorten — a shrink-to-fit host measures the full width, concludes everything fits,
+    // and leaves the text unsplit and overflowing. `min-width: 0` opts out of the `auto` minimum a flex item
+    // would otherwise inherit from the non-shrinkable end span.
+    max-width: 100%;
+    min-width: 0;
+    overflow: hidden;
+
+    .kbq-ellipsis-center_data-text-start {
+        // The only half that shrinks, and the only one carrying `text-overflow` — unsplit text lands here.
+        flex: 0 1 auto;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: pre;
+    }
+
+    .kbq-ellipsis-center_data-text-end {
+        // Holds the tail the split exists to keep visible (the file extension, the trailing digits), so it
+        // must not shrink; the directive sizes it from the width the text actually rendered at.
+        flex: 1 0 auto;
+        overflow: hidden;
+        white-space: pre;
+    }
+}
--- a/packages/components/title/title.directive.ts
+++ b/packages/components/title/title.directive.ts
@@ -204,7 +204,7 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
             .pipe(skip(1), takeUntilDestroyed())
             .subscribe(() => {
                 this.content = this.resolvedContent;
-                this.disabled = !this.isOverflown;
+                this.setDerivedDisabled(!this.isOverflown);
             });
     }
 
@@ -231,13 +231,13 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
         this.resizeObserver
             .observe(this.parent)
             .pipe(debounceTime(this.debounceInterval), takeUntilDestroyed(this.destroyRef))
-            .subscribe(() => (this.disabled = !this.isOverflown));
+            .subscribe(() => this.setDerivedDisabled(!this.isOverflown));
 
         this.contentObserver
             .observe(this.parent)
             .pipe(throttleTime(this.debounceInterval), takeUntilDestroyed(this.destroyRef))
             .subscribe(() => {
-                this.disabled = !this.isOverflown;
+                this.setDerivedDisabled(!this.isOverflown);
                 this.content = this.resolvedContent;
             });
 
@@ -264,7 +264,7 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
                     this.triggerName = 'blur';
                 }
 
-                // `disabled = true` hides an open tooltip through the base setter.
+                // Disabling hides an open tooltip on the way through `setDerivedDisabled`.
                 this.hideTooltip();
             });
     }
@@ -273,13 +273,22 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
      * @docs-private */
     protected handleElementEnter() {
         this.content = this.resolvedContent;
-        this.disabled = !this.isOverflown;
+        this.setDerivedDisabled(!this.isOverflown);
     }
 
     /** Always disables (hides) the tooltip. Bound to `mouseleave` and non-keyboard focus changes.
      * @docs-private */
     protected hideTooltip() {
-        this.disabled = true;
+        this.setDerivedDisabled(true);
+    }
+
+    /**
+     * The tooltip repeats text the host had to clip, so `kbqTooltipDisabled="false"` cannot conjure one for a
+     * title that is fully visible — unlike the base, where an explicit value wins outright. A consumer's
+     * `kbqTooltipDisabled="true"`, in turn, is no longer overwritten by the next hover.
+     * @docs-private */
+    protected override foldDisabled(): boolean {
+        return this.explicitlyDisabled === true || (this.derivedDisabled ?? false);
     }
 
     /**
--- a/packages/components/tooltip/tooltip.component.ts
+++ b/packages/components/tooltip/tooltip.component.ts
@@ -393,7 +393,12 @@ export class KbqTooltipTrigger
         this.updateData();
     }
 
-    /** Input (`kbqTooltipDisabled`) controlling whether the tooltip is disabled; setting it to `true` hides it. */
+    /**
+     * Input (`kbqTooltipDisabled`) controlling whether the tooltip is disabled; setting it to `true` hides it.
+     *
+     * Reads back the *effective* state, which for subclasses that derive one of their own (see
+     * `foldDisabled`) is not necessarily the value that was written.
+     */
     // TODO: Skipped for migration because:
     //  Accessor inputs cannot be migrated as they are too complex.
     @Input('kbqTooltipDisabled')
@@ -403,19 +408,23 @@ export class KbqTooltipTrigger
 
     set disabled(value) {
         this.explicitlyDisabled = coerceBooleanProperty(value);
-        this._disabled = this.explicitlyDisabled;
 
-        if (this._disabled) {
-            this.hide();
-        }
+        this.applyDisabled();
     }
 
     /**
      * Value the consumer assigned to `kbqTooltipDisabled`, or `undefined` while the input was never set. It
      * makes the input win over the state `forDisabledComponent` derives, instead of the two fighting over
      * `_disabled` in whichever order they happen to run.
-     */
-    private explicitlyDisabled: boolean | undefined;
+     * @docs-private */
+    protected explicitlyDisabled: boolean | undefined;
+
+    /**
+     * State a subclass derived from what it renders — the wrapped control being disabled here, text overflow
+     * in `KbqTitleDirective` and `KbqEllipsisCenterDirective`. Kept apart from `explicitlyDisabled` so both
+     * survive, whichever order they are written in.
+     * @docs-private */
+    protected derivedDisabled: boolean | undefined;
 
     /** Input (`kbqEnterDelay`) — delay in milliseconds before the tooltip is shown. Defaults to `400`. */
     // TODO: Skipped for migration because:
@@ -685,11 +694,9 @@ export class KbqTooltipTrigger
                 this.renderer.removeClass(nativeElement, 'kbq-tooltip-trigger_for-disabled');
             }
 
-            // An explicit `kbqTooltipDisabled` binding wins: the wrapper only derives the state when the
-            // consumer left the input alone.
-            if (this.explicitlyDisabled === undefined) {
-                this._disabled = !disabled;
-            }
+            // Derived, not assigned: an explicit `kbqTooltipDisabled` binding still wins, whichever of the
+            // two runs first.
+            this.setDerivedDisabled(!disabled);
         });
     }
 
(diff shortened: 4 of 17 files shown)


packages/components/tooltip/tooltip.component.ts
```diff
@@ -705,6 +712,37 @@ export class KbqTooltipTrigger
         super.ngOnDestroy();
     }
 
+    /**
+     * Records the state the subclass derived from its own content and re-folds it with the consumer's
+     * `kbqTooltipDisabled`. Subclasses call this instead of writing `disabled`, which would be read back as
```
lskramarov, line 721: **[1/15 · correctness · confirmed]** This contract is stated here but only honoured by two of the four subclasses.

`KbqOptionTooltip` still writes `this.disabled` at `select/select-option.directive.ts:53, 58, 64, 82, 86`, so its overflow verdict lands in `explicitlyDisabled` — the very clobbering this PR removes elsewhere. There is a live victim in this repo:

`datepicker/calendar-header.html:24` renders `<kbq-option [disabled]="month.disabled" [kbqTooltipDisabled]="true" [value]="month.value">`, and `KbqCalendarHeader` imports `KbqSelectModule`, which declares `KbqOptionTooltip` on selector `kbq-option`. Angular applies the input first (`explicitlyDisabled = true`), then `ngAfterViewInit` runs `this.disabled = !this.isOverflown` and overwrites it with `false` for a clipped month name; every `mouseenter` re-clobbers it. So the datepicker's explicit suppression fails precisely in the case it was added for. Same at `:51` for the year list.

Migrating is mechanical: five `this.disabled = X` → `this.setDerivedDisabled(X)` plus the same `foldDisabled()` override.

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -120,6 +144,14 @@ export class KbqEllipsisCenterDirective extends KbqTooltipTrigger implements OnI
         super.ngOnDestroy();
     }
 
+    /**
+     * The hint spells out what the split hid, so `kbqTooltipDisabled="false"` cannot conjure one for a name
+     * that fits — unlike the base, where an explicit value wins outright.
+     * @docs-private */
+    protected override foldDisabled(): boolean {
```
lskramarov, line 152: **[2/15 · correctness · confirmed]** This override activates a binding that used to be inert, and it changes shipped behaviour.

`file-upload/multiple-file-upload.component.html:54` has `[kbqTooltipDisabled]="disabled"` on the `kbqEllipsisCenter` span. Before: the input set `explicitlyDisabled`, then `refresh()`'s `setTimeout` ran `this.disabled = !truncated` and overwrote it, so a disabled multiple upload still revealed the full name of a middle-truncated file on hover. After: `explicitlyDisabled = true` survives the fold, so that hint is gone — the user of a disabled upload can no longer read a truncated file name at all.

`single-file-upload.component.html:27` binds no `kbqTooltipDisabled`, so the two variants now behave differently for the same reason.

Worth confirming this is intended (a disabled control still displays text the user cannot read), and either way the two variants should agree. Nothing covers it: `file-upload.spec.ts:879` (`describe('with ellipsis in the center')`) opens with `component.disabled = false` and exists only under the single variant.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -685,11 +694,9 @@ export class KbqTooltipTrigger
...
             }
 
-            // An explicit `kbqTooltipDisabled` binding wins: the wrapper only derives the state when the
-            // consumer left the input alone.
-            if (this.explicitlyDisabled === undefined) {
-                this._disabled = !disabled;
-            }
+            // Derived, not assigned: an explicit `kbqTooltipDisabled` binding still wins, whichever of the
+            // two runs first.
+            this.setDerivedDisabled(!disabled);
```
lskramarov, line 703: **[3/15 · correctness · confirmed]** The removed `if (this.explicitlyDisabled === undefined)` guard kept two producers in separate slots. Routing this through `setDerivedDisabled()` funnels both into one `derivedDisabled` field, where they overwrite each other.

`forDisabledComponent` is a plain inherited `input()` (line 297), so it is bindable on `[kbqEllipsisCenter]` and `[kbq-title]` hosts too. With `<span [kbqEllipsisCenter]="longName" [forDisabledComponent]="ctrl">`, `refresh()` sets `derivedDisabled = false`, then any flip of `ctrl.disabledSignal()` re-runs this effect and overwrites it — a genuinely clipped name loses its hint until the next resize or text change, and `KbqEllipsisCenterDirective` has no hover recompute path to recover it. Pre-PR the subclass write defined `explicitlyDisabled` and the guard locked this effect out for good.

Secondary: `applyDisabled()` calls `hide()`, and `KbqTooltipTrigger.hide()` (line 794) clears `mutedBySiblingPopup`. So a wrapped-control state flip can now un-mute a tooltip a sibling pop-up had muted — the old direct `_disabled` write never did that.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -705,6 +712,37 @@ export class KbqTooltipTrigger
         super.ngOnDestroy();
     }
 
+    /**
+     * Records the state the subclass derived from its own content and re-folds it with the consumer's
+     * `kbqTooltipDisabled`. Subclasses call this instead of writing `disabled`, which would be read back as
+     * the consumer's own value and lose whichever of the two was written first.
```
lskramarov, line 722: **[4/15 · correctness · confirmed]** Same non-migration as finding 1, fourth subclass: `timezone/timezone-option.directive.ts`.

`checkTooltipDisabled()` writes `this.disabled = count <= TOOLTIP_VISIBLE_ROWS_COUNT` (line 94) through the public setter, from `ngAfterViewInit`, from the debounced `resizeObserver` and `contentObserver` streams, and from `mouseenter`. `handleElementLeave()` (line 86) writes `this.disabled = true` on every pointer exit.

So `<kbq-timezone-option kbqTooltipDisabled>` is silently defeated, and on this host `explicitlyDisabled` is never `undefined` — the base fold can never fall through to any derived producer.

lskramarov (commented):
Automated review (max effort: 10 finder angles + verification + gap sweep). The affected Jest suites were run and all pass — nothing existing regresses.

The arbitration split itself looks right, and it genuinely fixes the `[kbqTooltipDisabled]` clobbering on `kbq-title` and `kbqEllipsisCenter`. Two things stand out:

1. **The fix is applied to two of four subclasses.** `KbqOptionTooltip` and `KbqTimezoneOptionTooltip` still write `this.disabled` directly, and `datepicker/calendar-header.html:24` is a live in-repo victim — its `[kbqTooltipDisabled]="true"` is defeated exactly when it matters.
2. **A previously inert binding now takes effect.** `multiple-file-upload.component.html:54` means a disabled multiple upload loses the full-name hint for middle-truncated names, while the single upload keeps it. That is a user-visible change with no test and no changelog line (the `chore` type is dropped by the release transform).

15 findings inline, most severe first.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -403,19 +408,23 @@ export class KbqTooltipTrigger
 
     set disabled(value) {
         this.explicitlyDisabled = coerceBooleanProperty(value);
```
lskramarov, line 410: **[5/15 · correctness · plausible]** `coerceBooleanProperty` collapses `undefined`/`null` to `false`, so the `undefined` sentinel the base `foldDisabled()` arbitrates on is unreachable once the input appears in a template at all.

`<div kbqTooltip="…" [forDisabledComponent]="btn" [kbqTooltipDisabled]="cfg.suppressHint">` with `cfg.suppressHint` unset: Angular calls the setter on the first binding pass with `undefined`, `explicitlyDisabled` becomes `false`, and `foldDisabled()` short-circuits on the first `??` operand forever. The tooltip is pinned enabled and `forDisabledComponent` can never disable it — while the same effect still removes `role="group"`, sets `tabindex="-1"` and drops `kbq-tooltip-trigger_for-disabled`, so the a11y state and the hint disagree.

The repo already ships the right transform for this exact trap — `navbar-item.component.ts:531`:

```ts
transform: (value: unknown) => (value === undefined ? undefined : booleanAttribute(value))
```

packages/components/ellipsis-center/ellipsis-center.scss
```diff
@@ -0,0 +1,31 @@
+// Layout contract for the two spans `KbqEllipsisCenterDirective` renders. Loaded once by the directive
+// itself through `_CdkPrivateStyleLoader`, so a host no longer copies these declarations — every copy was a
+// chance for the directive to behave differently depending on where it was used.
+.kbq-ellipsis-center {
```
lskramarov, line 4: **[6/15 · correctness · plausible]** The selector drops from `.kbq-file-upload .kbq-ellipsis-center` (0,2,0, inside a component stylesheet) to a flat `.kbq-ellipsis-center` (0,1,0) that `_CdkPrivateStyleLoader` appends to the end of `<head>` at runtime.

Nothing in this repo loses — I checked every rule touching `.kbq-file-item__text` and the breadcrumb hosts, and they set only `color`/`flex` at 0,4,0+. The risk is consumer-facing: anyone who followed the old contract (the JSDoc this PR deletes said the host ships it) with a flat rule in `styles.scss` — say `.kbq-ellipsis-center { display: inline-flex; overflow: visible; }` — has it in a build-time `<link>`, and Angular appends component styles to `document.head` afterwards, so `display: flex; overflow: hidden` now wins the tie. Their layout changes on upgrade with no code change of their own, and there is no token or opt-out class — only `!important` or added specificity.

Worth a line in the migration guide, at minimum.

packages/components/title/title.directive.ts
```diff
@@ -273,13 +273,22 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
...
 
     /** Always disables (hides) the tooltip. Bound to `mouseleave` and non-keyboard focus changes.
      * @docs-private */
     protected hideTooltip() {
-        this.disabled = true;
+        this.setDerivedDisabled(true);
+    }
+
+    /**
+     * The tooltip repeats text the host had to clip, so `kbqTooltipDisabled="false"` cannot conjure one for a
```
lskramarov, line 286: **[7/15 · conventions · confirmed]** This paragraph describes a change to the documented public meaning of `kbqTooltipDisabled`, but it lives only in a `@docs-private` JSDoc. Three places that should record it do not:

- `tooltip.en.md:119` and `tooltip.ru.md:119` still document one flat contract ("Suppresses the tooltip and hides it if it is open" / "Отключает тултип…"). Both languages are mandatory.
- `docs/guides/migration.{en,ru}.md` — the open 21.0.0 list already carries the sibling component reviews; this gets no entry.
- The `chore` commit type makes it invisible at release time: `packages/cli/src/release/changelog-writer-options.spec.ts:126` pins that the transform returns `undefined` for a `chore` commit with no notes, so this ships with no changelog line at all.

AGENTS.md: *"A breaking change to a component ships with a migration there, or at least a warning that points at the manual steps."* The sibling review PRs on this branch line use `fix(...)!` for exactly this.

packages/components/ellipsis-center/ellipsis-center.directive.ts
```diff
@@ -1,28 +1,43 @@
...
+class EllipsisCenterStyleLoader {}
+
 /**
  * Renders text as two spans and moves the ellipsis into the middle, so the tail — a file extension, the
  * digits that tell two reports apart — stays readable when the host is too narrow for the whole string. A
  * tooltip spells out the full text, and is enabled only while something is actually hidden.
  *
- * The host needs the layout contract the consuming component ships (see `.kbq-ellipsis-center` in
- * `file-upload.scss`): the two spans have no styles of their own.
+ * The tooltip's `forDisabledComponent` has no meaning here: the hint always repeats the text this directive
```
lskramarov: **[8/15 · correctness · plausible]** The comment asserts this, but nothing enforces it — `forDisabledComponent` is inherited unchanged and the base effect still acts on it.

Binding `[forDisabledComponent]` on a `[kbqEllipsisCenter]` host runs `tooltip.component.ts:678-699`, which stamps `tabindex="0"`, `role="group"` and `kbq-tooltip-trigger_for-disabled` onto a span whose only children are the two truncation halves — a decorative text node in the tab order, with a group role and no accessible explanation — on top of clobbering the truncation verdict (finding 3). The same applies to `KbqTitleDirective`.

Either enforce the claim or drop it from the doc: as written it tells a reader the input is inert when it is actively harmful.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -705,6 +712,37 @@ export class KbqTooltipTrigger
...
+     * Subclasses whose derived state means *there is nothing to show* override this — a hint that repeats
+     * text the host had to clip cannot be conjured for text that was never clipped.
+     * @docs-private */
+    protected foldDisabled(): boolean {
+        return this.explicitlyDisabled ?? this.derivedDisabled ?? false;
+    }
+
+    /** Recomputes the effective `disabled` and closes an open tooltip that just lost its reason to be open. */
+    private applyDisabled(): void {
+        this._disabled = this.foldDisabled();
```
lskramarov: **[9/15 · efficiency · confirmed]** No transition guard: this re-assigns `_disabled` and re-runs `hide()` even when the folded value did not move.

Measured on this head with a 10-item `kbq-title` list, one hover cycle each: **30 `hide()` calls, all 30 with `instance === null`**, and 53 `NgZone.run` entries. Because `triggerName` rests on `'mouseleave'` after any pointer sweep, each no-op `hide()` still falls into `this.ngZone.run(() => this.instance?.hide(delay))` (`pop-up-trigger.ts:465`) — and CDK builds `SharedResizeObserver` inside `runOutsideAngular`, so that re-entry is what makes `checkStable()` schedule an `ApplicationRef.tick()`. Once per visible title per debounced resize tick, on a directive that sits on every dropdown item, list option and tree option.

The waste predates the PR, but `applyDisabled()` is now the single place a guard removes all of it:

```suggestion
        const next = this.foldDisabled();

        if (next === this._disabled) return;

        this._disabled = next;
```

packages/components/tooltip/tooltip.component.ts
```diff
@@ -705,6 +712,37 @@ export class KbqTooltipTrigger
...
+
+        if (this._disabled) {
```
lskramarov, line 754: **[10/15 · test-coverage]** This `hide()` is the genuinely new base-class behaviour — the old code was `if (this.explicitlyDisabled === undefined) { this._disabled = !disabled; }` with no `hide()` — and it ships with no spec.

Neither `describe('forDisabledComponent input')` (`tooltip.spec.ts:311`) nor `describe('forDisabledComponent precedence')` (`tooltip.spec.ts:1579`) ever opens the tooltip before flipping `controlDisabled`, and both pin a static `[kbqTooltipDisabled]="true"` — so neither reaches the `??` fold's `explicitlyDisabled === false` + derived `true` row either. Dropping this `hide()` again, or reverting `foldDisabled()`, leaves the whole tooltip suite green.

The PR touches no file under `packages/components/tooltip/` other than `tooltip.component.ts`.

packages/components/title/title.directive.ts
```diff
@@ -273,13 +273,22 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
...
+     * title that is fully visible — unlike the base, where an explicit value wins outright. A consumer's
+     * `kbqTooltipDisabled="true"`, in turn, is no longer overwritten by the next hover.
+     * @docs-private */
+    protected override foldDisabled(): boolean {
```
lskramarov, line 290: **[11/15 · test-coverage]** `KbqTitleDirective` gets the same user-visible behaviour change as `KbqEllipsisCenterDirective`, but with zero new specs — the ellipsis-center side got three.

`title.directive.spec.ts` contains no `kbqTooltipDisabled` binding at all; its 16 `directive.disabled` assertions are all overflow-derived. The two behaviours this override's own JSDoc promises — `kbqTooltipDisabled="true"` surviving a hover or keyboard focus that used to overwrite it, and `kbqTooltipDisabled="false"` not conjuring a hint for a title that fits — are unprotected: reverting this method to the base fold leaves the entire title suite green.

The three specs added at `ellipsis-center.directive.spec.ts:184-224` are directly portable.

packages/components/ellipsis-center/ellipsis-center.scss
```diff
@@ -0,0 +1,31 @@
+// Layout contract for the two spans `KbqEllipsisCenterDirective` renders. Loaded once by the directive
```
lskramarov, line 1: **[12/15 · test-coverage]** The relocated contract has no e2e coverage of its own — `packages/components/ellipsis-center/` has no `e2e.ts`, no Playwright spec, no `__screenshots__`, and `grep -i ellipsis packages/e2e/routes.ts` returns nothing.

The regression this PR must prevent is "a host that does not ship the contract renders unsplit, overflowing text". The only guard is `file-upload/e2e.playwright-spec.ts`, whose locators run inside markup that still carries a `.kbq-file-upload` wrapper — so it would pass even for a host outside file-upload. Meanwhile `breadcrumbs-truncate-center-items-example.ts` loses 26 lines of styles with zero automated coverage (`packages/components/breadcrumbs/e2e.ts` never mentions ellipsis).

An `E2eEllipsisCenter*` scenario on a bare host — no file-upload, no button — is what pins the loader actually firing.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -403,19 +408,23 @@ export class KbqTooltipTrigger
...
-        this._disabled = this.explicitlyDisabled;
 
-        if (this._disabled) {
-            this.hide();
-        }
+        this.applyDisabled();
     }
 
     /**
      * Value the consumer assigned to `kbqTooltipDisabled`, or `undefined` while the input was never set. It
```
lskramarov: **[13/15 · conventions · confirmed]** This JSDoc was carried over unchanged while the declaration under it was edited, and it is now wrong on both halves.

- "makes the input win" — the two subclass overrides make an explicit `false` lose outright.
- "over the state `forDisabledComponent` derives" — `derivedDisabled`, three lines below, is documented as also holding text-overflow state, so it is no longer only that.

This block is the field's only explanation of the arbitration rule, and it now describes the pre-PR rule; the next subclass author will read it and assume `kbqTooltipDisabled` always wins. AGENTS.md: *"Update or remove comments that become outdated because of your changes."*

packages/components/title/title.directive.ts
```diff
@@ -273,13 +273,22 @@ export class KbqTitleDirective extends KbqTooltipTrigger implements AfterViewIni
...
+        return this.explicitlyDisabled === true || (this.derivedDisabled ?? false);
```
lskramarov, line 291: **[14/15 · reuse]** This is byte-identical to `ellipsis-center.directive.ts:152`, under near-identical JSDoc, and each copy adds an entry to its golden API file.

The two subclasses from findings 1 and 4 have the same semantics and would need a third and fourth copy. Nothing type-checks the copies against each other, and a subclass that forgets the override silently inherits the *opposite* precedence.

A `protected readonly derivedMeansNothingToShow = true` consulted by the single base `foldDisabled()` states the policy as data in one place, removes both overrides from the API guard, and lets `KbqOptionTooltip`/`KbqTimezoneOptionTooltip` opt in with one line each.

packages/components/tooltip/tooltip.component.ts
```diff
@@ -705,6 +712,37 @@ export class KbqTooltipTrigger
         super.ngOnDestroy();
     }
 
+    /**
+     * Records the state the subclass derived from its own content and re-folds it with the consumer's
+     * `kbqTooltipDisabled`. Subclasses call this instead of writing `disabled`, which would be read back as
+     * the consumer's own value and lose whichever of the two was written first.
+     * @docs-private */
+    protected setDerivedDisabled(value: boolean): void {
```
lskramarov, line 724: **[15/15 · reuse]** `protected` closes this to the repo's two existing owners of exactly this fold, which drive the trigger by *composition* rather than inheritance.

`navbar-item.component.ts:661` runs `this.tooltip.disabled = this.tooltipDisabled() ?? (!this.isCollapsed() && !this.hasCroppedText)` from an effect on every collapse or title change (same shape at `navbar-brand.component.ts:209`), via `inject(KbqTooltipTrigger, { self: true })`. After the first run `explicitlyDisabled` is permanently defined on every navbar item, so the base fold discards every derived producer on that trigger.

The `hostDirectives` block at `navbar-item.component.ts:471` states the intent outright — *"Composition, not inheritance: the item is not a tooltip trigger, it merely owns one"* — and that is the case the new API has no channel for.

packages/components/tooltip/tooltip.component.ts
NikGurev: Проблему подтверждаю, и в датапикере она действительно видна: подсказка там выключена в разметке явно, но всё равно показывается.

Только это не регрессия от этого PR. На базовой ветке сеттер вёл себя ровно так же, и пункт списка точно так же затирал указание разработчика — отключение в календаре не работало никогда, с момента, как его туда добавили. Этот PR его не сломал и не починил, он лишь объявил правило, которое пункт списка пока не соблюдает.

Вынес отдельной задачей, вместе с 4, 5 и 14 — переводить компоненты по одному смысла нет. Сам баг датапикера выделил в отдельную задачу с приоритетом выше.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Согласен — это единственное, что этот PR реально изменил для пользователя.

Убрал привязку из множественной загрузки. Читать имя файла, которое не поместилось, — это не взаимодействие с контролом, и блокировка не должна такую возможность отбирать. Теперь оба варианта ведут себя одинаково.

Добавил тест, который это фиксирует.

packages/components/tooltip/tooltip.component.ts
NikGurev: Механику подтверждаю, но воспроизвести её нечем.

Директива сама заполняет содержимое элемента — создаёт и подменяет внутри два куска текста. Обернуть ею контрол нельзя, он будет затёрт. А текст подсказки у неё всегда её собственный, обрезанный, а не объяснение, почему кнопка недоступна. Это разные назначения, и совместить их не получится даже при желании. В репозитории такой связки нет нигде — все четыре использования на обычной подсказке.

Решили не закрывать это кодом. Вместо этого поправил описание директивы: там теперь не утверждение, что вход бездействует, а прямое указание, что привязывать его сюда не нужно, и почему.

Побочный эффект с закрытием подсказки уходит сам после правки из 9 — лишние повторные вызовы туда больше не доходят.

packages/components/tooltip/tooltip.component.ts
NikGurev: То же самое, что и в 1 — проблема старая, не от этого PR. Ушло в общую задачу.

packages/components/tooltip/tooltip.component.ts
NikGurev: Верно, и обратите внимание: эта строка в PR не менялась, поведение такое же было и раньше.

Правка тут задевает все наследники сразу и зафиксированные описания API, так что уходит в ту же задачу, что 1, 4 и 14. Готовый образец с сохранением `undefined` в навбаре я в задаче указал.

packages/components/ellipsis-center/ellipsis-center.scss
NikGurev: Арифметику подтверждаю. Проверил независимо, что внутри репозитория ничего не проигрывает — правила на этом элементе задают только цвет и растягивание, пересечений по свойствам с новыми правилами нет ни одного.

Но понижение специфичности само по себе сломать чужое переопределение не может: селектор, который бил старый вариант, тем более бьёт новый. Переопределения от этого могут только начать работать, но не перестать.

Настоящее изменение здесь другое — стили появились там, где их раньше не было вовсе. Сценарий узкий: чтобы в него попасть, надо было самостоятельно повторить ровно тот же набор свойств, который директива теперь и так даёт. Решили в документацию это не выносить.

packages/components/title/title.directive.ts
NikGurev: Обсудили и решили ничего не писать.

Публичный API не менялся, ничего не удалено и не переименовано — чужой код от этого не перестанет собираться. Вход начинает делать то, что за ним и было записано.

packages/components/ellipsis-center/ellipsis-center.directive.ts
NikGurev: Справедливо — как было написано, это утверждение ничем не обеспечивалось.

Кодом закрывать не стали, поэтому поправил формулировку. Теперь там не «вход не имеет смысла», а прямое «привязывать его сюда не нужно» с объяснением, почему: подсказка здесь всегда повторяет собственный обрезанный текст, и оба источника решения пишут в одно место.

packages/components/tooltip/tooltip.component.ts
NikGurev: Принял, добавил.

packages/components/tooltip/tooltip.component.ts
NikGurev: Принял. Добавил два теста.

Первый открывает подсказку и снимает блокировку с контрола — подсказка должна закрыться сразу, а не ждать, пока уйдёт курсор. Проверил, что без правки он падает.

Второй закрывает тот случай, где явное указание и вычисленное состояние расходятся, — до этого он не был покрыт ни одним из существующих блоков.

packages/components/title/title.directive.ts
NikGurev: Принял, перенёс. Добавил два теста на заголовок — ровно те два поведения, которые обещает описание метода.

Проверил откатом: если убрать переопределение, один из них падает.

packages/components/ellipsis-center/ellipsis-center.scss
NikGurev: Сделал.

Добавил отдельный сценарий на голом элементе — без загрузки файлов, без кнопок, без единого стиля от хоста. Шесть проверок: что стили применились, что текст разрезан и расширение видно, что элемент не свернулся в перенос строки, что подсказка показывает полный текст, что при коротком тексте подсказки нет, и случай с минимальной длиной.

Проверок на скриншотах не делал — у соседней директивы заголовка их тоже нет, а нужное здесь проверяется утверждениями напрямую.

Отдельно убедился, что сценарий действительно ловит регрессию: временно отключил загрузку стилей — упало пять проверок из шести, прошла только та, что и должна была.

packages/components/tooltip/tooltip.component.ts
NikGurev: Принял, переписал. Обе неточности были на месте.

packages/components/title/title.directive.ts
NikGurev: С направлением согласен, но вынести это в один флаг имеет смысл только когда на новый механизм переведены все четыре компонента. Иначе получится ещё одно промежуточное состояние.

Ушло в общую задачу вместе с 1, 4 и 5.

packages/components/tooltip/tooltip.component.ts
NikGurev: Случай верный, оба места в навбаре подтверждаю.

Это расширение публичного API — нужно отдельное решение и обновление зафиксированных описаний. Вынес отдельной задачей.

lskramarov (approved):

PR #2027, opened 2026-09-10
fix(sidepanel)!: errors following a full review of the component

NikGurev:
## Summary

The sidepanel's non-modal mode is documented as opening "without page blocking", but every panel — backdrop
or not — got CDK's `BlockScrollStrategy`, which pins the document with `position: fixed`. That is the
defect the review is named after: a default that contradicts the mode it describes. The scroll strategy
follows modality now, and `KbqSidepanelConfig` gained a `scrollStrategy` factory for the panels that
actually wanted the page frozen.

The container also announced nothing. It rendered no `role`, no `aria-modal` and no accessible name, so a
screen-reader user got an unlabelled region and could walk the page behind an open modal panel with the
virtual cursor. It renders `role="dialog"`, `aria-modal` and an `aria-labelledby` pointing at the
`kbq-sidepanel-header` title, and a modal sidepanel marks the rest of the page `aria-hidden` while it is
open.

The third defect had the widest blast radius and nothing to do with sidepanels: `KbqSidepanelModule`
provided `ConfigurableFocusTrapFactory` for the `providedIn: 'root'` `FocusTrapFactory` token, so importing
the sidepanel module swapped the focus-trap implementation for **every** trapping component in the
application — modal, dropdown, popover. The container keeps the override for itself, because the inert
strategy that pulls escaping focus back is what the sidepanel wanted; nothing else in the package asked
for it.

The second commit is a follow-up pass over the ownership of the shared bookkeeping the first one
introduced. The one worth naming is `ngOnDestroy`: it restored the page's `aria-hidden` state from any
injector level, so destroying a component-provided `KbqSidepanelService` revealed the page under a modal
sidepanel that a different instance had opened, and cleared the state that would have re-hidden it.

`ng update` reporting ships as the `sidepanel-non-modal-behavior` schematic, registered for `21.0.0-0`.
It is warn-only — see below.

## List of notable changes:

- **fixed** the scroll strategy, which follows modality instead of blocking the page under every panel:
  `block()` with a backdrop, `reposition()` without. `KbqSidepanelConfig` gained `scrollStrategy`, and the
  block survives closing the lower panel of a stack instead of being released and re-taken
- **added** the container ARIA layer — `role="dialog"`, `aria-modal`, and `aria-labelledby` resolved
  against the `kbq-sidepanel-header` title, with `ariaLabel`/`ariaLabelledBy` on the config. A modal
  sidepanel hides the rest of the page from assistive technology while it is open
- **fixed** `trapFocusAutoCapture`, which defaults to `true` in both modalities rather than following
  `hasBackdrop`, and Escape now calls `preventDefault()`. Both guides scope the focus-trap claim to modal
- **fixed** the `FocusTrapFactory` override: scoped to `KbqSidepanelContainerComponent` instead of being
  swapped application-wide by `KbqSidepanelModule`
- **fixed** `close()`, which guards on a dedicated flag so the first result wins, and the config merge,
  which no longer lets `undefined` values overwrite the defaults
- **fixed** `closeSidepanels()`, which no longer reverses the live stack in place, and the indent strip,
  recomputed when a lower panel closes. An outside click closes only the panels of the same position stack
- **fixed** the lifecycle: a component-level `KbqSidepanelService` closes the sidepanels it opened when it
  is destroyed instead of leaving an orphaned overlay. The service is `providedIn: 'root'`, and
  `KbqSidepanelModule` still provides it, so existing per-module instances are unchanged
- **fixed** `KbqSidepanelClose`, which injects `KbqSidepanelRef` non-optionally — the `setTimeout` fallback
  that looked the ref up by container id is gone, along with `ngOnInit`
- **updated** the styles: the package styles the component-portal host, so the four hand-rolled
  `host: { class: 'layout-column flex' }` workarounds are deleted; the dark backdrop rule drops
  `!important` and stops repainting backdrops the package does not own; the dead overlay-background token
  is consumed and the unstyled `kbq-sidepanel-overlay` class is dropped
- **updated** the public API: `KbqSidepanelAnimationState` is exported, so the container's
  `animationState` and `setAnimationState()` resolve to a type consumers can see instead of tripping
  `ae-forgotten-export`. The animation trigger and the per-position transform table stay internal to the
  container. The template context and the animation state are typed
- **added** focus coverage that reads `document.activeElement` after the panel opens — restoration,
  `cdkFocusInitial` and `trapFocusAutoCapture` — plus a DOM-asserted Playwright tab-order scenario and a
  top/bottom screenshot row
- **updated** both guides: the keyboard table marks what the component actually implements, the containing
  block an open panel creates is documented, and the dynamic-config section names only `disableClose` —
  the one config property re-read after open

- **The schematic writes nothing, deliberately.** `packages/schematics/src/migrations/sidepanel-non-modal-behavior/`
  reports and never touches the tree. Whether a given non-modal panel wanted the page frozen, and whether
  an application had come to depend on the leaked focus-trap implementation, are not derivable from the
  call site — there is no single mechanical translation for either. The README carries the
  `Pattern → Manual migration` table for the five reachable shapes and a "no call site to point at"
  section for the rest.
- **`trapFocusAutoCapture` defaulting to `true` for non-modal panels** is the change most likely to
  surprise. A non-modal panel now captures focus on open and returns it to the trigger on close. That is
  the behaviour the config property's name promises, but it is a behaviour change with no call site the
  schematic can flag, so it lives in the report's notes.
- **Ownership of the `aria-hidden` bookkeeping** (second commit). The root service owns the state; a
  non-owning instance's `ngOnDestroy` no longer drops it. Worth a look if you disagree about where that
  state should live.
- **`setAriaLabelledBy` marks the container for check.** A `kbq-sidepanel-header` rendered after the
  panel's first change detection would otherwise leave `role="dialog"` with no accessible name.
- **The `KbqSidepanelClose` ref injection is now non-optional**, which is the sharpest edge in the API
  diff: a directive used outside a sidepanel throws at construction instead of silently doing nothing on
  a timeout.

- **The merge with state saving, in `closeSidepanels()`.** This branch was rebased onto a `main` that
  had gained `stateSavingKey` / `wasOpen()` in the meantime, and the two changes meet in one method.
  `main` marks each panel `bulkClosed` before closing it so a group close is not recorded as the user
  closing the panel; this branch stops `reverse()` from mutating the live stack. Both had to survive, so
  the loop is now `[...sidepanels].reverse().forEach(...)` with the marking kept inside it — deliberate,
  not a merge artifact. `ngOnDestroy` is the same story: it drops the assistive-technology bookkeeping
  only from the owning instance *and* unregisters the state-saving refs.

## Behaviour measured against `main`

The unit suite was written against the code this branch replaces, so it cannot by itself prove nothing
broke. Every behaviour below was measured on `main` and on the branch with the same browser probe driving
the dev app. Nine further rows — modal focus capture, Tab containment, Escape, focus restore to the
trigger, default panel width, the stacking offset, closing the top panel by its exposed strip, and the
pinned footer — measured identically on both sides and are not repeated here.

| Scenario | `main` | This branch |
| --- | --- | --- |
| Page behind a non-modal panel scrolls | no (0px; `cdk-global-scrollblock` applied) | yes (400px) |
| Focus moves into a non-modal panel on open | no | yes |
| Page content hidden from assistive technology behind a modal panel | nothing hidden | hidden |
| Container role | none | `dialog` |
| `aria-modal` | absent | `true` |
| Accessible name of the panel | none | the header title |

One row deserves a note because the number is unchanged and the mechanism is not: the footer stays pinned
to the bottom of a component sidepanel in both, but on `main` that depended on the consumer putting
`host: { class: 'layout-column flex' }` on its own content. The package styles the portal host now, and
the dev app's copy of that workaround is deleted in this PR — the measurement is what shows the removal
was safe.


packages/components/sidepanel/e2e.playwright-spec.ts
```diff
@@ -7,6 +7,9 @@ import {
     e2eWaitForSettledScrollbars
 } from '../../e2e/utils';
 
+// The committed baselines are WebKit renderings: WebKit is the browser whose compositing of the panel's
+// `transform` slide over the dim backdrop this suite was pinned to, and `scrollbar/deprecated` does the
+// same. Regenerate them with `yarn run e2e:docker:update-snapshots`.
 test.use({ browserName: 'webkit' });
```
artembelik: по хорошему этот метод нужно вызывать внутри describe для которого это требуется

packages/components/sidepanel/sidepanel-directives.ts
```diff
@@ -32,47 +18,31 @@ import { KbqSidepanelService } from './sidepanel.service';
...
+    /** The same value, under the camel-case spelling. */
     readonly kbqSidepanelClose = input<any>();
 
-    ngOnInit() {
+    constructor() {
         // A button with no type submits the form it sits in, so closing a sidepanel from inside a form would
-        // submit it too. A type the author set is left alone.
+        // submit it too. A type the author set is left alone. The element carries its static attributes by
+        // the time a directive is constructed, so this needs no lifecycle hook.
         if (!this.elementRef.nativeElement.hasAttribute('type')) {
```
artembelik: нужно вызывать после первого рендера, раньше вызывалось в ngOnInit

packages/components/sidepanel/sidepanel-ref.ts
```diff
@@ -34,21 +34,24 @@ export class KbqSidepanelRef<T = any, R = any> {
...
 
+    /** Whether `close()` has already started the closing animation. */
+    private closing = false;
+
     constructor(
         public readonly containerInstance: KbqSidepanelContainerComponent,
         public readonly overlayRef: OverlayRef,
         public readonly config: KbqSidepanelConfig
     ) {
         this.id = this.config.id || `kbq-sidepanel-${uniqueId++}`;
```
artembelik: вместо `uniqueId++` можно использовать `_IdGenerator` из cdk

artembelik (approved):

packages/components/sidepanel/sidepanel.service.ts
```diff
@@ -64,13 +64,42 @@ class KbqSidepanelStateSavingRef implements KbqStateSavingRef {
...
+    if (!source) return;
+
+    for (const key of Object.keys(source) as (keyof T)[]) {
+        if (source[key] !== undefined) {
+            target[key] = source[key];
+        }
+    }
+};
+
+@Injectable({ providedIn: 'root' })
```
artembelik: по идее больше не нужно провайдить сервис в компоненты, нужно проверить и удалить если это так

PR #2028, opened 2026-09-10
fix(time-range): validate range borders and bound the swap (#DS-4642)

NikGurev:
## Summary

- A reversed range (from > to) is swapped, not rejected — but never while a field is invalid.
- `minDate` / `maxDate` are exact moments, not whole days: to allow all of 31 Dec, pass 31 Dec 23:59.
- The allowed period is shown under the range option, and turns into an error when both ends fall outside it.
- A date outside the bounds turns both fields red; a time outside turns only the time red.
- Errors appear when you leave a field, and clear when you start typing.
- Apply blocks an invalid range and focuses the first bad field. Cancel drops it — reopening starts clean.
- The editor always opens on a valid range, one day long.


packages/components/time-range/time-range-editor.html
```diff
@@ -17,7 +17,7 @@
         @if (isRangeVisible()) {
             @let localeConfig = localeConfiguration();
 
-            <div class="kbq-time-range-editor__range" role="group">
+            <div class="kbq-time-range-editor__range" role="group" (focusout)="onRangeFocusOut($event)">
```
artembelik, line 20: почему не focusmonitor?

NikGurev: здесь origin не нужен, только факт выхода фокуса

\+ дополнительные подписки / отписки ни к чему

rmnturov:
Линии-разделители доводить бы до краев. Сейчас 1px отступ есть, такого не ожидается
image

packages/components/time-range/time-range.service.ts
```diff
@@ -85,9 +85,15 @@ export class KbqTimeRangeService<T> {
...
+     * opens on a range its own datepickers reject.
+     */
+    getDefaultRangeValue(minDate?: T | null, maxDate?: T | null): Required<KbqRangeValue<T>> {
+        const today = this.dateAdapter.today();
+        // Clamping is monotonic, so `from` cannot overtake `to`.
+        const from = this.clampToBounds(this.dateAdapter.addCalendarUnits(today, { days: -1 }), minDate, maxDate);
+        const to = this.clampToBounds(today, minDate, maxDate);
 
         return {
             fromTime: this.omitMilliseconds(from),
```
lskramarov: **`omitMilliseconds` runs *after* the clamp, so a clamped default can land below `minDate`.**

`clampToBounds` returns the `minDate` object itself, and every field is then passed through `omitMilliseconds`. If the consumer's `minDate` carries a millisecond component — `[minDate]="dateAdapter.today()"` is the obvious case, `DateTime.now()` always has one — `fromDate`/`fromTime` end up strictly *before* `minDate`:

```
minDate            2024-03-05T14:23:45.678
clamped from       2024-03-05T14:23:45.678
omitMilliseconds   2024-03-05T14:23:45.000
```

`KbqDatepickerInput.minValidator` compares with `compareDateTime`, which is millisecond-sensitive (`compareDateTime(min, value) === 678 > 0`), so the field opens with `kbqDatepickerMin` — the exact "editor opens on a range its own datepickers reject" this change sets out to fix. Only `minDate` is affected: stripping milliseconds can never push a value past `maxDate`.

Strip the milliseconds *before* clamping, so a clamped value is exactly the bound:

```ts
const today = this.omitMilliseconds(this.dateAdapter.today());
// Clamping is monotonic, so `from` cannot overtake `to`.
const from = this.clampToBounds(this.dateAdapter.addCalendarUnits(today, { days: -1 }), minDate, maxDate);
const to = this.clampToBounds(today, minDate, maxDate);

return { fromTime: from, fromDate: from, toTime: to, toDate: to };
```

The new spec cannot catch this: `createDate(2015, 0, 1)` has `millisecond === 0`.

packages/components/time-range/time-range.ts
```diff
@@ -149,7 +149,7 @@ export class KbqTimeRange<T> implements ControlValueAccessor, OnInit {
      * Used to calculate time range.
      * @docs-private */
     protected readonly normalizedDefaultRangeValue = computed(() => ({
-        ...this.timeRangeService.getDefaultRangeValue(),
+        ...this.timeRangeService.getDefaultRangeValue(this.minDate(), this.maxDate()),
```
lskramarov: **A reversed range arriving through the bound `formControl` is still surfaced as-is.**

The invariant "never show or emit a reversed range" is now enforced on two of three entry points — the edit path (`onRangeFocusOut`) and the emit path (`orderRange`). The write path is not. `checkAndCorrectTimeRangeValue` only checks that the type is available and that both ends are non-empty, so

```ts
control.setValue({ type: 'range', startDateTime: '2024-03-10T18:00:00Z', endDateTime: '2024-03-01T09:00:00Z' });
```

reaches `this.titleValue.set(corrected)` untouched and the trigger renders "10 Mar 2024 – 01 Mar 2024". `TestComponentWithRange` seeds an already-ordered value, so nothing covers it.

Ordering inside `checkAndCorrectTimeRangeValue` (or in `KbqTimeRangeService`) would put all three paths on one implementation.

packages/components/time-range/time-range-editor.ts
```diff
@@ -251,21 +255,56 @@ export class KbqTimeRangeEditor<T> implements ControlValueAccessor, Validator, O
...
+    private orderRange({ startDateTime, endDateTime }: KbqRange): KbqRange {
+        if (!startDateTime || !endDateTime) return { startDateTime, endDateTime };
+
+        const { dateAdapter } = this.timeRangeService;
+        const reversed =
+            dateAdapter.compareDateTime(dateAdapter.deserialize(startDateTime), dateAdapter.deserialize(endDateTime)) >
+            0;
+
+        return reversed ? { startDateTime: endDateTime, endDateTime: startDateTime } : { startDateTime, endDateTime };
+    }
```
lskramarov: **The guard clause rebuilds the object, so every non-`range` type now emits explicit `undefined` keys.**

`calculateTimeRange('allTime')` returns `{}` and `calculateTimeRange('lastHour')` returns `{ startDateTime }` — with no `endDateTime` key. Destructuring and returning an object literal materialises both keys, so the value handed to `onChange` is now `{ type, startDateTime: …, endDateTime: undefined }`.

That is observable: `KbqTimeRange.writeValue` branches on `'startDateTime' in value` (time-range.ts:208) to decide whether to emit `valueCorrected`. For a type whose range is empty the key used to be absent (`in` → `false`, emit) and is now present (`in` → `true`, no emit), so writing an applied value back into the control silently changes behaviour. Returning the argument untouched sidesteps the whole question.

```suggestion
    private orderRange(range: KbqRange): KbqRange {
        const { startDateTime, endDateTime } = range;

        if (!startDateTime || !endDateTime) return range;

        const { dateAdapter } = this.timeRangeService;
        const reversed =
            dateAdapter.compareDateTime(dateAdapter.deserialize(startDateTime), dateAdapter.deserialize(endDateTime)) >
            0;

        return reversed ? { startDateTime: endDateTime, endDateTime: startDateTime } : range;
    }
```

packages/components/time-range/time-range-editor.ts
```diff
@@ -251,21 +255,56 @@ export class KbqTimeRangeEditor<T> implements ControlValueAccessor, Validator, O
...
 
+    /**
+     * Once focus leaves the manual range fields, a reversed range is silently swapped back into order.
+     * @docs-private
+     */
+    protected onRangeFocusOut({ currentTarget, relatedTarget }: FocusEvent): void {
+        const next = relatedTarget as HTMLElement | null;
+
+        // The datepicker calendar is a separate overlay, so focus moving into it is still editing.
+        if ((currentTarget as HTMLElement).contains(next) || next?.closest('.kbq-datepicker__popup')) return;
```
lskramarov: **`contains(next)` is defeated whenever `relatedTarget` is `null`.**

`focusout` carries `relatedTarget === null` whenever focus goes nowhere: the window loses focus (alt-tab, devtools, another app), or the user mousedowns on a non-focusable spot — including the "From"/"To" labels and the padding *inside this very block*. `contains(null)` is `false` and `null?.closest()` is `undefined`, so the handler falls through and reorders the fields while the user is still in the popover, mid-edit.

Bailing out on a null `relatedTarget` is cheap, and `orderRange` already guarantees the *applied* value is ordered even when the swap never runs:

```ts
if (!next || (currentTarget as HTMLElement).contains(next) || next.closest('.kbq-datepicker__popup')) return;
```

Worth a check in Safari either way: clicking a `<button>` does not move focus there, so the "focus moves straight to apply" scenario reaches this handler with `relatedTarget === null` rather than with the button.

packages/components/time-range/time-range.service.ts
```diff
@@ -85,9 +85,15 @@ export class KbqTimeRangeService<T> {
...
-        const from = this.dateAdapter!.addCalendarUnits(this.dateAdapter!.today(), { days: -1 });
-        const to = this.dateAdapter!.today();
+    /**
+     * "Yesterday to today", pulled inside `minDate`/`maxDate` when they are given - otherwise the editor
+     * opens on a range its own datepickers reject.
+     */
+    getDefaultRangeValue(minDate?: T | null, maxDate?: T | null): Required<KbqRangeValue<T>> {
+        const today = this.dateAdapter.today();
+        // Clamping is monotonic, so `from` cannot overtake `to`.
+        const from = this.clampToBounds(this.dateAdapter.addCalendarUnits(today, { days: -1 }), minDate, maxDate);
```
lskramarov: **When the whole "yesterday → today" window falls outside the bounds, both ends clamp to the same instant.**

With the documented 2015–2017 example (`time-range-min-max-example.ts`), `today()` and `today() - 1 day` are both past `maxDate`, so `from === to === maxDate` and the editor opens on a zero-length range (31.12.2017 00:00 → 31.12.2017 00:00). In bounds, but not a usable default — and the same collapse happens for any `minDate` in the future.

Deriving `from` from the already-clamped `to` keeps the one-day window wherever the bounds allow it:

```ts
const to = this.clampToBounds(today, minDate, maxDate);
const from = this.clampToBounds(this.dateAdapter.addCalendarUnits(to, { days: -1 }), minDate, maxDate);
```

`from <= to` still holds, so the "clamping is monotonic" note stays true.

packages/components/time-range/time-range-editor.ts
```diff
@@ -135,21 +136,24 @@ export class KbqTimeRangeEditor<T> implements ControlValueAccessor, Validator, O
...
-                    nonNullable: true
-                }),
-                fromTime: new FormControl<T>(defaultRangeValue.fromTime, { nonNullable: true }),
-                fromDate: new FormControl<T>(defaultRangeValue.fromDate, { nonNullable: true }),
-                toTime: new FormControl<T>(defaultRangeValue.toTime, { nonNullable: true }),
-                toDate: new FormControl<T>(defaultRangeValue.toDate, { nonNullable: true })
-            },
-            { validators: rangeValidator(this.timeRangeService) }
-        );
+        this.form = new FormGroup({
```
lskramarov, line 185: **`RangeErrorStateMatcher`'s `form?.invalid` branch is a leftover now that the group validator is gone.**

```ts
isErrorState(control, form) { return !!form?.invalid || !!control?.invalid; }
```

`form` here is the `FormGroupDirective` of the whole editor form. The `form.invalid` half existed so that a group-level `fromIsGreaterThanTo` would paint *both* date fields red. With the group carrying no validators, it can only be invalid because one of its controls is — so the branch now just smears one field's error onto the other: a `kbqDatepickerMin` on `fromDate` marks `toDate` invalid too. `control?.invalid` alone is enough.

lskramarov (commented):
## Code review — `fix(time-range): add proper range swap (#DS-4642)`

Swapping instead of invalidating is the right call, and the `toDate: toTime` copy-paste fix in `mapTimeRange` is a good catch. Eleven notes below; the first five are the ones worth acting on.

**Correctness**

1. `getDefaultRangeValue` strips milliseconds *after* clamping, so a `minDate` carrying milliseconds (e.g. `dateAdapter.today()`) yields a default the datepicker's `minValidator` rejects — the very failure this overload is meant to prevent.
2. A reversed range written in through the bound `formControl` is never normalised; only the edit and emit paths are covered.
3. `orderRange`'s guard clause materialises `startDateTime`/`endDateTime` keys that used to be absent, flipping the `'startDateTime' in value` check in `KbqTimeRange.writeValue`.
4. The `contains(next)` guard does not hold when `relatedTarget` is `null` (window blur, or a click on a non-focusable spot inside the block).
5. When both ends clamp to the same bound the editor opens on a zero-length range — reproducible with the documented 2015–2017 example.

**Cleanup left behind by the removed group validator**

6. `RangeErrorStateMatcher`'s `form?.invalid` branch now only smears one control's error onto the other field.
7. `this.form.errors` is permanently `null`; `validate()` and `concatControlValidationErrors()` still read it.
8. Reversal is decided twice, in two representations, with a `toIso8601` → `deserialize` round trip per keystroke.
9. The editor's own `rangeValue` default, and the two service-internal fallbacks, still call `getDefaultRangeValue()` without bounds.

**Tests**

10. The datepicker-overlay branch is verified against a hand-built element, so the selector is not pinned to the real `panelClass`.
11. The min/max test passes on both of the clamp defects above.

_Generated with Claude Code_

packages/components/time-range/time-range-editor.ts
```diff
@@ -135,21 +136,24 @@ export class KbqTimeRangeEditor<T> implements ControlValueAccessor, Validator, O
...
+            type: new FormControl<KbqTimeRangeType>(this.timeRangeService.DEFAULT_RANGE_TYPE, {
+                nonNullable: true
+            }),
+            fromTime: new FormControl<T>(defaultRangeValue.fromTime, { nonNullable: true }),
+            fromDate: new FormControl<T>(defaultRangeValue.fromDate, { nonNullable: true }),
+            toTime: new FormControl<T>(defaultRangeValue.toTime, { nonNullable: true }),
+            toDate: new FormControl<T>(defaultRangeValue.toDate, { nonNullable: true })
+        });
```
lskramarov, line 193: **`this.form.errors` is now permanently `null`, and two call sites still read it.**

With the group validator dropped the group has no validators of its own, so:

- `validate()` → `this.form.errors || this.lastValidationErrorOnEmit` — the left operand can never be truthy;
- `concatControlValidationErrors()` seeds `result` with `this.form.errors` before merging the per-control errors — always seeding `null`.

Both are dead after this change. Dropping them makes it obvious that the editor's validity is now purely control-level.

packages/components/time-range/time-range-editor.ts
```diff
@@ -251,21 +255,56 @@ export class KbqTimeRangeEditor<T> implements ControlValueAccessor, Validator, O
...
+                    fromTime: this.form.controls.fromTime.value,
+                    fromDate: this.form.controls.fromDate.value,
+                    toDate: this.form.controls.toDate.value,
+                    toTime: this.form.controls.toTime.value
+                })
+            )
         };
     }
 
+    /**
```
lskramarov, line 441: **"Is this range reversed?" is answered twice, in two different representations.**

`checkRangeReversed` asks it over the form values (`combineDateAndTime` + `compareDateTime`); `orderRange` asks it again over ISO strings that `calculateTimeRange` built from those same form values one call earlier. So every `valueChanges` — every keystroke in a timepicker — serialises with `toIso8601` and immediately deserialises back just to run a comparison that was available for free a moment before.

Ordering the four form values *before* handing them to `calculateTimeRange` answers it once, reuses the existing `rangeValidator` predicate, and removes `orderRange` entirely:

```ts
const ordered = this.orderedRangeValue(); // fromTime/fromDate/toTime/toDate, swapped when reversed
return { type, ...this.timeRangeService.calculateTimeRange(type, ordered) };
```

The two representations are also a divergence risk: `checkRangeReversed` reads `form.value` and so returns `null` while the range controls are disabled, `orderRange` has no equivalent notion.

packages/components/time-range/time-range.service.ts
```diff
@@ -85,9 +85,15 @@ export class KbqTimeRangeService<T> {
...
     }
 
-    getDefaultRangeValue(): Required<KbqRangeValue<T>> {
-        const from = this.dateAdapter!.addCalendarUnits(this.dateAdapter!.today(), { days: -1 });
-        const to = this.dateAdapter!.today();
+    /**
+     * "Yesterday to today", pulled inside `minDate`/`maxDate` when they are given - otherwise the editor
+     * opens on a range its own datepickers reject.
+     */
+    getDefaultRangeValue(minDate?: T | null, maxDate?: T | null): Required<KbqRangeValue<T>> {
```
lskramarov: **The editor's own default still ignores its own bounds.**

`KbqTimeRangeEditor` owns `minDate`/`maxDate` and the datepickers that reject out-of-bounds values, yet its `rangeValue` input default is still a bare `this.timeRangeService.getDefaultRangeValue()` (time-range-editor.ts:102). `KbqTimeRangeEditor` is exported from `KbqTimeRangeModule`, so using it directly reproduces exactly the bug this overload fixes for `KbqTimeRange`.

An input initialiser cannot read another input, so this needs a `computed()` in the editor — or the clamp moves into the editor and `KbqTimeRange` stops pre-clamping. The two service-internal fallbacks that still call `getDefaultRangeValue()` bare are in the same position: `getTimeRangeDefaultValue` (line 114) and `calculateTimeRange` (line 125).

packages/components/time-range/time-range.spec.ts
```diff
@@ -268,6 +289,245 @@ describe('KbqTimeRange', () => {
...
+        }));
+
+        it('should keep the values while focus moves into the datepicker overlay', fakeAsync(() => {
+            const fixture = setup(TestComponentWithRange);
+            const { dateAdapter, form, rangeElement } = setupReversedRange(fixture);
+            // The calendar is rendered in a CDK overlay, outside of the range block.
+            const overlayPane = document.createElement('div');
+            const calendarCell = document.createElement('td');
+
+            overlayPane.classList.add('kbq-datepicker__popup');
```
lskramarov: **The overlay branch is tested against a hand-made element, so the selector is not pinned to the real datepicker.**

This test builds its own `div.kbq-datepicker__popup` instead of opening the calendar, so nothing connects `'.kbq-datepicker__popup'` in `onRangeFocusOut` to `panelClass: 'kbq-datepicker__popup'` in `datepicker.component.ts:454`. Rename that `panelClass` and this test stays green while the feature breaks — picking a date would swap the range under the user mid-interaction.

Clicking the real `kbq-datepicker-toggle-icon` and asserting against the actual `.cdk-overlay-pane` (the way `datepicker.spec.ts:223` locates it) would tie the two together.

packages/components/time-range/time-range.spec.ts
```diff
@@ -268,6 +289,245 @@ describe('KbqTimeRange', () => {
...
+            fixture.detectChanges();
+
+            const applied = dateAdapter.deserialize(componentInstance.control.value.endDateTime!);
+
+            expect([dateAdapter.getDate(applied), dateAdapter.getHours(applied)]).toEqual([endDay, endHours]);
+        }));
+    });
+
+    describe('Min and max dates', () => {
+        it('should open the range editor on a default range the datepickers accept', fakeAsync(() => {
```
lskramarov, line 631: **This test cannot fail for either way the clamped default is wrong.**

It asserts only `minDate <= value <= maxDate`, which holds for both defects in the clamped default:

- `from === to === maxDate` — the assertion passes just as happily on a zero-length range;
- the millisecond drift, because `createDate(2015, 0, 1)` has `millisecond === 0`, so `omitMilliseconds` is a no-op for these bounds.

`expect(dateAdapter.compareDateTime(form.value.fromDate, form.value.toDate)).toBeLessThan(0)` plus a second case with bounds that carry a time (e.g. `dateAdapter.today()` as `minDate`) would cover both.

packages/components/time-range/time-range.service.ts
NikGurev: Поправил: миллисекунды снимаются до клампа.

---

Milliseconds are now dropped from `today()` before clamping, so a clamped border lands exactly on its bound instead of a fraction below it.

packages/components/time-range/time-range-editor.ts
NikGurev: Поправил: `orderRange` убран целиком.

---

Ordering is decided once, on the form values, before `calculateTimeRange` is called — so nothing rebuilds the object and no key is materialised. The `toIso8601` → `deserialize` round trip is gone with it.

packages/components/time-range/time-range.service.ts
NikGurev: Поправил по твоему предложению.

---

`from` is derived from the already-clamped `to`, so the one-day window survives bounds that lie entirely in the past. One case is still open: with a `minDate` in the future both ends collapse onto it, since the window cannot be widened downwards. Flagged for the author.

packages/components/time-range/time-range-editor.ts
NikGurev: Убрал, ветки `form?.invalid` больше нет.

---

The matcher now reads only the control, and painting is gated on `touched` instead. One field's error no longer smears onto the other, which the spec needs — it asks for the field at fault to be painted, not the pair.

packages/components/time-range/time-range-editor.ts
NikGurev: Убрал оба чтения.

---

`validate()` now returns the concatenated control errors, and `concatControlValidationErrors()` no longer seeds from `this.form.errors`. Validity is purely control-level, as you said.

packages/components/time-range/time-range-editor.ts
NikGurev: Поправил: проверка осталась одна.

---

`orderRange` is gone; reversal is asked once, over the form values, and the result is reused for both the visible swap and the emitted value.

packages/components/time-range/time-range.service.ts
NikGurev: Поправил: границы переехали в сервис.

---

`bindBounds` takes the owner's own input signals rather than a copy of their values, so both service-internal fallbacks and the editor's `rangeValue` default read them live. Note that the premise does not quite hold: `KbqTimeRangeService` is not exported from the entry point, so a standalone `KbqTimeRangeEditor` cannot resolve it at all today.

packages/components/time-range/time-range.spec.ts
NikGurev: Усилил тест.

---

It now asserts a non-zero-length range and adds a case with bounds carrying a time. Both new assertions were checked against the old implementation and fail there.

packages/components/time-range/time-range.ts
NikGurev: Оставляем как есть — так требует спека.

---

The spec is explicit that a value written by the parent, or restored from saved state, is not changed automatically: the error is shown when the group is left or after "apply". So a reversed range written through the control is surfaced as-is, and the swap happens when the user applies. Confirmed with the author.

packages/components/time-range/time-range-editor.ts
NikGurev: Решили иначе: на `null` своп нужен.

---

A click on a blank spot inside the editor also reports a `null` `relatedTarget`, and there the swap is wanted. So bailing on `null` outright is not an option. Instead the footer is recognised by the `mousedown` gesture it starts — which works in Safari, where a click on a button moves no focus — and a `null` `relatedTarget` is only ignored when `document.hasFocus()` is false, i.e. the whole window lost focus. Your observation is what forced the gesture channel, thanks.

packages/components/time-range/time-range.spec.ts
NikGurev: Согласен, пока не сделал — в списке на завтра.

---

Still pinned to a hand-built element. Will be reworked to open the real calendar and assert against the actual overlay pane.

NikGurev:
Scope grew past the original swap fix: the editor now follows the behaviour spec end to end — a border is its date and time together, checked when focus leaves the pair, bounds are exact instants checked on the combined value, and nothing is reordered while a border is wrong. "Apply" is no longer disabled; it reports and focuses the first field at fault. Two further bugs turned up on the way: a nullable trigger was being filled by `handleAvailableTypesChange`, and an abandoned edit survived "cancel" to prefill the next open.

Still open: the calendar-overlay test, and per-border validation messages (pending wording and locales).

rmnturov (approved):

artembelik (approved):

lskramarov (approved):

PR #2033, opened 2026-09-11
fix(toggle,radio)!: keep the same height with and without a label (#DS-4293)

artembelik:
## Summary

A checkable control should occupy one line of text whether or not it has a label, so that `kbq-toggle`, `kbq-checkbox` and `kbq-radio-button` take the same box. Only the checkbox did — 20px in normal, 24px in big. Measured heights of `<div><kbq-... /></div>` before this change:

| | normal, no label | normal, with label | big, no label | big, with label |
| --- | --- | --- | --- | --- |
| checkbox | 20 ✓ | 20 ✓ | 24 ✓ | 24 ✓ |
| radio | 20 ⚠️ | 20 ✓ | **20** ✗ | 24 ✓ |
| toggle | **21** ✗ | **23** ✗ | **21** ✗ | **25** ✗ |

`kbq-toggle` kept the switch in flow, so it was 16px without a label and as tall as the label's line height with one; its baseline came from the bottom edge of the bar container rather than from the text, which added the parent's descender on top of that. `kbq-radio-button` collapsed to zero height without a label — the circle is out of flow and nothing was left in it — and only looked right in normal mode because the wrapper kept the line height inherited from the page (the ⚠️ above).

Both now follow what `kbq-checkbox` already does: an inline-level layout inside an inline-block host, aligned to the top, so the host's own line box owns the height and the baseline.

## List of notable changes:

- **fixed** `kbq-toggle`: `.kbq-toggle-layout` is `inline-flex` with `vertical-align: top`, and `.kbq-toggle-bar-container` gets a height equal to the text line height instead of the `kbq-css-half-difference` padding — the 16px switch is centred in it with a label and without.
- **fixed** `kbq-radio-button`: the host is `inline-block` rather than `inline-flex`, and `.kbq-radio-label` is aligned `top` rather than `middle`.
- **added** `E2eToggleHeight`, `E2eCheckboxHeight` and `E2eRadioHeight` scenarios plus a test per component asserting 20px / 24px. Each test measures the wrapper **and** the component itself — a component collapsed to zero height would otherwise hide behind the wrapper's inherited line height, which is exactly how the radio defect stayed invisible in normal mode.
- **fixed** the existing toggle `should have correct size` test: it measured the host (legitimately 20px now, not 16px) and, after clicking `big`, re-asserted variables destructured before the click, so the second half checked nothing. It now measures `.kbq-toggle-bar` — 28×16 in both modes — and actually re-measures.
- **updated** three toggle baselines (`01-light`, `01-dark`, `02-light`), regenerated via Docker.

1. **The `!`.** No API changes and nothing to migrate, but standalone controls grow: a bare `<kbq-toggle />` from 16px to 20px, a label-less `<kbq-radio-button />` from 0 to 20px. Consumers laying out around the old sizes will see a few pixels of shift. Happy to drop the `!` if that reads as too strong for a CSS-only change.
2. **The mechanism.** Both fixes lean on the host's line box (inline-block host + inline-level layout + `vertical-align: top`). It is the checkbox's existing approach rather than a new one, but it is subtle enough to be worth a second pair of eyes.
3. **Blast radius.** Inside `kbq-radio-group` and `kbq-form` nothing moved — flex items are blockified, so those renders stayed byte-identical, and the radio, checkbox and forms baselines are untouched. `E2eFormHorizontal` was run separately to confirm.

Verified: toggle + checkbox + radio visual e2e in Docker (13 passed), `E2eFormHorizontal` in Docker (passed, no baseline change), jest for the three components (104 passed), eslint / stylelint / prettier / check-e2e-types clean.

Note for anyone reproducing the Docker run on Apple Silicon: the default 8 workers crash Chromium under amd64 emulation; `PLAYWRIGHT_WORKERS=2` runs clean.


packages/components/toggle/toggle.scss
```diff
@@ -20,24 +19,28 @@ $toggle-thumb-offset: 8px;
     }
 
     .kbq-toggle-layout {
-        display: flex;
+        display: inline-flex;
```
lskramarov, line 22: **`inline-flex` makes `.kbq-toggle__content { flex: 1 }` inert whenever the host is wider than its content.**

`display: flex` was block-level, so the layout filled the host and `flex: 1` pushed the switch to the far right with `labelPosition="left"`. `inline-flex` is shrink-to-fit, so it never gets that space.

Reproduced in Chrome with a 300px host, `row-reverse`, `flex: 1` content:

| | layout width | bar left edge |
| --- | --- | --- |
| `display: flex` (before) | 300px | 272px |
| `display: inline-flex` (after) | 49px | 21px |

The clickable `<label>` shrinks the same way, from 300px to 49px. This is exactly the "label left, switch pinned right" settings row demoed at [`packages/components-dev/toggle/template.html:61-72`](https://github.com/koobiq/angular-components/blob/bc1dc7c280301ed0d066041fd6f11d7bedcff63a/packages/components-dev/toggle/template.html#L61-L72) (`.layout-fill` → `width: 100%`), and `.layout-fill` is a public core utility.

`width: 100%` restores it without costing the new height — verified in the same repro: 300px host → bar back at x=272, auto-width host → unchanged at 49px, host height 20px in both.

```suggestion
        display: inline-flex;
        box-sizing: border-box;
        width: 100%;
```

Worth noting the radio side of this PR kept `width: 100%` on `.kbq-radio-label` ("for maximum clickable area") — the toggle has no equivalent.

packages/components/toggle/toggle.scss
```diff
@@ -20,24 +19,28 @@ $toggle-thumb-offset: 8px;
...
-        }
     }
 
     .kbq-toggle-bar-container {
         display: flex;
         align-items: center;
+
+        // The switch is centred in a box as tall as one line of the label, so that a toggle keeps
+        // the same height whether it has a label or not.
+        height: var(--kbq-typography-text-normal-line-height);
```
lskramarov: **A fixed `height` decouples the row from `--kbq-toggle-size-normal-height`.**

The container used to be auto-height, so it grew with the bar. Now it is pinned to the typography line height while `.kbq-toggle-bar` is still sized by the component token. Set `--kbq-toggle-size-normal-height: 28px` and the bar overflows the 20px container by 4px on each side (`align-items: center`), the host stays 20px, and the switch collides with neighbouring rows.

AGENTS.md is explicit that this is the supported extension point: "A component-specific look is expressed by redefining that component's tokens, not by editing the shared theme mixins."

`min-height` gives an identical 20px box for the default 16px bar and still grows with the token:

```suggestion
        min-height: var(--kbq-typography-text-normal-line-height);
```

(and the same for `--kbq-typography-text-big-line-height` below)

packages/components/checkbox/e2e.playwright-spec.ts
```diff
@@ -49,4 +49,32 @@ test.describe('KbqCheckboxModule', () => {
...
+                .map((testId) => locator.getByTestId(testId))
+                .flatMap((wrapper) => [wrapper, wrapper.locator('kbq-checkbox')]);
+
+        test('should take the same height with and without label', async ({ page }) => {
+            await page.goto('/E2eCheckboxHeight');
+
+            const component = getComponent(page);
+
+            for (const target of getTargets(component)) {
+                await expect(target).toHaveCSS('height', '20px');
```
lskramarov, line 70: **This locks in 20px/24px as the checkbox contract, but the number only holds because two `var()` references are broken.**

[`checkbox.scss:37,39`](https://github.com/koobiq/angular-components/blob/bc1dc7c280301ed0d066041fd6f11d7bedcff63a/packages/components/checkbox/checkbox.scss#L37-L39) reads:

```scss
padding-top: var(--kbq-checkbox-size-normal-padding-top);
padding-bottom: var(--kbq-checkbox-size-normal-padding-bottom);
```

Neither token is defined anywhere — not in `checkbox-tokens.scss` (which has `-top`, `-horizontal-content-padding`, `-vertical-content-padding`, but no `-padding-top`/`-padding-bottom`) and not in `@koobiq/design-tokens`. Both are invalid at computed-value time and fall back to `0`, which is why `.kbq-checkbox__layout` happens to be exactly one line tall.

Not introduced here, but this PR is what turns it into a guarded contract. When someone defines those tokens as the naming intends, the checkbox grows, this test goes red, and the failure points at the e2e scenario instead of at the dangling `var()`s. Either drop the two declarations or define the tokens before pinning 20px.

packages/components/toggle/e2e.playwright-spec.ts
```diff
@@ -22,15 +22,17 @@ test.describe('KbqToggleModule', () => {
...
-            expect(width).toBe(28);
-            expect(height).toBe(16);
+            await expect(bar).toHaveCSS('width', '28px');
+            await expect(bar).toHaveCSS('height', '16px');
 
             await getBigToggle(component).click();
 
-            expect(width).toBe(28);
-            expect(height).toBe(16);
+            await expect(bar).toHaveCSS('width', '28px');
```
lskramarov, line 36: **The post-click half still asserts nothing.**

`--kbq-toggle-size-big-width/height` are identical to the normal ones (`28px` / `var(--kbq-size-l)`), so these two lines expect exactly what the pre-click lines already established. `toHaveCSS` polls until it matches — and it matches immediately, against the pre-click state. If `[big]` stopped propagating or the click missed the input, the test still passes.

That is the same vacuity as the destructured-variables version the PR fixes, just one step removed. Gating on the class makes the second half mean something:

```suggestion
            await expect(getFirstToggle(component)).toHaveClass(/kbq-toggle_big/);

            await expect(bar).toHaveCSS('width', '28px');
```

packages/components/toggle/toggle.scss
```diff
@@ -120,10 +123,8 @@ $toggle-thumb-offset: 8px;
         width: var(--kbq-toggle-size-big-width);
     }
 
-    .kbq-toggle-layout {
-        &:has(.kbq-toggle-label:not(:empty)) .kbq-toggle-bar-container {
-            padding-top: tokens.kbq-css-half-difference(typography-text-big-line-height, toggle-size-big-height);
-        }
+    .kbq-toggle-bar-container {
+        height: var(--kbq-typography-text-big-line-height);
```
lskramarov: **Two typography levels now size the same box.**

`_toggle-theme.scss` still sets `line-height` on this element:

```scss
.kbq-toggle:not(.kbq-toggle_big) {
    & .kbq-toggle-bar-container {
        @include tokens.kbq-css-font-variable(typography, text-compact, line-height, '');
    }
}
```

So in the normal size the container carries `line-height: text-compact` and `height: text-normal`. The `line-height` is inert today (flex container, single blockified child), but it now reads as the authoritative row height and disagrees with the real one — anyone later simplifying the explicit `height` away in favour of the co-located `line-height` silently gets a compact-sized row.

Since this PR takes ownership of the container's height, dropping the two dead `.kbq-toggle-bar-container` `line-height` rules from `_toggle-theme.scss` would leave one source of truth.

packages/components/radio/e2e.ts
```diff
@@ -111,3 +112,23 @@ export class E2eRadioStateAndStyle {
...
     protected readonly kbqComponentColors = KbqComponentColors;
 }
+
+@Component({
+    selector: 'e2e-radio-height',
+    imports: [KbqRadioButton, KbqRadioGroup, FormsModule],
+    template: `
+        <input type="checkbox" data-testid="e2eBigToggle" [(ngModel)]="big" />
+
+        <kbq-radio-group [big]="big()">
```
lskramarov, line 122: **The wrapper `<div>`s mean the shipped nesting is never height-checked.**

The `<div>` is the flex item of `.kbq-radio-group`, so `kbq-radio-button` inside it keeps `display: inline-block`. In real markup the button is the direct child, becomes a flex item, and `inline-block` blockifies to `block` — a different box-generation path, and the one that actually ships.

Both paths land on 20px today, so this is a coverage gap rather than a bug: a future rule that only affects the blockified host would sail past this guard. One un-wrapped button measured alongside the wrapped ones closes it.

(The `[big]` plumbing forces the group here — `kbq-radio-button` only reads `radioGroup?.big()` — so a truly standalone big radio isn't expressible. No objection to the group itself.)

packages/components/radio/radio.scss
```diff
@@ -17,13 +17,15 @@
 
 // Top-level host container.
 .kbq-radio-button {
-    display: inline-flex;
+    // Inline-block rather than inline-flex: the host's own line box keeps the button one line tall
+    // even when the label is empty, and puts the baseline where the surrounding text expects it.
+    display: inline-block;
```
lskramarov, line 22: **Breaking change with nothing in the upgrade path.**

The title carries `!`, and the PR body is clear about what moves: a label-less `<kbq-radio-button>` goes 0 → 20px, a bare `<kbq-toggle>` 16 → 20px. But there is no entry under `packages/schematics/src/migrations/`, and neither `radio.{en,ru}.md` nor `toggle.{en,ru}.md` mentions the new box.

AGENTS.md (Schematics): "A breaking change to a component ships with a migration there, or at least a warning that points at the manual steps."

A codemod can't fix CSS-driven layout shift, which is precisely why the manual-steps note carries the weight here — consumers get a few pixels of shift around every standalone control with nothing in the release path telling them where to look.

On the `!` question from the description: keeping it looks right to me. 0 → 20px for a label-less radio is not a rounding difference.

packages/components/radio/e2e.playwright-spec.ts
```diff
@@ -16,4 +16,32 @@ test.describe('KbqRadioModule', () => {
...
         });
     });
+
+    test.describe('E2eRadioHeight', () => {
+        const getComponent = (page: Page): Locator => page.getByTestId('e2eRadioHeight');
+        const getBigToggle = (locator: Locator): Locator => locator.getByTestId('e2eBigToggle');
+
+        // The component is measured alongside its wrapper: collapsed to zero height it would still
+        // leave the wrapper at the line height inherited from the page and hide the defect.
+        const getTargets = (locator: Locator): Locator[] =>
```
lskramarov: **`getTargets`, `getBigToggle`, the test body and this comment are byte-identical in three spec files.**

The toggle, checkbox and radio specs each carry the same four pieces, differing only in two test-id strings and one element selector. `packages/e2e/utils` already exists for exactly this — all three files import `e2eEnableDarkTheme` from it.

Something like `e2eExpectOneLineHeight(page, { route, hostTestId, selector })` exported from `packages/e2e/utils/index.ts` would leave one call per spec, and the contract (20px normal / 24px big, wrapper **and** component) would be stated once instead of drifting in three places.

packages/components/toggle/toggle.scss
```diff
@@ -20,24 +19,28 @@ $toggle-thumb-offset: 8px;
...
 
     .kbq-toggle-layout {
-        display: flex;
+        display: inline-flex;
         flex-direction: row;
         align-items: flex-start;
 
+        // Keeps the host's own line box in charge of the baseline, so that the box a toggle
+        // occupies in text is exactly one line tall.
+        vertical-align: top;
```
lskramarov, line 30: **Third hand-rolled copy of the same line-box mechanism.**

"Inline-block host + inline-level layout + `vertical-align: top`" is now written out three times, each with its own inner geometry: checkbox via padding tokens on `.kbq-checkbox__layout`, toggle via a fixed `.kbq-toggle-bar-container` height, radio via an absolutely positioned circle offset by `kbq-css-half-difference`.

Following the checkbox rather than inventing something new is the right call — the observation is that the invariant "a checkable control occupies exactly one line of text" still has no single home. A `kbq-checkable-line-box()` mixin under `packages/components/core/styles` would state it once, let each component supply only its own inner geometry, and stop the fourth checkable control from repeating this a fourth time.

Fine as a follow-up; raising it because this PR is the moment the pattern became a pattern.

lskramarov (commented):
## Code review — xhigh

Ten findings, most severe first. The mechanism itself (inline-block host + inline-level layout + `vertical-align: top`) reads correct to me, and the reasoning in the description matches what I measured; the substantive issue is a side effect of `flex` → `inline-flex` that the description does not cover.

| # | File | Finding |
| --- | --- | --- |
| 1 | `toggle.scss:22` | `inline-flex` makes `.kbq-toggle__content { flex: 1 }` inert — switch no longer pinned right, label area shrinks |
| 2 | `toggle.scss:43` | Fixed `height` decouples the row from `--kbq-toggle-size-*-height` |
| 3 | `checkbox/e2e.playwright-spec.ts:70` | New test locks in a height that only holds because two `var()`s are undefined |
| 4 | `toggle/e2e.playwright-spec.ts:34` | Big half of `should have correct size` still asserts nothing |
| 5 | `toggle.scss:127` | `_toggle-theme.scss` still sizes the same box from a different typography level |
| 6 | `radio/e2e.ts:122` | Wrapper `<div>`s mean the shipped group nesting is never height-checked |
| 7 | `radio.scss:22` | Breaking change with no migration entry and no docs note |
| 8 | `radio/e2e.playwright-spec.ts:26` | Helper + test body copied byte-identically into three specs |
| 9 | `toggle.scss:28` | Third hand-rolled copy of the line-box mechanism; no shared mixin |
| 10 | `radio/e2e.playwright-spec.ts:31` | Test name covers half the body |

Finding 1 is the one I would not merge without: reproduced in Chrome, the switch moves from x=272 to x=21 in a 300px host with `labelPosition="left"`, and the clickable `<label>` drops from 300px to 49px. `width: 100%` on `.kbq-toggle-layout` restores it with no cost to the new height.

packages/components/radio/e2e.playwright-spec.ts
```diff
@@ -16,4 +16,32 @@ test.describe('KbqRadioModule', () => {
...
+            ['e2eRadioWithoutLabel', 'e2eRadioWithLabel']
+                .map((testId) => locator.getByTestId(testId))
+                .flatMap((wrapper) => [wrapper, wrapper.locator('kbq-radio-button')]);
+
+        test('should take the same height with and without label', async ({ page }) => {
```
lskramarov: **The name covers half the body.**

The second half of this test asserts the normal-vs-big contract (20px → 24px after clicking the size toggle), which the name never mentions. A failure there reports "should take the same height with and without label" and sends the reader to the label logic.

`should take one line of text in both sizes, with and without a label` — or splitting the big half into its own test — makes the failure self-describing. Same in the toggle and checkbox specs.

NikGurev (approved):

lskramarov (approved):

PR #2034, opened 2026-09-11
feat(components): use KbqScrollbarViewport in code-block, file-upload and vertical tabs (#DS-5453)

artembelik:
## Summary

Moves the last three components that customized the browser-rendered scrollbar (`kbqNativeScrollbar`) onto the custom one, and fixes what that exposed in `KbqScrollbarViewport` itself. `textarea` is deliberately left out — the viewport cannot be applied to an element that cannot host children; the analysis is written up separately.

Along the way the components that already used the viewport were switched from assigning `scrollTop` to its scroll API.

## List of notable changes:

**Scrollbar fixes** (each has a unit test that fails without it):

- **fixed** `applyGeometry` so it clamps the track size once and derives the end margin from the clamped value. An unclamped `size - 1` put `block-size: -1px` — silently rejected by CSSOM — against an applied `margin-block-end: 1px`, leaving a zero-sized viewport oscillating through its own `ResizeObserver` forever. Measured: 600 observer callbacks and 600 forced layouts per 5 s, ~38 % of a core on an idle page. After the fix: 0 and 0.
- **fixed** bar visibility so it is gated on the computed `overflow`. `hidden`, `clip` and `visible` all rule out user scrolling, so a track no longer paints a draggable thumb for an axis the browser refuses to scroll. This is what made a collapsed `[maxHeight]` code block show a working scrollbar it should not have had.
- **fixed** the reveal timer to schedule through the zone-free scheduler. `flashes` carries its caller's zone, so a flash from inside Angular left `NgZone` unstable for the whole `hideDelay` — blocking `whenStable()` and SSR serialization.

**Components:**

- **added** `kbqScrollbarViewport` to `code-block`, the `file-upload` list and the vertical tab header. `tab-body` deliberately stays on `kbqNativeScrollbar`: the `ResizeObserver` fix above makes it viable, but 100 tracks for inactive tab bodies is a separate cost.
- **fixed** the tab strip to pick its scrollbar mode from the class the stylesheet keys `overflow-y: auto` on, not from the `vertical` input. `KbqVerticalTabsCssStyler` matches on attribute *presence*, so `[vertical]="false"` would otherwise leave the strip scrollable with no scrollbar of any kind.
- **added** a brief scrollbar reveal when each component's content arrives or grows, so whether it scrolls is answered on sight rather than only on hover.
- **updated** `select`, `tree-select`, `autocomplete` and `dropdown` to scroll through the viewport API instead of assigning `scrollTop`.
- **fixed** `code-block`'s `<pre>` to stretch to its content, so the code's right padding is painted where the line ends instead of at the scrollport edge.
- **updated** `file-upload` markup: the list is wrapped in `<kbq-scrollbar>`, which keeps the scrollbar track out of the `role="list"` subtree. `kbq-file-upload__list` stays on `<kbq-list>`; the wrapper carries the new `kbq-file-upload__list-viewport`.

**Tooling:**

- **added** an `Element.prototype.scrollTo` polyfill to `tools/jest/setup.ts` — jsdom does not implement it and the viewport now reaches it through `CdkScrollable`. Replaces two file-local stubs.

1. **The tab strip mode condition.** `verticallyScrolled` reads `closest('.kbq-tab-group_vertical')` rather than the `vertical` input, because that is exactly what the stylesheet keys the vertical scrollport on. Reading the DOM class is deliberate — the two conditions have to agree or the strip scrolls with no scrollbar — but it is worth a second opinion on whether there is a cleaner way to express "the same thing the CSS decided".

2. **`native` mode loses Koobiq theming here.** Removing `kbqNativeScrollbar` from these components means an app that sets `kbqScrollbarOptionsProvider({ mode: 'native' })` gets the raw OS scrollbar in vertical tab headers, code blocks and file-upload lists. Every other component that already uses the viewport behaves this way, but `tab-body` still carries `kbqNativeScrollbar`, so a tab group is visibly inconsistent in that mode.

3. **Overlay scrollbars change hit areas.** The custom track overlays content instead of reserving space. In `file-upload` it sits over the right ~12 px of the delete button. That is inherent to overlay scrollbars rather than to this change, but it is a design call worth confirming.

4. **`e2e` fixtures use `kbqScrollbarOptionsProvider({ hideDelay: 5000 })`.** The reveal lasts `hideDelay` and nothing brings it back, so the default second would make those tests a race against page load. Every new negative assertion also waits for the sibling overflowing viewport to paint first — the track reports no bars for its first frame plus one throttle window regardless of content, so asserting an empty track too early passes on the startup window rather than on the behaviour.

Screenshots are unchanged across the whole suite: headless Chromium uses overlay scrollbars, and a `hover`-mode track is transparent at rest.


packages/components/file-upload/multiple-file-upload.component.ts
```diff
@@ -280,6 +286,32 @@ export class KbqMultipleFileUploadComponent
...
+        //
+        // Gated on growth, not on any write: `list` is re-set wholesale by the `files` setter and by
+        // `writeValue`, so an unguarded effect would re-announce scrollability while the user is deleting
+        // rows, and would fire on every change-detection pass for a consumer binding `[files]` to an
+        // expression that returns a fresh array.
+        effect(() => {
+            const count = this.fileList.list().length;
+            const viewport = this.listViewport();
+
+            if (!viewport) {
```
lskramarov, line 302: **correctness — `announcedFileCount` goes stale when the list unmounts**

The early return leaves `announcedFileCount` at its old value, so a list that is emptied and then refilled with *fewer* files is never announced again.

User uploads 5 files (`announcedFileCount = 5`), then deletes all 5. `files.length` hits 0, the `@else` branch is torn down, `listViewport()` becomes `undefined`, and this branch returns without touching the count. The user then drops 3 new files: the effect runs with `count = 3`, `3 > 5` is false, `flashScrollIndicators()` is never called — even though this is exactly the "the list arrives" case the effect exists for.

The comment above says the count is left alone "so its next appearance still counts as growth", but that only holds when the new list is larger than the old one. Nothing has been announced for the *new* list, so the counter should go back to zero with it:

```suggestion
            if (!viewport) {
                // The list is gone; nothing has been announced for the next one that appears.
                this.announcedFileCount = 0;

                return;
            }
```

packages/components/tabs/tab-header.html
```diff
@@ -11,8 +11,9 @@
 
 <div
     #tabListContainer
-    kbqNativeScrollbar
+    kbqScrollbarViewport
     class="kbq-tab-header__container kbq-tab-header__scroll-container"
```
lskramarov, line 15: **correctness — the tab strip is now a global `ScrollDispatcher` scroll source**

`kbqScrollbarViewport` carries `CdkScrollable` as a host directive, which registers the strip with the app-wide `ScrollDispatcher`. `kbqNativeScrollbar` did not.

`ScrollDispatcher.scrolled()` is global — it merges every registered scrollable, it is not scoped to ancestors of the scroller. `KbqPaginatedTabHeader` drag-scrolls the strip with inertia, emitting a native `scroll` per frame, so dragging a tab strip now drives the scroll strategy of every overlay open anywhere on the page (reposition or close, depending on the strategy). The same applies to the new `<kbq-scrollbar>` wrapper around the file-upload list.

That may well be what you want — an overlay anchored to a tab *should* follow it — but it is a behaviour change well outside the tab header, and it lands on a code path that fires at frame rate. Worth a deliberate check against the default overlay scroll strategies before merge.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -253,6 +258,28 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
...
+
+    /**
+     * Whether the strip is laid out as a vertical scrollport — read off the class rather than the
+     * `vertical` input, because that class is what the stylesheet keys `overflow-y: auto` on and
+     * `KbqVerticalTabsCssStyler` matches on attribute *presence*: `[vertical]="false"` still gets it.
+     * Deriving the mode from the input alone would let the two disagree and leave such a strip
+     * scrollable with no scrollbar of any kind.
+     */
+    private get verticallyScrolled(): boolean {
+        return !!this.elementRef.nativeElement.closest('.kbq-tab-group_vertical');
```
lskramarov: **efficiency — `closest()` runs on every change-detection pass**

`verticallyScrolled` is reached from `[kbqScrollbarMode]="scrollbarMode"`, and `KbqTabHeader` is declared with `changeDetection: ChangeDetectionStrategy.Default` — so this DOM walk re-runs on every app-wide tick, for every tab header on the page.

The answer can never change: `KbqVerticalTabsCssStyler` applies `.kbq-tab-group_vertical` as a *static* host class, fixed at element-creation time. The host element is already appended to its parent by the time the constructor runs, so this can be read once and cached in a field, which also removes the per-tick cost from the `ngAfterContentChecked` call site.

packages/components/scrollbar/scrollbar.ts
```diff
@@ -697,10 +709,14 @@ class KbqScrollbarTrack {
 
     private get scrollbars(): ScrollbarVisibility {
         const { clientHeight, scrollHeight, clientWidth, scrollWidth } = this.viewportElement;
+        // An overflowing axis is not necessarily a scrollable one: `overflow: hidden` still scrolls
+        // programmatically, so the size ratio alone would paint a bar — and hand over a draggable
+        // thumb — for an axis the browser itself refuses to give a scrollbar to.
+        const { overflowX, overflowY } = this.window.getComputedStyle(this.viewportElement);
```
lskramarov, line 718: **efficiency — `getComputedStyle` added to the 300 ms poll**

`scrollbars` is evaluated by `visibility` on a self-rescheduling `requestAnimationFrame` throttled to `TRACK_THROTTLE_TIME` (300 ms), for the lifetime of every track on the page. Adding a `getComputedStyle` read there means a style flush ~3.3×/s per track — on a docs page with 14 code blocks, ~47/s, forever, whether or not anything changed. That sits awkwardly next to this PR's own framing about idle observer/layout cost.

`getViewportMetrics()` already reads `getComputedStyle` but only on the `SharedResizeObserver` tick. Caching the two overflow values and refreshing them from a cheaper trigger (the resize subscription plus a `MutationObserver` on `style`/`class`) would keep the fix without putting a style resolution in the hot loop.

packages/components/code-block/e2e.playwright-spec.ts
```diff
@@ -36,9 +36,45 @@ test.describe('KbqCodeBlockModule', () => {
...
+        test('reveals nothing for a code block that does not scroll', async ({ page }) => {
+            // Waited on first: the track reports no bars for its first frame plus one throttle window
+            // whatever the content is, so asserting the empty track before any tick has run would pass
+            // on the pre-computation window rather than on the behaviour. The overflowing block shares
+            // this page's frame loop, so its bars appearing prove a tick has been through.
+            await expect(
+                getTrack(page, 'e2eCodeBlockFlashOverflowing').locator('.kbq-scrollbar-track__bar')
+            ).not.toHaveCount(0);
+
+            await e2eExpectNoScrollbarAfterFlash(
```
lskramarov, line 67: **test robustness — the negative flash tests race a 5 s reveal window**

`e2eExpectNoScrollbarAfterFlash` asserts the track carries `kbq-scrollbar-track_revealed`. That class exists only for `hideDelay` after the flash and nothing re-arms it, so this test must reach the assertion within 5000 ms of the reveal, which starts at page load.

The wait above (for the sibling's bars) plus a cold page load plus highlight.js loading can eat that budget on a loaded CI agent running several Playwright workers. `retries` is 0 everywhere per AGENTS.md, so it fails the run outright rather than retrying. The same shape is in the tabs and file-upload specs.

Asserting *absence* — no bars, no thumb, no overflow — does not need the reveal to still be on. If the reveal genuinely has to be observed, gate on something that does not expire (e.g. have the fixture keep flashing, or assert the class on the overflowing sibling once and drop it from the helper).

packages/components/code-block/code-block.ts
```diff
@@ -558,6 +557,40 @@ export class KbqCodeBlock implements AfterViewInit {
...
 
+    /**
+     * Briefly reveals the scrollbar once highlighting settles, so whether the code scrolls is answered
+     * on arrival rather than only after the pointer enters the block. Switching files needs nothing of
+     * its own: `onSelectedTabChange` scrolls the content back to the top, and that scroll reveals the
+     * track by itself.
+     *
+     * A viewport with nothing to scroll paints no track, so this stays silent for code that fits.
+     */
+    private revealScrollbarOnHighlight(): void {
```
lskramarov, line 568: **altitude — reveal-on-arrival is hand-rolled three times for a condition the track already computes**

This PR adds three independent "announce the scrollbar when content arrives" mechanisms: this highlight-settled subscription, a tab-count delta in `KbqPaginatedTabHeader.ngAfterContentChecked`, and a file-count delta effect in `KbqMultipleFileUploadComponent`. Select, tree-select, autocomplete and dropdown already carry their own.

`KbqScrollbarTrack.visibility` already transitions `[false, false] → [x, y]` at exactly the moment an axis becomes user-scrollable. Revealing on that transition inside the track would collapse all of these into one rule and would also cover what none of them handle: a code block whose container narrows into horizontal overflow, a file list pushed past its cap by a re-wrapped row, any viewport whose content grows without a count changing.

As it stands, every component that wants the behaviour has to invent its own growth signal — and each one gets a different edge case wrong (see the `announcedFileCount` bug in the file-upload effect).

packages/components/scrollbar/scrollbar.ts
```diff
@@ -165,6 +165,14 @@ type Orientation = 'horizontal' | 'vertical';
 
 const TRACK_THROTTLE_TIME = 300;
 
+/**
+ * Computed `overflow` values that rule out user scrolling. `visible` belongs here as much as the other
+ * two: it only computes to `auto` when the other axis scrolls, so a box left at the initial value on
+ * both axes is not a scroll container at all. Listed as the exclusions rather than as the scrollable
+ * values so an unreadable computed style leaves the scrollbar alone instead of erasing it.
+ */
+const NON_SCROLLABLE_OVERFLOW: readonly string[] = ['clip', 'hidden', 'visible'];
```
lskramarov, line 177: **conventions — the `always` mode documentation is now inaccurate**

This gate changes what `always` promises, but neither the JSDoc nor the docs pages were updated:

- `KbqScrollbarMode`'s own JSDoc, a few lines up: "`always` — track is always visible while the content overflows"
- `scrollbar.en.md:10`: "`always` | Remains visible while the content overflows" (and the `ru` equivalent)

Both are false now for an axis computing to `hidden`, `clip` or `visible`. AGENTS.md: "Update or remove comments that become outdated because of your changes", and "Every page exists in both languages — update both".

Worth calling out for consumers too: anyone who put `kbqScrollbarViewport` on an `overflow: hidden` element they scroll programmatically silently loses the track, with nothing in the docs explaining why.

packages/components/tabs/e2e.playwright-spec.ts
```diff
@@ -71,12 +71,48 @@ test.describe('KbqTabsModule', () => {
...
+            page
+        }) => {
+            const track = getTrack(page, 'e2eTabsFlashOverflowing');
+
+            await expect(track).toHaveClass(/kbq-scrollbar-track_revealed/);
+            await expect(track.locator('.kbq-scrollbar-track__bar')).not.toHaveCount(0);
+        });
+
+        test('reveals nothing for a strip whose tabs fit', async ({ page }) => {
+            // Waited on first: the track reports no bars for its first frame plus one throttle window
```
lskramarov: **reuse — this preamble is pasted into three specs**

This four-line comment and the assertion under it are byte-identical in `tabs`, `code-block` and `file-upload` playwright specs, immediately before the shared `e2eExpectNoScrollbarAfterFlash` call they exist to set up.

Three copies drift independently — a fix to the wait (for instance the `hideDelay` race flagged on the code-block spec) has to be found and applied three times, and the next component to adopt the pattern makes a fourth. The helper in `packages/e2e/utils/scrollbar.ts` already owns the rest of this assertion; an optional `settledSibling: Locator` parameter would let all three preambles go away.

packages/components/tabs/paginated-tab-header.ts
```diff
@@ -150,6 +153,8 @@ export abstract class KbqPaginatedTabHeader implements AfterContentChecked, Afte
 
     abstract readonly items: QueryList<KbqPaginatedTabHeaderItem>;
     abstract readonly tabListContainer: ElementRef<HTMLElement>;
+    /** The strip's scroll viewport — the same element as {@link tabListContainer}. */
+    protected abstract readonly scrollbarViewport: KbqScrollbarViewport | undefined;
```
lskramarov: **simplification — `| undefined` that can never be undefined**

Both implementations resolve this as `@ViewChild('tabListContainer', { static: true, read: KbqScrollbarViewport })`, and `kbqScrollbarViewport` sits unconditionally on `#tabListContainer` in both `tab-header.html` and `tab-nav-bar.html` — no structural directive, no mode gate (the directive instance exists even in `hidden` mode; only the *track* is skipped). A `static: true` query resolves during view creation, so the value is never absent.

The optionality buys a `?.` that can never fire, and it is now part of the published surface of both subclasses in `tabs.api.md`. Declaring it non-optional would state the real contract and turn a future template change that drops the directive into a compile error instead of a silent no-op.

lskramarov (commented):
Automated review (`/code-review xhigh`) — 12 findings, ranked most severe first. One correctness bug in the file-upload reveal effect, one cross-cutting side effect worth confirming, plus efficiency, test-robustness and cleanup notes. Everything else in the diff (the `applyGeometry` clamp, the zone-free hide timer, the scroll-API migrations in select/tree-select/autocomplete/dropdown, the `viewChild.required(CdkScrollable)` that now resolves through the host directive, the `jest` `scrollTo` polyfill) checked out.

packages/components/code-block/code-block.spec.ts
```diff
@@ -1060,7 +1070,30 @@ describe(KbqCodeBlock.name, () => {
...
 
             expect(scrollSpy).toHaveBeenCalledWith({ top: 100 });
-        }));
+        });
+
+        it('reveals the scrollbar once highlighting settles, so an arriving block advertises its scroll', async () => {
+            let resolveCore!: (value: { default: HLJSApi }) => void;
+            // Installed on the prototype, before the component exists: a spy taken off the instance
+            // afterwards could not have recorded a premature flash, which is half of what this asserts.
+            const flashSpy = jest.spyOn(KbqScrollbarViewport.prototype, 'flashScrollIndicators');
```
lskramarov, line 1079: **test hygiene — the prototype spy is never restored**

`jest.config.js` sets `clearMocks: true` but not `restoreMocks`, so this spy is only `mockClear`ed between tests — the wrapper stays installed on `KbqScrollbarViewport.prototype` for the rest of the file. It calls through, so nothing breaks today, but any later test that spies on the same member or compares the prototype method sees the leftover wrapper.

The `console.warn` spy earlier in this file already uses `try`/`finally` with `mockRestore()`; same treatment here.

packages/components/code-block/code-block.scss
```diff
@@ -113,6 +113,21 @@
         }
     }
 
+    // A block box is as wide as its containing block, never as wide as the content overflowing it, so
```
lskramarov: **conventions — the comment explains CSS, not the constraint**

Eight lines of block-box intrinsic-sizing exposition above a two-declaration rule. AGENTS.md: "Prefer clear naming and structure over comments. Add concise comments only to explain non-obvious rationale or constraints; never restate the code" and "Omit details apparent from names, types, or signatures".

The one fact a maintainer needs is that the code's right padding is painted at the unstretched box edge, and that soft wrap has to be excluded — two lines. The rest is CSS-spec background that will go stale as soon as the rule moves.

Same pattern elsewhere in the diff: the four-line clamp rationale in `applyGeometry`, the five-line `waitForAsync` note at the top of `code-block.spec.ts`, and the three-line provider comment repeated verbatim in all three new e2e fixtures.

packages/components/file-upload/e2e.ts
```diff
@@ -359,3 +360,54 @@ export class E2eFileUploadDropzone {
...
+        kbqScrollbarOptionsProvider({ hideDelay: 5000 })
+    ],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        'data-testid': 'e2eFileUploadScrollbarFlash'
+    }
+})
+export class E2eFileUploadScrollbarFlash {
+    // Against the 120px cap the fixture sets: this many rows overflow it, a single one does not.
+    protected readonly overflowingFiles: KbqFileItem[] = Array.from({ length: 12 }, (_, index) => ({
```
lskramarov, line 407: **simplification — redundant `satisfies` on already-annotated fields**

Both fields carry an explicit `: KbqFileItem[]` annotation *and* a trailing `satisfies KbqFileItem[]`, plus an inner `['test'] satisfies BlobPart[]` on a literal that is already in a `BlobPart[]` parameter position.

`satisfies` exists to check an expression that has no declared type; here the annotation already does that check, so all three add noise a reader has to parse — and invite the same copy-paste in the next fixture. Dropping them gives identical type safety.

packages/components/scrollbar/scrollbar.ts
artembelik: Строку про `always` не менял - в 3418a49e3 добавил оговорку про `overflow` в JSDoc и под таблицами режимов на обоих языках, она касается и `hover`.

packages/components/tabs/tab-header.html
artembelik: Проверил тестом: скролл полосы вкладок и списка file-upload закрывает `KbqTooltip` и select/dropdown внутри `kbq-tab`, остальные оверлеи только перепозиционируются. На `main` так же уже ведут себя 12 других вьюпортов, поэтому оставляю как есть.

packages/components/scrollbar/scrollbar.ts
artembelik: Вынесено в отдельную задачу про производительность скролбара: там опрос заменяется событиями, и `getComputedStyle` из цикла пропадает вместе с ним.

packages/components/code-block/code-block.ts
artembelik: Оставляю как есть: правило в треке показывало бы полосу в любом вьюпорте сразу при переполнении, а modal, sidepanel, popover и select намеренно флешат только после анимации открытия.

packages/components/code-block/code-block.scss
artembelik: Здесь сократил до ограничения, а комментарий над клампом в `applyGeometry` оставил - он единственный объясняет петлю `ResizeObserver`.

packages/components/tabs/_tabs-common.scss
```diff
@@ -192,24 +192,17 @@
 
         overflow-x: auto;
         overflow-y: hidden;
-        scrollbar-width: none;
```
NikGurev, line 195: Image

то что scrollbar track смещен на 1 пиксель - так задумано или это баг?

Я полагаю что табы со скроллом должны отображаться аналогично опциям в селекте. Сейчас справа есть небольшой отступ

Image

Image

@rmnturov что скажешь?

NikGurev (approved):

packages/components/tabs/_tabs-common.scss
artembelik: > то что scrollbar track смещен на 1 пиксель - так задумано или это баг?

так задумано, в некоторых браузерах одинаковый размер может привести к отображению скрола

lskramarov (approved):

packages/components/tabs/_tabs-common.scss
artembelik: завел отдельную задачу DS-5579

PR #2035, opened 2026-09-11
feat(splitter): rewrite the component on top of KbqResizer (#DS-292)

artembelik:
## Summary

Rewrites `KbqSplitter` from scratch on top of the `KbqResizer` primitive, and moves the previous implementation to `@koobiq/components/splitter/deprecated` unchanged, the way the scrollbar deprecation was done. An `ng update` migration rewrites consumer imports to the deprecated path, so nothing breaks on upgrade.

The new component lays its panels out as grid tracks and owns the layout itself: `KbqResizer` gains a `disableSizeUpdate` input so it reports the size a drag asks for instead of writing it onto the element.

## List of notable changes:

- **added** `KbqSplitter` / `KbqSplitterPanel` at `@koobiq/components/splitter` — standalone components with `size` / `minSize` / `maxSize` (pixels or a share of the splitter), `snapSizes` + `snapTolerance`, `collapsible` + `collapsedSize` + two-way `collapsed`, `orientation`, `appearance`, `disabled` and a two-way `layout`
- **added** `KBQ_SPLITTER_OPTIONS` and `kbqSplitterOptionsProvider`, which set the default appearance and snap tolerance for an application or an injector
- **added** `disableSizeUpdate` to `KbqResizer`, so a host that owns the layout can apply sizes itself
- **added** an `ng update` migration `splitter-deprecated-path` that rewrites `@koobiq/components/splitter` imports to `@koobiq/components/splitter/deprecated`
- **added** `a11y.resizePanels` to the locale configuration, used as the separator's default accessible name
- **updated** the deprecated splitter's root class to `kbq-splitter-component`, so the old and the new stylesheet no longer overwrite each other in an application that renders both during a migration
- **updated** the docs and examples: one example per setting (appearance, orientation, disabled, constraints, snapping, collapsing, nesting) instead of a single combined overview

**Why percentages and not `fr`.** `gridTemplate` writes the track sizes as percentages of the measured content box. Flex factors are normalized by their sum, so a layout whose `minSize` / `maxSize` cannot be satisfied inside the container would be silently rescaled to fill it — three panels at `minSize: 125` in a 340px splitter render 113px each, breaking every declared minimum while `aria-valuenow` and the drag arithmetic still report 125. Percentages under- or overflow instead, which is visible. Covered by `should keep a track below the container when the minimums cannot fit` and the `constraints that cannot be satisfied` e2e test.

**The collapse gap.** A collapsible panel's sizes are `collapsedSize` and everything from `minSize` up, with nothing in between. A drag stops at `minSize`, waits while the pointer crosses the gap, and flips at the midpoint — the same point in both directions, so the panel collapses and reopens in the same place. That is what `gapPins` / `freePanels` carry.

**Known follow-ups, deliberately not in this PR:**

- a drag ends only through the separator's `lostpointercapture`; destroying that separator mid-drag leaves the splitter in its dragging state. The fix is a `resizeEnd` output on `KbqResizer`, which touches two other consumers
- changing the panel count at runtime leaves a two-way `layout` stale and unemitted; fixing it properly means keying the layout by panel rather than by index
- before the first measurement the splitter renders equal tracks even when every panel declares a pixel size, so SSR ships that layout and it shifts on hydration

**Base.** The branch is cut from `a73f1356a` and is behind `main`; it still needs a rebase, and the conflicts will be in the generated and append-only files (`example-module.ts`, `e2e/routes.ts`, `schematics/src/{collection,migrations}.json`, the locale files and `core.api.md`).


rmnturov:
Селекты-переключалки в примерах замени, плиз, на Button Toggle Group. Опций мало, в селекте они скрыты под выпадашку, в группе кнопок будут сразу видны
image

rmnturov:
Увеличь, плиз, высоту свернутого состояния панели, чтоюы текст в две стркои помещался и оставались вертикальные внутренние поля между текстом и панелью
image

rmnturov:
Линии-разделители доводить бы до конца. Сейчас отступы нежелательные остаются
image

rmnturov (approved):

artembelik:
> Селекты-переключалки в примерах замени, плиз, на Button Toggle Group. Опций мало, в селекте они скрыты под выпадашку, в группе кнопок будут сразу видны

заменил

artembelik:
> Увеличь, плиз, высоту свернутого состояния панели, чтоюы текст в две стркои помещался и оставались вертикальные внутренние поля между текстом и панелью

это баг, поправил

вторая строка должна была схлопываться при драге, событие схлопывания срабатывало после отпускания сплиттера, теперь проверка происходит во время

artembelik:
> Линии-разделители доводить бы до конца. Сейчас отступы нежелательные остаются

поправил,  добавил скриншот тест

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+                    ? (snapshot.layout?.[index] ?? (snapshot.sizes[index] / snapshot.total) * 100)
+                    : share
+            )
+        );
+
+        return true;
+    }
+
+    private applySizes(sizes: number[]): void {
+        const total = this.containerSize();
```
lskramarov, line 1188: **Collapse does not restore the size it remembers (drag → host/keyboard expand).**

`settleCollapse` rewrites only the collapsing panel's `layout` entry and leaves every other entry at the value the drag left it at, so the stored layout no longer sums to 100. `fitSplitterSizes` then spreads the surplus **equally** across every open panel instead of taking it back from the panel that absorbed it.

Walking through the shipped `splitter-collapsible` example (600px splitter, panel 0 `size=150 minSize=125 maxSize=200 collapsible`, panel 1 flexible):

1. Drag panel 0 shut. The last `applySizes` leaves `layout ≈ [0, 100]`; this update rewrites index 0 to the remembered `25` → `layout = [25, 100]` (sum 125). Rendering is still correct, because `fitSplitterSizes` clamps index 0 to `collapsedSize` first and the sum lands back on `total`.
2. Flip the `Sidebar collapsed` toggle back off. Now `min = [125, 0]`, `max = [200, ∞]`:
   - clamp → `[150, 600]`, `diff = −150`, both panels open, `step = −75` → `[125, 525]`
   - second pass: `diff = −50`, only panel 1 open → `[125, 475]`

The sidebar reopens at its **125px minimum**, not the 150px it was at — the opposite of what the comment on this line promises.

Nothing catches it: the Jest `should toggle a collapsible panel on Enter and restore the size it had` and the e2e `should collapse and expand from the keyboard` both run with `layout === null`, where `defaultSizes()` re-derives from the declared sizes and masks the defect. It only appears once a drag has made the layout owned.

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+
```
lskramarov, line 1189: **`(layout ?? [])` publishes `[]` when the layout is still derived.**

Reachable without any drag movement:

1. A splitter with an initially collapsed collapsible panel (`collapsible [collapsed]="true"`, as in `E2eSplitterNestedCollapsed`) has `layout === null`.
2. Press and release the separator with **no** `pointermove` in between — a trackpad tap, a fast click, a programmatic `.click()`.
3. `handleResizeStart` skips `applySizes` because the layout is null; `handleResizeTo` never runs; `handleResizeEnd` → `settleCollapse` finds `pins[index]` already `true` (seeded from `panel.collapsed()`), so `collapsing` is non-empty and this `update` maps over `[]`.

Two consequences:

- the host's `[(layout)]` receives an **empty array** for a splitter that has panels — not a value the documented "sizes of the panels as percentages of the splitter, in panel order" contract allows;
- the next double-click is swallowed by `if (this.layout() !== null) return this.layout.set(null)` in `handleSeparatorDblClick`, so the documented two-step ("resets the sizes … a second double-click sets the panel to its minimum") silently loses its first step.

Guarding on `layout === null` — or seeding from `snapshot.sizes` when there is no layout to update — would settle both.

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+        this.panels().forEach((panel, panelIndex) => {
+            if (free[panelIndex] && panel.collapsed() !== pins[panelIndex]) panel.collapsed.set(pins[panelIndex]);
+        });
+
+        this.applySizes(resizeSplitterSizesAt(snapshot.sizes, index, delta, this.minSizes(), this.maxSizes()));
+    }
+
+    /** @docs-private */
+    handleResizeEnd(): void {
+        const snapshot = this.dragSnapshot;
```
lskramarov, line 887: **The drag baseline mixes two different measurements; the one that would be right is discarded.**

`size` here is what `KbqResizer` measured off the DOM at pointer-down plus the pointer travel; `snapshot.sizes[index]` is the splitter's own logical track size. Whenever the two diverge, the difference lands as an instant jump of the boundary on the first `pointermove`.

Two concrete sources of divergence:

- `KbqResizer.getResizableSize()` returns `parseFloat(style.width)` under `content-box`, and the library ships no global `border-box` reset (see the note in `core/styles/common/_select.scss`). Give `kbq-splitter-panel` `padding: 8px` in a consuming app and the resizer reports `track − 16`: the first move computes `delta = travel − 16` and the boundary jumps 16px before it starts tracking the pointer.
- `containerSize` comes from `clientWidth`/`clientHeight`, which are integers, while the panel is measured with fractional `getBoundingClientRect()` under `border-box` — a permanent sub-pixel offset on every drag.

`KbqResizerSizeChangeEvent` from `resizeStart` already carries the exact origin, and `KbqSplitterPanel.handleResizeStart` (line 449) receives it as `_event` and throws it away. Storing it in `dragSnapshot` and computing `delta = size - startSize` makes the arithmetic independent of how either side measures.

packages/components/splitter/deprecated/splitter.component.ts
```diff
@@ -236,7 +264,7 @@ export class KbqGutterGhostDirective {
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     host: {
-        class: 'kbq-splitter'
+        class: 'kbq-splitter-component'
```
lskramarov, line 267: **Renaming the root class is a breaking change with no migration and no warning.**

`kbq-splitter` → `kbq-splitter-component` silently stops matching every consumer stylesheet written against the old component. This repo's own dev app had to delete exactly that kind of rule in this PR:

```scss
kbq-splitter.dev-custom-color > kbq-gutter { background-color: #03a9f4; }
```

After `ng update` a consumer's imports are rewritten, TypeScript compiles, nothing warns — and the splitter renders unstyled.

`migrations/splitter-deprecated-path/README.md` has a "What it does _not_ do (manual)" table listing templates and moving off `/deprecated`, but the class rename is not in it. At minimum it belongs in that table; better still, the migration could `logMessage` a warning whenever it rewrites a project, the way a breaking change is normally surfaced.

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+ * @example
+ * ```html
+ * <kbq-splitter>
+ *     <kbq-splitter-panel [minSize]="100">…</kbq-splitter-panel>
+ *     <kbq-splitter-panel>…</kbq-splitter-panel>
+ * </kbq-splitter>
+ * ```
+ */
+@Component({
+    selector: 'kbq-splitter',
```
lskramarov, line 575: **Selector and `exportAs` collide with the deprecated component, in exactly the state the migration creates.**

`KbqSplitter` keeps `selector: 'kbq-splitter'` / `exportAs: 'kbqSplitter'`, identical to `KbqSplitterComponent` at `deprecated/splitter.component.ts:260`.

A consumer migrating screen by screen will at some point have one template that renders a migrated splitter and a not-yet-migrated one. Importing both `KbqSplitter` and the deprecated `KbqSplitterModule` into the same standalone component's `imports` throws:

> NG0300: Multiple components match node with tagname kbq-splitter

and `#ref="kbqSplitter"` becomes ambiguous too. Unlike the `KbqSplitterModule` name clash the README already calls out ("`KbqSplitterModule` is the silent one"), this one has no import-alias workaround — the consumer has to split the template.

Worth either documenting the constraint in the README's manual table or giving the deprecated component a distinct selector alongside the old one.

apps/docs/src/app/seo-descriptions.ts
```diff
@@ -264,8 +264,8 @@ export const DOCS_SEO_DESCRIPTIONS = {
         "ru": "Сплит-кнопка объединяет несколько связанных действий в одном элементе управления. Основное действие доступно сразу, а дополнительные варианты скрыты в выпадающем меню."
     },
     "splitter": {
-        "en": "Splitter divides an area into resizable horizontal or vertical panels and supports nested layouts.",
-        "ru": "Splitter разделяет область на изменяемые по размеру горизонтальные или вертикальные панели и поддерживает вложенные компоновки."
+        "en": "Splitter resizes neighboring panels by dragging the separator.",
```
lskramarov, line 267: **AGENTS.md lists this file as generated and off-limits in feature branches.**

> Committed generated files — never hand-edit: `packages/docs-examples/example-module.ts`, `tools/public_api_guard/**`, `tools/check-public-api-any/baseline.json`, `apps/docs/src/llms.txt`, `apps/docs/src/llms-full.txt`, `apps/docs/src/sitemap.xml`, `apps/docs/src/prerender-routes.txt`, **`apps/docs/src/app/seo-descriptions.ts`**, `apps/docs/src/assets/versions.json`. The docs metadata files are refreshed by the release scripts (`release:extract-docs-meta`); leave them alone in feature branches.

Both splitter entries are edited here by hand. `release:extract-docs-meta` regenerates the file from the markdown anyway, so the edit either shows up as a spurious diff at release time or conflicts on rebase. Unlike `example-module.ts` or the API guards, there is no in-branch regeneration command that makes it legitimate — the updated `splitter.{en,ru}.md` is the source and is enough.

packages/components-dev/splitter/deprecated/splitter-overview-example.ts
```diff
@@ -0,0 +1,33 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
```
lskramarov, line 1: **These five dev files are dead on arrival — nothing imports them.**

`packages/components-dev/splitter/module.ts` now imports only the new `packages/docs-examples/components/splitter` examples, and no `dev-splitter-deprecated` project was added to `angular.json` (nor a `dev:*` script to `package.json`).

`git grep` for `DevDeprecatedSplitterOverviewExample`, `…FixedExample`, `…VerticalExample`, `…NestedExample`, `…DynamicDataExample` and their selectors finds only their own definitions. ~150 lines that CI lints and type-checks on every run but that can never be opened, and that will rot as the deprecated component changes.

Either wire them into a dev app (angular.json project + `dev:splitter-deprecated` script) or drop them — the deprecated splitter already has its own e2e scenarios and spec for coverage.

packages/schematics/src/migrations/splitter-deprecated-path/index.ts
```diff
@@ -0,0 +1,68 @@
...
+ * component it used to export (`KbqSplitterComponent`, `KbqSplitterAreaDirective`,
+ * `KbqGutterDirective`, `KbqGutterGhostDirective`, `KbqSplitterModule`, `Direction`) moved to
+ * `@koobiq/components/splitter/deprecated`. This rewrites the import specifier only; the API itself
+ * is unchanged at its new path.
+ *
+ * Quote-anchored (`(['"])@koobiq/components/splitter\1`) so it matches only the exact, bare module
+ * specifier — never a prefix of an already-migrated `/deprecated` import, nor an unrelated sibling
+ * package whose name merely starts with "splitter".
+ */
+const FROM = `(['"])@koobiq/components/splitter\\1`;
```
lskramarov: **This migration is a byte-for-byte copy of `scrollbar-deprecated-path/index.ts`** with `scrollbar` swapped for `splitter`.

Same `FROM`/`TO` construction, same `migrate()`, same `rootDir.visit` walk with the same `node_modules` / `/dist/` filters, same `fix ?? true` comment, same two log lines — 68 duplicated lines.

Any fix to the traversal now has to be made twice and will drift. For instance `filePath.includes('/dist/')` misses a workspace-root `dist`, and neither copy visits `.html`/`.scss`. A shared

```ts
deprecatedPathMigration('splitter')
```

factory under `packages/schematics/src/utils/` would reduce each of these to its entry-point name — and the next `*-deprecated-path` migration to one line.

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+    if (size === null || size === undefined) return null;
+
+    if (typeof size === 'number') return Number.isFinite(size) ? size : null;
+
+    const trimmed = size.trim();
+    const value = parseFloat(trimmed);
+
+    if (!Number.isFinite(value)) return null;
+
+    return trimmed.endsWith('%') ? (value / 100) * total : value;
```
lskramarov: **Any unit that is not `%` is silently reinterpreted as pixels.**

`parseFloat('10rem')` → `10`, which does not end with `%`, so `<kbq-splitter-panel minSize="10rem">` yields a **10px** minimum. Same for `em`, `vw`, `ch`, `pt`. The panel appears to have no meaningful minimum and nothing warns.

That is a natural mistake to make given the `KbqSplitterSize` docs ("a number of pixels, a `<number>px` string, or a `<number>%` share") — the type is `number | string`, so nothing stops it at compile time either.

Rejecting a trailing unit other than `px`/`%` (returning `null`, which already means "take whatever is left") would make the mistake visible, and matches the function's own stated contract that a size which "cannot be parsed" returns `null`.

packages/components/splitter/splitter.spec.ts
```diff
@@ -1,291 +1,1048 @@
...
+        expect(getSizes(fixture)).toEqual([300, 300]);
+    });
+
+    it('should fall back to equal tracks until it has been measured', () => {
+        const fixture = createComponent(TestSplitter);
+
+        expect(getSplitter(fixture).style.gridTemplateColumns).toBe('repeat(2, minmax(0, 1fr))');
+    });
+
+    describe('layout', () => {
```
lskramarov, line 520: **The pointer-drag lifecycle has no unit coverage at all.**

The suite covers the pure helpers, the keyboard, the cursor, ARIA, appearance, options and the `layout` input — but never `handleResizeStart` → `handleResizeTo` → `handleResizeEnd`. So `settleCollapse`, `pinsFor`, `gapPins` and `applySizes` are untested by Jest.

The one test that touches the path (`should keep following its declared sizes after a press that moved nothing`) dispatches a lone `pointerdown` and never ends the drag.

Both `settleCollapse` defects flagged in this review live in that gap, and the e2e collapse tests happen to run with `layout === null`, where `defaultSizes()` hides the restore bug.

These methods are public on `KbqSplitterGroup`, so a spec can drive the whole gesture directly on the component instance — no pointer capture needed in jsdom:

```ts
const splitter = fixture.debugElement.query(By.directive(KbqSplitter)).componentInstance;

splitter.handleResizeStart(0);
splitter.handleResizeTo(0, 40);
splitter.handleResizeEnd();
```

apps/docs/src/app/structure.ts
```diff
@@ -857,7 +857,8 @@ const structure: DocsStructure = makeStructure({
                     svgPreview: 'scrollbar',
                     hasApi: true,
                     apiId: 'scrollbar',
-                    hasExamples: true
+                    hasExamples: true,
+                    isNew: expiresAt('2026-10-11')
```
lskramarov, line 861: Unrelated to the splitter: this adds the "new" badge to the **scrollbar** entry.

`structure.ts` is the single source of the docs navigation, the sitemap, the prerender route list and `llms.txt`, so a revert of the splitter work also reverts this — and a reviewer scanning the diff for splitter changes has no reason to look here. If the scrollbar badge is genuinely missing, it deserves its own commit.

packages/components/splitter/splitter.en.md
```diff
@@ -1,21 +1,83 @@
...
+| <span class="docs-hot-key-button">Home</span>                                                              | Sets the smallest available size of the panel. A panel already at its minimum collapses, when it is collapsible. |
+| <span class="docs-hot-key-button">End</span>                                                               | Sets the largest available size of the panel, squeezing the neighboring panels when needed.                      |
+| <span class="docs-hot-key-button">Enter</span>                                                             | Collapses and expands the panel, when `collapsible` is set.                                                      |
+| <span class="docs-hot-key-button">F6</span> \ <span class="docs-hot-key-button">Shift + F6</span>          | Moves focus to the next or the previous panel.                                                                   |
+| <span class="docs-hot-key-button">Double-click</span>                                                      | Resets the sizes to the ones the panels declare. A second double-click sets the panel to its minimum size.       |
```
lskramarov, line 83: `Double-click` is a pointer gesture sitting inside the `## Keyboard interaction` table (same at `splitter.ru.md:83`). A keyboard-only reader is told a mouse gesture is available to them, and someone looking for the pointer gestures will not think to look under a keyboard heading.

Also missing from both markdown pages: the two-way **`layout`** input — the one a host needs in order to persist and restore an arrangement, arguably the headline addition — and `separatorAriaLabel`. Right now the only description of `layout`'s `null`-means-derived semantics is the source JSDoc.

packages/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1116 @@
...
+    total: number
+): number[] {
+    const next = sizes.map((size, index) => clamp(size, min[index], max[index]));
+
+    for (let pass = 0; pass <= next.length; pass++) {
+        const diff = total - sumOf(next);
+
+        if (Math.abs(diff) < EPSILON) break;
+
+        const open = next.reduce<number[]>(
```
lskramarov: Minor: `[...indices, index]` inside a `reduce` allocates a fresh array per element. `fitSplitterSizes` runs this once per redistribution pass, the pass loop is bounded by the panel count, and `sizes()` recomputes on every `pointermove` of a drag.

It is exactly a `filter`, which allocates once instead of once per element and reads more directly:

```ts
const open = next
    .map((_, index) => index)
    .filter((index) => (diff > 0 ? next[index] < max[index] : next[index] > min[index]));
```

The same applies to the `flexible` reduce in `defaultSizes` (line 649).

lskramarov (commented):
## Code review — `xhigh`

A thorough rewrite with unusually good comments and e2e coverage. 13 findings below, ranked most severe first.

**Correctness**

1. A drag-collapsed panel does not return to its remembered size when a host or the keyboard expands it — it lands on its `minSize` instead. Demonstrable on the shipped `splitter-collapsible` example.
2. `settleCollapse` turns a still-derived (`null`) layout into `[]`, emitting a meaningless layout and swallowing the first double-click.
3. The drag delta is measured against the splitter's own logical size rather than the `resizeStart` size the resizer already reports and the panel discards.
4. The deprecated root class rename (`kbq-splitter` → `kbq-splitter-component`) is an unannounced breaking change for consumer CSS.
5. Old and new splitter share `selector` and `exportAs`, so they cannot appear in one template scope — the mid-migration state the migration is meant to support.
6. `resolveSplitterSize` silently reads `10rem` as `10px`.

**Conventions / cleanup**

7. `apps/docs/src/app/seo-descriptions.ts` is hand-edited; AGENTS.md lists it as generated and off-limits in feature branches.
8. Five dev files under `components-dev/splitter/deprecated/` are never imported anywhere.
9. The migration is a verbatim copy of `scrollbar-deprecated-path`.
10. The `isNew` badge added to the **scrollbar** docs entry is unrelated to this PR.
11. No unit coverage for the drag lifecycle — which is where findings 1 and 2 hide.
12. Docs: double-click listed under "Keyboard interaction"; `layout` and `separatorAriaLabel` undocumented.
13. Spread-in-`reduce` in `fitSplitterSizes` / `defaultSizes`.

packages/components/splitter/splitter.spec.ts
artembelik: уже покрыто в 7b1f61bae: `describe('collapsing by drag')` проходит весь жест через handleResizeStart, handleResizeTo и handleResizeEnd, а `gapPins` в этом коммите удалён

packages/components/splitter/splitter.ts
artembelik: размер восстанавливается точно, но место забирается обратно поровну у всех открытых панелей, а не у той, что его заняла - так `layout` всегда в сумме 100 и при отпускании ничего не сдвигается

packages/components-dev/splitter/deprecated/splitter-overview-example.ts
artembelik: осознанное решение, перенес примеры в dev приложение, они будут удалены вместе с устаревшим модулем

NikGurev (approved):

lskramarov (approved):

PR #2046, opened 2026-09-15
feat(docs): compile documentation pages from MDX and prerender them with live examples (#DS-5014)

artembelik:
## Summary

documentation pages move from Markdown, which was turned into HTML at build time and fetched by the app at runtime, to MDX next to the code. `tools/docs-pages` compiles each page into an Angular component (`dist/docs-pages`, imported as `@koobiq/docs-pages`), and `docsPageResolver` picks the page for the route. live examples are now part of the page and render during prerender, instead of being attached in the browser once the document was fetched

a page places an example with `<Example id="alert-overview" />`. the compiler fails the build, with the position in the file, on an unknown example id, on HTML it does not accept and on MDX it cannot compile, so a broken page does not reach the site

## List of notable changes:

- **added** `tools/docs-pages`: the MDX compiler, the page emitter and `build:docs-pages`, which `build:docs-content` runs and `docs:start:dev` keeps running in watch mode
- **updated** 302 pages from `.md` to `.mdx` (overview and examples tabs, guides, data grid): the `<!-- example(...) -->` markers become `<Example id="..." />`, the comments for cspell and prettier become `{/* */}`, the rest of what MDX rejects is rewritten, and headings that skipped a level are fixed
- **updated** the docs app: the overview and examples tabs render the compiled page through `DocsComponentPageComponent`, a route without a page redirects to `/404`, and a lazy chunk missing after a deploy reloads the page (`docsReloadOnChunkLoadError`)
- **updated** `DocsLiveExampleViewerComponent` to take the example class from the page. examples that cannot render on the server (`packages/docs-examples/server-rendering.ts`) load in the browser after render behind a skeleton, and their code stays out of the page chunk
- **added** `SsrErrorHandler` to the docs server config: an error during prerender fails `docs:build` instead of publishing a page without the part that failed
- **removed** the Markdown pipeline (`tools/markdown-to-html`), `DocsExampleViewerComponent` and the unused `additionalComponents` of the example catalogue
- **updated** the SEO descriptions and `llms.txt` generators to read the MDX pages, cspell and lint-staged to cover `.mdx`, and Jest to compile the ESM-only unified packages
- **added** MDX support in VS Code with the recommended `unifiedjs.vscode-mdx`: the editor-only `tsconfig.mdx.json`, reached through `packages/tsconfig.json` and `docs/tsconfig.json`, and `tools/docs-pages/mdx-components.ts`, which completes example ids, flags an unknown one and knows the HTML elements the compiler accepts
- **fixed** the StackBlitz package path of examples, and renamed the `icon-button-custom size` example folder to `icon-button-custom-size`

- **the converted pages.** most of the diff is `.md` → `.mdx`, and the content edits are what MDX syntax requires. worth opening a few pages with raw HTML in the preview, such as the keyboard tables and `<details>` blocks
- **server rendering of examples.** every example outside `server-rendering.ts` now renders during `docs:build`. any render error fails the build, and `docs:start:dev` stops on it too, on purpose
- **the API tab is unchanged.** its HTML still comes from `docs:api-gen` and is fetched at runtime
- **the tsconfig files for MDX** are read by the editor only (TypeScript 5.6+), no build uses them


artembelik:
кратко:
- подсказки и валидация при добавлении `<Example id="..." />` в документацию (mdx) - для мясных прокси
- документация работает без включения js - для роботов
- живое обновление при изменении документации (mdx) в режиме `yarn docs:start:dev`

NikGurev:
Предлагаю сделать переключение страницы сразу по нажатию, сейчас выглядит "залипшим"

https://github.com/user-attachments/assets/458c3d75-133c-491a-8465-797c4ad63271

NikGurev (approved):

packages/docs-examples/server-rendering.ts
```diff
@@ -0,0 +1,18 @@
+import type { LiveExample } from './example-module';
+
+/**
+ * Examples that cannot render on the server: the documentation site loads them in the browser once the page has
+ * rendered, and `components-dev/ssr` does not prerender them. Both lists are matched against `EXAMPLE_COMPONENTS`.
+ */
+export const EXAMPLE_IDS_WITHOUT_SERVER_RENDERING = new Set([
```
lskramarov: **Clock-reading examples now prerender the build machine's time.**

Only AG Grid is excluded from server rendering, so every example that reads the clock renders at prerender time and ships CI's date, time and timezone as static HTML.

Scanning all 601 catalogue entries against all 302 MDX pages: 17 example sources call `adapter.today()` / `DateTime.now()` / `new Date()`, and **15 of them are referenced by an `<Example>` tag** — `absolute/relative/range/duration-date-formatter`, `date-formatter-typical-use`, `date-formatter-special-use`, `timepicker-overview`, `timepicker-variations`, `timepicker-field-validation`, `table-sticky-header`, `filter-bar-date-limits`, `inline-edit-date-time`, `notification-center-push`, `time-range-custom-option`, `time-range-custom-range-types`.

`/en/other/date-formatter/overview` is in `prerender-routes.txt:189` and embeds `<Example id="relative-date-formatter" />`, whose entire output is built from `this.adapter.today()` at construction. The published page therefore reads "yesterday, 12:00" relative to whenever CI ran (UTC); crawlers and pre-hydration readers see frozen, wrong dates, and every value jumps once hydration re-runs change detection.

Before this PR examples rendered only in the browser, so the clock was always the reader's. Worth deciding whether these belong in the browser-only set, or whether the examples should take an injectable "now".

tools/docs-pages/emit-page.ts
```diff
@@ -0,0 +1,128 @@
...
+        members.push(
+            `protected readonly codeBlocks: KbqCodeBlockFile[] = ${indent(JSON.stringify(page.codeBlocks, null, 4), 1)};`
+        );
+    }
+
+    if (page.examples.length > 0 || page.browserExamples.length > 0) {
+        for (const [importPath, componentNames] of [...componentNamesByImportPath].sort(([a], [b]) =>
+            a.localeCompare(b)
+        )) {
+            imports.push(renderImport(componentNames, `@koobiq/docs-examples/${importPath}`));
```
lskramarov, line 54: **This barrel import defeats the browser-only deferral for `content-panel`.**

`content-panel.{en,ru}.mdx` uses `content-panel-overview` and `content-panel-state-saving` (server-renderable) plus `content-panel-with-grid` (in `EXAMPLE_IDS_WITHOUT_SERVER_RENDERING`). All three share `importPath: 'components/content-panel'`, so `emitPage` emits **both**:

- `import { ContentPanelOverviewExample, ContentPanelStateSavingExample } from '@koobiq/docs-examples/components/content-panel'` (this line), and
- `ContentPanelWithGridExample: { load: () => import('@koobiq/docs-examples/components/content-panel') }` (line 67)

— the same module. `packages/docs-examples/components/content-panel/index.ts` statically imports `ContentPanelWithGridExample` and holds it in the `EXAMPLES` array, so it cannot be tree-shaken; that example imports `ag-grid-angular`/`ag-grid-community` and runs `ModuleRegistry.registerModules([AllCommunityModule])` at module scope.

Net effect on `/{en,ru}/components/content-panel/overview`: AG Grid lands in exactly "the page chunk that the route and hydration wait for" that the comment on lines 65–66 says this avoids, and its module side effect runs on the server. (It doesn't crash — I checked that importing `ag-grid-community` in bare Node is fine — but the deferral is gone.)

This is the only mixed-`importPath` page today, and no test covers it: `emit-page.spec.ts:24-28` deliberately gives the browser example a different `importPath`.

Either emit per-example deep imports instead of the barrel, or exclude the whole `components/content-panel` path.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+    if (node.type === 'mdxTextExpression' || node.type === 'mdxFlowExpression') return '';
+
+    return 'children' in node ? node.children.map(getVisibleText).join('') : toString(node);
+};
+
+/**
+ * The id of a heading, derived from its text, so links such as `#size` or `#размер` keep working.
+ * Braces are dropped: an attribute has no room for the comment that keeps them from interpolating.
+ */
+const getHeadingId = (heading: Heading): string =>
```
lskramarov, line 137: **Heading ids are not de-duplicated within a page.**

Scanning the corpus:

| page | duplicates |
| --- | --- |
| `docs/guides/migration.en.mdx` | 14× `running-the-migration`, 13× `what-is-fixed-automatically`, 12× `what-you-need-to-fix-manually` |
| `docs/guides/migration.ru.mdx` | 14× / 13× / 8× / 4× |
| `packages/components/toast/toast.{en,ru}.mdx` | 2× `disappearance` / `исчезновение` |

`DocsAnchorsComponent.createAnchors` maps each `.docs-header-link` to `href: header.id`, so the table of contents lists 14 identical "Running the migration" entries and `document.getElementById(fragment)` always resolves the first — every per-version anchor scrolls to the v20 section.

The old renderer had the same slug bug, so it isn't a regression, but this function is the rewrite of it and the duplicate ids now ship as static prerendered HTML. A per-page counter suffix (`-2`, `-3`) would fix it where it is now cheapest.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -61,34 +72,49 @@ interface ExampleFileData {
...
-    readonly example = input<string | null>(null);
+    readonly component = input.required<Type<unknown> | DocsExampleLoader>();
 
     protected readonly isSourceShown = signal(false);
 
     files: KbqCodeBlockFile[] = [];
 
     /** Data for the currently selected example. */
-    exampleData: LiveExample;
+    readonly exampleData = computed(() => EXAMPLE_COMPONENTS[this.example()]);
```
lskramarov, line 92: **Both guards for an unknown key were removed, but the compile-time guarantee validates a different artifact than the app loads.**

`exampleData` has no guard, and `exampleId` (line 94) reads `.selector` on every render — via the class interpolation and `[exampleId]` on the StackBlitz button — so a missing key throws `TypeError: Cannot read properties of undefined (reading 'selector')` *during rendering*.

The replacement guarantee is split across two artifacts:

- `tools/docs-pages/generate-docs-pages.ts` validates ids against `packages/docs-examples/example-module.ts` (source)
- `apps/docs/tsconfig.app.json` maps `@koobiq/docs-examples` to `dist/docs-examples`

and `generate-docs-pages.ts:99` states the catalogue is read once per watcher process ("restart the watcher after adding an example"). Add an example while `docs:watch-pages` is running and they drift.

Previously this logged `Could not find example: X` and rendered the localized "code example not found" block, leaving the rest of the page usable. Now the whole page render dies — and under prerender the new `SsrErrorHandler` rethrows, so `docs:build` fails on the entire route instead of degrading one example.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -149,35 +177,31 @@ export class DocsLiveExampleViewerComponent extends DocsLocaleState {
...
-        this.loadExampleComponent(exampleName)
-            .then(() => this.exampleHeight.set(null))
-            .catch((error) => console.error(`Could not load example '${exampleName}': ${error}`));
+        if (!isLoader(component)) return;
 
-        this.generateExampleTabs();
+        component
+            .load()
+            .then((type) => this.loadedComponent.set(type))
+            .catch((error) => console.error(`Could not load example '${this.example()}': ${error}`));
```
lskramarov: **A rejected `load()` leaves the skeleton and `aria-busy="true"` permanently.**

`isLoading = computed(() => isLoader(this.component()) && !this.loadedComponent())` (line 103) has no failure input, and `afterNextRender(() => this.loadComponent())` (line 132) runs once per instance, so nothing retries.

A tab opened before a deploy visits a page with a browser-only example; the dynamic import 404s on the stale chunk name. `docsReloadOnChunkLoadError` does **not** cover this — it is wired through `withNavigationErrorHandler` and only sees router `NavigationError`s, never a component-initiated import.

The reader is left with `<div class="docs-live-example__example" aria-busy="true">` and a skeleton indefinitely, a screen reader announcing "busy" forever, and only a console line. `docs-live-example-viewer.spec.ts:222` covers the resolve path only.

A rejected-state signal that renders an error with a retry (or routing this through the chunk-reload path) would close it.

apps/docs/src/config.server.ts
```diff
@@ -53,10 +57,24 @@ const provideServerWindow = (): Provider => {
     };
 };
 
+/**
+ * Fails the prerender, and with it `docs:build`, on an error that Angular's default handler only logs: the page would
+ * be published without the part that failed. The development server renders on the server as well and stops on such
+ * an error too, which is intended: the bug shows up at once instead of hiding in the log.
+ */
+class SsrErrorHandler extends ErrorHandler {
+    handleError(error: unknown): never {
```
lskramarov, line 66: **Rethrowing here only fails the prerender for bootstrap-chain errors; async ones escape.**

`NgZone.onHandleError` does `zone.runOutsideAngular(() => zone.onError.emit(error))`. Throwing from that subscriber is caught by RxJS's `SafeSubscriber` and re-reported via `reportUnhandledError` → `setTimeout(() => { throw err })`. `ChangeDetectionSchedulerImpl.tick()` takes the same shape: it catches, calls `applicationErrorHandler(e)`, and the rethrow escapes into a callback scheduled with `runOutsideAngular`.

So an example whose `setTimeout` or rxjs stream throws after the first tick lets `renderApplication()` **resolve**; the page is serialized and written with the broken markup, and the deferred exception then kills the prerender worker — possibly while it is rendering an unrelated route, so the build failure names the wrong page or none at all.

Live examples now rendering on the server makes this class newly reachable. Only synchronous/bootstrap-chain errors get the intended behaviour, and those already rethrow via `_callAndReportToErrorHandler`.

Worth noting the gate is also half-blind by construction: hydration mismatches — the failure mode the `BLOCK_ELEMENTS`/`inParagraph` machinery exists to prevent — are thrown by the *browser* after it parses the server markup and never reach this handler.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+
+    const renderParagraph = (paragraph: Paragraph, parentContext: RenderContext): string => {
+        const context = { ...parentContext, inParagraph: true };
+        const imageIndex = paragraph.children.findIndex((child) => child.type === 'image');
+        // The text after an image is its caption.
+        const content =
+            imageIndex === -1
+                ? renderAll(paragraph.children, context)
+                : renderAll(paragraph.children.slice(0, imageIndex + 1), context) +
+                  `<em>${renderAll(paragraph.children.slice(imageIndex + 1), context)}</em>`;
```
lskramarov: **The caption `<em>` is emitted whenever the paragraph contains an image, even with nothing to caption.**

`renderAll(children.slice(0, imageIndex + 1)) + \`<em>${renderAll(children.slice(imageIndex + 1))}</em>\`` fires whenever `imageIndex !== -1`, with no check that anything follows the image and no handling of a second one.

Running the compiler over both cases:

- `a` alone → `<p class="kbq-markdown__p">a<em></em></p>` — an empty `<em>` on every uncaptioned image.
- `a` / Caption A / `b` / Caption B in one paragraph (soft breaks) → `<em>Caption A  Caption B</em>` — the **second image and its caption render inside the first image's italic caption**, and `.kbq-markdown__img + em` no longer matches the second one.

The replaced renderer (`tools/markdown-to-html/marked/docs-marked-renderer.ts:39` `setImageCaption`) looped over every `` case fires the moment anyone adds an image without a caption line.

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
```diff
@@ -61,34 +72,49 @@ interface ExampleFileData {
...
 
     protected readonly isSourceShown = signal(false);
 
     files: KbqCodeBlockFile[] = [];
 
     /** Data for the currently selected example. */
-    exampleData: LiveExample;
+    readonly exampleData = computed(() => EXAMPLE_COMPONENTS[this.example()]);
+
+    protected readonly exampleId = computed(() => this.exampleData().selector.replace('-example', ''));
```
lskramarov, line 94: **The StackBlitz key is reconstructed from the selector, and the guard that should catch a miss is dead.**

This computed feeds `<docs-stackblitz-button [exampleId]="exampleId()" />`, which does `new ExampleData(id)` and then `createStackBlitzForExample(id, data)`, dereferencing `EXAMPLE_COMPONENTS[exampleId].packagePath` in `stackblitz-writer.ts:174`. The viewer already receives the authoritative key as `example()`.

The guard cannot help: `ExampleData`'s constructor returns early on an unknown key but still yields a truthy instance (`example-data.ts:29-31` assigns nothing), so `if (!id || !data) return;` in `stackblitz-button.ts:53` never fires and the click throws an uncaught `TypeError` instead.

I checked all 601 entries and `selector.replace('-example','') === id` holds today, so this is latent — but the parallel assumption is *already* violated once for paths (`tree-select-with-multiline-matcher` has `packagePath: components/tree-select/tree-select-with-multiline-matcher-overview`), which is exactly why line 174 was changed to use `packagePath` in this PR. The key half of the same bug is still there.

The fix is `[exampleId]="example()"` in the template — the viewer already has the real key — plus a real miss signal on `ExampleData` so the guard in `stackblitz-button.ts` can do its job.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+
+        for (const attribute of element.attributes) {
+            if (
+                attribute.type !== 'mdxJsxAttribute' ||
+                (attribute.value !== null && typeof attribute.value !== 'string')
+            ) {
+                return fail(element, `the attributes of <${tag}> must be plain strings`);
+            }
+
+            attributes.set(attribute.name, attribute.value);
```
lskramarov, line 319: **Attribute *names* are never validated — only values — so Angular's canonical binding prefixes survive into the generated template.**

The check above only rejects non-string values; the name is then copied verbatim. `bind-title`, `on-click`, `let-x` and `ref-x` are all valid JSX attribute names, so `<span on-click="doThing()">x</span>` in a page compiles to exactly that markup.

Fed to `@angular/compiler`'s `parseTemplate`, that parses with **zero errors** as an *event binding* on the generated page component; `bind-title="codeBlocks"` likewise becomes a property binding — to a member the generated class actually has on any page with a fenced code block.

The outcome is either an AOT error naming a member of a generated file the author cannot see, or a silently dead listener — in a compiler whose stated invariant (lines 112–118 and the `{{`/`}}` guard on 167–174) is that page text can never become Angular syntax. Two independent passes confirmed this against `parseTemplate`; it is the one path left open.

Rejecting names matching `/^(bind|on|let|ref)-/` (or allow-listing plain attribute names) would close it the same way the value guard does.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+ * The id of a heading, derived from its text, so links such as `#size` or `#размер` keep working.
+ * Braces are dropped: an attribute has no room for the comment that keeps them from interpolating.
+ */
+const getHeadingId = (heading: Heading): string =>
+    getVisibleText(heading).trim().toLowerCase().replace(/\s/g, '-').replace(/[{}]/g, '');
+
+/** Heading levels that get an id and appear in the table of contents. */
+const LINKED_HEADING_DEPTHS = [2, 3, 4, 5];
+
+const isComment = (expression: string): boolean => /^\s*\/\*[\s\S]*\*\/\s*$/.test(expression);
```
lskramarov: **`isComment` drops content sandwiched between two comments.**

Verified in node: `/^\s*\/\*[\s\S]*\*\/\s*$/.test('/* a */ foo /* b */')` → `true`.

So an MDX page writing `{/* cspell:ignore x */ title /* prettier-ignore */}` renders as **nothing**, silently, instead of reaching the `imports, exports and {expressions} are not supported` guard on line 396. Every other unsupported construct in this compiler fails loudly with a `file:line:column`; this one loses reader-visible content without a word.

The current corpus only uses single comments, so nothing breaks today.

```suggestion
const isComment = (expression: string): boolean => /^\s*\/\*(?:[^*]|\*(?!\/))*\*\/\s*$/.test(expression);
```

package.json
```diff
@@ -177,12 +184,14 @@
...
         "build:schematics": "rimraf packages/schematics/dist && rollup --failAfterWarnings -c packages/schematics/rollup.config.js && node ./packages/schematics/scripts/copy-meta-to-dist.js",
-        "docs": "yarn run build:components && yarn run build:components-experimental && yarn run styles:build-all && yarn run build:angular-luxon-adapter && yarn run build:angular-moment-adapter && yarn run build:tokens:data && yarn run build:docs-content && yarn run build:package-docs-content && yarn run build:docs-examples-module && yarn run build:docs-examples && yarn run docs:start:dev",
-        "docs:start:dev": "ng serve --configuration=development koobiq-docs",
+        "docs": "yarn run build:components && yarn run build:components-experimental && yarn run styles:build-all && yarn run build:angular-luxon-adapter && yarn run build:angular-moment-adapter && yarn run build:tokens:data && yarn run build:docs-examples-module && yarn run build:docs-content && yarn run build:package-docs-content && yarn run build:docs-examples && yarn run docs:start:dev",
+        "docs:start:dev": "yarn docs:watch-pages & ng serve --configuration=development koobiq-docs",
```
lskramarov: **`docs:start:dev` races the watcher against `ng serve`, and `docs:build` has no declared dependency on the pages at all.**

Both `apps/docs/tsconfig.json` and `tsconfig.app.json` map `@koobiq/docs-pages` to `../../dist/docs-pages/index.ts`, which exists only after `generate-docs-pages.ts` has compiled all 302 pages.

On a fresh clone or right after `yarn clean` (which `rimraf`s `dist`), `ng serve` starts resolving before the watcher writes `index.ts` and fails on an unresolved `@koobiq/docs-pages`. The watcher never sets a non-zero exit code, so a **first compile that fails** shows the same module-resolution error rather than the MDX error that actually caused it.

`yarn clean && yarn docs:build` hits the same wall, since `docs:build` is just `ng build koobiq-docs --configuration production`. The correct ordering exists only inside `.github/workflows/actions/build-docs/action.yml` and the `docs` mega-script.

Sequencing the first compile before `ng serve` (and making `docs:build` depend on `build:docs-content`) would turn both into the right error. For the record, Yarn 4.1.1's shell *does* implement `&` backgrounding — I tested it on this machine — so that part is fine on Windows.

lskramarov (commented):
**Automated review** — 15 findings from a multi-angle pass over the diff (correctness, removed behaviour, cross-file, language pitfalls, wrapper/SSR, reuse, simplification, efficiency, altitude, conventions), each verified against the branch — the last pass compiled all 302 pages through the new compiler in a sandbox and re-parsed every result with `@angular/compiler`. Ordered most severe first in the inline comments below.

The migration itself checks out well: all 302 `.mdx` files are prettier-clean and use only allowed elements, all 1201 `<Example id>` refs resolve, heading ids are byte-identical to the old renderer for all 1819 headings, the brace/interpolation escaping is correct (including runs of 3+ braces), `example-module.ts` is properly regenerated, CI ordering is right, and Yarn 4 does support `&` backgrounding on Windows.

The two I would not merge without a decision on:

1. **15 clock-reading examples now prerender the build machine's date/time** — the date-formatter and timepicker pages ship frozen UTC values as static HTML.
2. **The `content-panel` page pulls AG Grid into the chunk the route and hydration await**, via the barrel import in `emit-page.ts` — the exact thing `DocsExampleLoader` exists to prevent.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+    'summary',
+    'ul'
+] as const;
+
+const HTML_ELEMENT_NAMES: ReadonlySet<string> = new Set(HTML_ELEMENTS);
+
+const VOID_ELEMENTS = new Set(['br', 'img']);
+
+/** Elements the browser moves out of `<p>` while parsing, so the hydrated DOM would not match the prerendered one. */
+const BLOCK_ELEMENTS = new Set(['details', 'div', 'li', 'ol', 'p', 'summary', 'ul']);
```
lskramarov: **Four hand-maintained tag lists have to agree, and two of the three failure modes are silent.**

`HTML_ELEMENTS` (75), `VOID_ELEMENTS` (95), `BLOCK_ELEMENTS` (98) and `BLOCK_NODES` (101) each encode one property of the same set of tags. They agree today only because `HTML_ELEMENTS` is small.

- Add `pre`, `hr`, `blockquote`, `table` or `h2` to `HTML_ELEMENTS` — all plausible, none currently allowed — without touching `BLOCK_ELEMENTS`, and `Text <pre>x</pre>` compiles happily to `<p>Text <pre>x</pre></p>`. The browser hoists it out of the paragraph and you get an NG0500 hydration mismatch on one page at runtime, not a build error.
- A new void element (`hr`, `wbr`) missing from `VOID_ELEMENTS` emits `<hr></hr>`, which the browser re-parses as `<hr>` plus a stray end tag — again silent.

Collapsing them into one table (`Record<tag, { void: boolean; block: boolean }>`) makes adding an element force the answer. The more general fix is to drop the hand-written parser model: re-parse the emitted markup (parse5 is already in `node_modules` via jsdom) and fail when the serialized round-trip differs from the input — that covers every restructuring rule in the spec, including `<li>` auto-closing, `<a>` nesting and table foster-parenting.

packages/components/tags/tag.en.mdx
```diff
@@ -1,32 +1,26 @@
...
-<!-- example(tag-with-icon) -->
-
-### “Remove” Button
-
-Tags are most commonly used with a “Remove” button to allow users to delete a selected value from an input field. In some cases, tag removal should be disabled—when that’s the case, the close (×) icon should be hidden.
-
-<!-- example(tag-with-remove-button) -->
+<Example id="tag-with-icon" />
 
 ### Long Text
```
lskramarov, line 19: **The whole `### "Remove" Button` section was deleted, not just the dangling example marker.**

`tag.en.md` on main carried, between `tag-with-icon` and `### Long Text`:

> ### "Remove" Button
>
> Tags are most commonly used with a "Remove" button to allow users to delete a selected value from an input field. In some cases, tag removal should be disabled—when that's the case, the close (×) icon should be hidden.
>
> `<!-- example(tag-with-remove-button) -->`

`tag-with-remove-button` is absent from `EXAMPLE_COMPONENTS` at both revisions — it is the **only** dead example marker in the entire 302-page corpus — so the new compiler's unknown-example check forces the `<Example>` line out. But the heading and the prose are ordinary Markdown and did not have to go with it.

`tag.ru.mdx` never had this section, so after the change EN and RU heading structures match exactly (that part reads as deliberate), but nothing on the site now documents hiding the × when removal is disabled, and the anchor disappears for any external deep link. Worth either keeping the prose under `### Removal` or calling the deletion out in the PR description.

Separately, `breadcrumbs.ru.mdx:90` was demoted `####` → `###` while `breadcrumbs.en.mdx:90` keeps `####`, so the same content now nests differently per locale (different heading size and different ToC indent). `toggle` and `typography` got the same `#####` → `####` change symmetrically in both locales; breadcrumbs is the one that drifted.

tools/docs-pages/generate-docs-pages.ts
```diff
@@ -0,0 +1,110 @@
...
+        return false;
+    }
+};
+
+if (process.argv.includes('--watch')) {
+    run();
+
+    // The example catalogue is read once: restart the watcher after adding an example.
+    for (const directory of ['packages/components', 'docs']) {
+        watch(directory, { recursive: true }, (_event, fileName) => {
```
lskramarov, line 110: **No debounce and no incremental path: one save recompiles all 302 pages twice.**

Measured on this machine: `fs.watch(dir, { recursive: true })` fires **exactly 2 `change` events** for a single `writeFileSync`, and `run()`/`generate()` are fully synchronous, so both events run a complete recompile back to back.

Each `generate()` re-globs, re-reads and re-parses all 302 files (1.465 MB through remark-mdx) on top of a ~130 ms syscall floor — glob 20 ms, reads 20 ms, `syncOutput` 90 ms (605 `existsSync`+`readFileSync`, then `listOutputFiles()` walks the same tree again). That's ~260 ms of pure syscalls plus two full corpus parses per save, and `docs:start:dev` now puts this on the normal dev path.

The callback already receives `fileName` on this line and drops it. A trailing ~50 ms debounce plus a `Map<path, { source, emitted }>` cache would make one save cost one page: the three whole-set checks (duplicate key, `findRoutesWithoutPage`, `emitPagesRegistry`) can all run over the cached records without re-reading. In watch mode `syncOutput` can diff against the previous in-memory map instead of the disk.

tools/docs-pages/compile-page.ts
```diff
@@ -0,0 +1,407 @@
...
+        const context = { ...parentContext, inParagraph: true };
+        const imageIndex = paragraph.children.findIndex((child) => child.type === 'image');
+        // The text after an image is its caption.
+        const content =
+            imageIndex === -1
+                ? renderAll(paragraph.children, context)
+                : renderAll(paragraph.children.slice(0, imageIndex + 1), context) +
+                  `<em>${renderAll(paragraph.children.slice(imageIndex + 1), context)}</em>`;
+
+        return `<p class="kbq-markdown__p">${content}</p>`;
```
lskramarov: **The markdown class prefix is hardcoded here despite `CLASS_PREFIX` being imported.**

Twelve branches write the literal: lines 189, 201, 214, 217, 326, 331, 352, 355, 357, 364, 370, 373. Every one of those tags (`p, li, ol, ul, table, thead, tbody, tr, td, th, blockquote, code, img, hr, h1–h6`) is already covered by `getElementClass(tag)` on line 106, which returns exactly `\`${CLASS_PREFIX}__${tag}\``.

This matters beyond tidiness. `apps/docs/src/app/components/live-example/markdown-content.ts` derives `DOCS_MARKDOWN_HEADING_CLASSES` from `CLASS_PREFIX` deliberately — "so a prefix change stays in sync instead of silently breaking the DOM queries that consume it" — and `anchors.component.ts:248` computes the anchor level by `indexOf` into that list. Hardcoding the prefix here puts every page on the site outside that guard: renaming `CLASS_PREFIX` compiles cleanly and silently kills the table of contents.

artembelik: подставил `CLASS_PREFIX` напрямую - `getElementClass` типизирован как `string | null`, и в каждой из 12 веток пришлось бы снимать null

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
artembelik: снял скелетон и `aria-busy`, когда `load()` отклонён - UI с ошибкой и повтором пока не делал

tools/docs-pages/generate-docs-pages.ts
artembelik: добавил дебаунс 50 мс, одно сохранение теперь пересобирает страницы один раз - инкрементальную пересборку по файлу не делал

packages/components/tags/tag.en.mdx
artembelik: выровнял `breadcrumbs.ru.mdx` по английской версии, заголовок снова `####` - удалённый раздел в `tag.en.mdx` пока не вернул

apps/docs/src/config.server.ts
artembelik: гарантия и правда неполная: обработчик ловит синхронные ошибки, асинхронные остаются за `ssr:build`, где каждый пример прогоняется отдельным маршрутом и упавший воркер называет именно его, а ошибки гидрации - за браузерным прогоном `e2e:docs`

packages/docs-examples/server-rendering.ts
artembelik: DS-5583

tools/docs-pages/emit-page.ts
artembelik: перенёс пример с гридом на страницу ag-grid - бочка content-panel больше не тянет ag-grid, `emit-page.ts` не трогал

tools/docs-pages/compile-page.ts
artembelik: свёл `HTML_ELEMENTS`, `VOID_ELEMENTS` и `BLOCK_ELEMENTS` в одну таблицу - `BLOCK_NODES` оставил отдельно, там типы узлов Markdown, а не теги, сверку через parse5 не делал

apps/docs/src/app/components/live-example-viewer/docs-live-example-viewer.ts
artembelik: оставляю без проверки во вьюере: в сборке неизвестный id не пропустят компилятор страниц и TypeScript, а разойтись каталоги могут только в dev при уже запущенном watcher

artembelik:
> Предлагаю сделать переключение страницы сразу по нажатию, сейчас выглядит "залипшим"

страница и так грузится по нажатию, просто некоторые страницы имеют размер больше обычного и поэтому видно задержку, из-за этого есть небольшой рассинхрон между выделением активной страницы в боковом меню и страницей в документации

можно пойти несколькими путями:
1. отображать скелетон/лоадер страницы во время перехода (лучше делать отдельно)
2. префетчить страницу при фокусе или ховере  - пошел по этому сценарию

lskramarov (approved):

rmnturov:
> Предлагаю сделать переключение страницы сразу по нажатию, сейчас выглядит "залипшим"

Прям в нашу тему
https://guides.kontur.ru/components/progress-indicators/global-loader/

Нам подойдет глобальный мини лоадер, скелетон, трюки с задержкой при показе скелетона, как у лоадера в селекте

artembelik:
> Прям в нашу тему

DS-3140

PR #2048, opened 2026-09-15
feat(docs): interactive migration guide filtered by version range (#DS-5169)

lskramarov:
## Summary

`/{lang}/main/migration/overview` becomes interactive: the reader picks the release they are on and the one they are going to, and only the steps in between remain. The unfiltered article stays available on a second tab, `/{lang}/main/migration/full`.

The guide remains the single source of truth — there is no second manifest to keep in sync. A build-time transform (`tools/markdown-to-html/migration`) wraps each numbered step in a `<section>` carrying the release it lands in, read from the heading's `(x.y.z)` suffix or, for the three headings that cannot spell a semver, from a `<!-- migration-step-version(x.y.z) -->` directive placed below them so the anchor ids stay put.

The filter is progressive enhancement. The prerendered page ships every step with nothing hidden, so crawlers and readers without JavaScript get the whole document; hiding only happens after hydration. The picked range lives in `?from=` / `?to=`, is validated against the releases the guide actually documents, and can never run backwards.

Two gaps in the guide surfaced while building the CI gate and are fixed here: `actions-panel-overlay-container` and `splitter-deprecated-path` were undocumented, and both were registered for a release they never shipped in.

## List of notable changes:

- **added** an interactive version filter to the migration guide, with the unfiltered article moved to a `full` tab
- **added** a build-time transform that tags every guide step with its release, plus three CI gates: step numbering and en/ru parity, transform behaviour against the real guide, and coverage of every registered schematic
- **added** migration guide steps for the `actions-panel-overlay-container` and `splitter-deprecated-path` schematics, which were registered but undocumented
- **fixed** the registration of those two schematics: both said `20.3.0-0`, but neither shipped in 20.3.0 or 20.3.1 (the registry in the 20.3.1 tag holds four schematics and neither is among them), so `ng update` skipped them for anyone already on 20.3.x — they are now `21.0.0-0`
- **updated** `DocsAnchorsComponent` with a public `refresh()` that rebuilds the outline without moving the reader, drops headings inside a hidden subtree, and keeps query params on its links
- **updated** the heading-spacing rules in `component-viewer.scss` so they survive the new section wrappers on both migration tabs

**`packages/schematics/src/migrations.json` — two entries that are not mine.** The version change touches `actions-panel-overlay-container` and `splitter-deprecated-path`. Please confirm both really land in 21.0.0; if either is meant to be backported to a 20.3.x patch, its registration should go back. Authors of those two schematics are the right reviewers for that commit.

**The step renumbering.** Two new steps at positions 7 and 8 shift the old steps 7–24 to 9–26, which changes the anchor id of every heading below them. External links into those fragments break. That is unavoidable for an ordered guide, but worth a conscious sign-off.

**Hydration.** The prerendered document carries no filter and the arriving client applies one. The picker markup renders on the server but with an empty option list, so server and client agree; `e2e.playwright-spec.ts` asserts the unfiltered prerender, the filtered client state, and a reload of a filtered URL with no console error.

**`marked` output as a parse contract.** The transform matches the exact heading markup `DocsMarkdownRenderer.heading()` emits. A change to that shape would disable the filter, so `wrap-migration-steps.spec.ts` runs the real renderer over the real guide as a backstop.

Checked locally: 188 unit tests across `apps/docs` and `tools`, 863 in `packages/schematics`, the production docs build with prerendering, and eslint / stylelint / prettier / cspell / check-e2e-types.


docs/guides/migration.ru.mdx
```diff
@@ -4,32 +4,37 @@
...
-18. **21.0.0**: ревью компонентов — сокрытие внутренних членов, переход входов на сигналы и исправления поведения.
-19. **21.0.0**: удаление устаревших событий file-upload `fileQueueChanged`/`fileQueueChange`.
-20. **21.0.0**: хранилище состояния аккордеона переехало в `core` и стало общим для всех компонентов.
-21. **21.0.0**: сохранение состояния аккордеона включено по умолчанию, ключ выводится из документа.
-22. **21.0.0**: сохранение состояния дерева включено по умолчанию, ключ — значение узла из контрола.
-23. **21.0.0**: вкладки, боковая панель и контент-панель запоминают выбор пользователя, по умолчанию.
-24. **21.0.0**: `filter-bar` запоминает выбранный фильтр и его изменения, по умолчанию.
-
-### 1. Обновление до 18.5.3
+- **До 18.5.3**: безопасная база с обновлением темизации и иконок.
```
NikGurev, line 7: Давай прогоним скиллом от ред политики, хуже не будет, зато приведет к единому языку

Сейчас либо уже перед выпуском

apps/docs/src/app/components/migration-guide/docs-migration-guide.html
```diff
@@ -0,0 +1,126 @@
...
+                        <kbq-cleaner />
+                    </kbq-form-field>
+
+                    <div kbq-select-search-empty-result>{{ t('nothingFound') }}</div>
+
+                    @for (option of componentChoices(); track option.value) {
+                        <kbq-option [value]="option.value">{{ option.label }}</kbq-option>
+                    }
+                </kbq-select>
+            </kbq-form-field>
```
NikGurev, line 80: Давай добавим короткую подсказку, чтобы объяснить зачем выбирать компоненты

Кажется полезным пояснить заранее для чего происходит выбор компонента

Ну и можно добавить мастер кнопку выбрать все - чтобы зайти и кликнуть 

@rmnturov будет ли здесь работать правило выбрать все = не выбрано ничего или излишне?

NikGurev (commented):
Текст миграций по компонентам бы сократить, а так все гуд

NikGurev (approved):

apps/docs/src/app/components/migration-guide/docs-migration-guide.html
lskramarov: тут же классичекский фильтр - если нужно видеть по конкретным компонентам то выбираешь их, если ничего не выбрать то отображается все. Я категорически против тянуть сюда выбрать всё, она здесь не к чему 100%

artembelik:
может вместо чекбокса "выполнено" который ничего (судя по всему) не делает, сделать скрытие в акордеон ?
image

---

сильно скачет контент после выбора версии и скрытия этого сообщения:
image

apps/docs/src/app/components/migration-guide/docs-migration-guide.html
artembelik: классическое не выбрано ничего = выбрано всё 😊

docs/guides/migration.ru.mdx
lskramarov: все же позже прогоню текст, тут минимальные правки..

PR #2063, opened 2026-09-17
fix(breadcrumbs)!: errors following a full review of the component (#DS-5482)

NikGurev:
## Summary

A full review pass over `breadcrumbs`, in the same series as `checkbox`, `loader-overlay` and `title`. Three inputs were published and documented but did nothing, the trail was not a list to assistive technology, its landmark was named with an English literal that also overwrote a consumer's own `aria-label`, and the vendored roving-focus primitive swallowed the paging keys and lost arrow-key order whenever an item toggled focusable.

Breaking, so the title carries `!`. No schematic ships with it: no input was renamed, so there is nothing with a single mechanical translation to rewrite. The `#### Breadcrumbs` section of the migration guide is the report instead, in both languages.

## List of notable changes:

- **fixed** `KbqBreadcrumbs.disabled`, which was read nowhere — no class, no attribute, every link still clickable. It now disables every item and the expand button and takes them out of the tab order (BCR-API-01)
- **fixed** `firstItemNegativeMargin`, which hung off `.kbq-overflow-items` — a wrapper only `wrapMode="auto"` renders, leaving the offset inert in `wrap` and `none`. It is carried by `.kbq-breadcrumbs__list` too now, and deliberately not by the host: a negative margin on a flex item is subtracted from the space the trail is given to fit into, so `auto` mode stops collapsing and spills out of its container (BCR-CSS-01)
- **added** `aria-label` as a real input and sourced the landmark name from `kbqInjectA11yLocaleConfiguration()` through a new `breadcrumbs` a11y locale key; the element form also gets `role="navigation"` (BCR-A11Y-01)
- **added** `role="list"`/`role="listitem"` around the trail and `aria-hidden` on the separators (BCR-A11Y-02)
- **added** `current`, which drives `aria-current="page"` for a trail that does not end at the current page (BCR-API-02)
- **fixed** <kbd>PageUp</kbd>/<kbd>PageDown</kbd>, which the roving-focus primitive mapped to first/last and `preventDefault()`-ed, so a trail near the top of a long page ate the paging keys (BCR-A11Y-05)
- **fixed** arrow-key order: registration now sorts by DOM position instead of appending, so toggling an item's `focusable` no longer moves it to the end (BCR-LIFE-01)
- **fixed** the collapsed-items dropdown, which rendered a blank row for an item with a custom template (BCR-BUG-01)
- **fixed** `<a kbq-button kbqBreadcrumb>`, which overwrote a consumer's `color`/`kbqStyle` and opted the button out of a surrounding `KbqButtonGroupRoot` (BCR-BUG-02)
- **updated** the direction handling: `dir` is emitted only when bound and otherwise follows the ambient CDK `Directionality`; `dir`/`loop` are forwarded through `hostDirectives` (BCR-A11Y-04)
- **updated** `Direction` to come from `@angular/cdk/bidi` rather than being redeclared, and renamed `Orientation` to `KbqRovingFocusOrientation` — the entry point was about to publish two bare names, one of which collides with the CDK type it duplicates (BCR-API-04)
- **updated** documentation for the breadcrumbs component: truncation modes are presented as the CSS recipes they actually are, and `max`, `disabled` and a new Accessibility section are documented in both languages (BCR-DOC-01)

**The collapsed-items dropdown nests an anchor inside an anchor.** BCR-BUG-01 makes the dropdown replay `item.customTemplateRef()`, and the row is itself `<a kbq-dropdown-item [routerLink]>`, so a custom view containing its own link renders as:

```html
<a kbq-dropdown-item href="/Library" tabindex="0">
  <div class="kbq-dropdown-item__text">
    <a kbq-button kbqBreadcrumb href="/Library" tabindex="-1">Library</a>
```

axe stays silent — neither element carries an explicit widget role and the inner one is `tabindex="-1"`, so `nested-interactive` does not trip — but `<a>` may not contain interactive content, and the two hrefs can disagree because the row's comes from `KbqBreadcrumbItem.routerLink` while the inner one comes from the consumer's template. Only `itemsExcludingEdges` can collapse, so head and tail are safe. The alternatives are to build the row from `text`/`routerLink` and drop the custom view, or to make the row a non-link container, which trades the nesting for a focusable row that does nothing on <kbd>Enter</kbd>. Worth a decision rather than leaving it implicit.

**What the review deliberately did not convert.** Imperative fields with no reactive consumer are left alone: `RdxRovingFocusGroupDirective.orientation` stays a decorator input because `KbqBreadcrumbs` assigns it directly once from its own constructor, and `itemIds` stays a `WeakMap` because nothing reads it reactively.

**`registerItem` now ignores an element outside the group's host subtree.** A `TemplateRef` declared inside the group keeps resolving the injected group through its declaration-site injector even when the overflow dropdown instantiates it a second time via `ngTemplateOutlet`; without the containment check the trail grew a phantom, unfocusable duplicate. The assumption is that a legitimate item is always a DOM descendant of its group — covered by `should not register a hidden item a second time when its dropdown copy renders`.

**Two DOM details that only matter to a stylesheet reaching inside.** The default separator is a plain `<i class="kbq-breadcrumb__separator">` instead of an empty `KbqIcon`, so it no longer carries `kbq-icon`/`kbq-contrast-fade`; and `.kbq-breadcrumbs__list` is a real flex box rather than `display: contents`, because several browser/AT stacks drop `role="list"` from a box-less element.


NikGurev:
## BCR-CSS-01 deviates from the fix this PR originally carried

Flagging this explicitly, because the finding is still called BCR-CSS-01 but the fix under it is no longer the one described when the branch was first pushed.

**Originally:** the offset hung off `.kbq-overflow-items`, a wrapper only `wrapMode="auto"` renders, so `firstItemNegativeMargin` silently did nothing in `wrap` and `none`. The first fix moved it onto the host, which reaches all three modes in one line.

**That regressed the component.** The host is a flex item, and a negative margin on a flex item is subtracted from the space the flex algorithm hands it. The trail stopped being squeezed by a constrained parent, `KbqOverflowItems` went on measuring a box that was never narrowed, concluded that every item fits, and the trail spilled out of its container uncut with the expand button hidden. Reproduced in `breadcrumbs-with-auto-wrap-adaptive`, measured in the browser:

| Container | `nav` width | Expand button |
| --- | --- | --- |
| 828px | 802px | hidden |
| 480px | **519.8px** | hidden |
| 320px | **519.8px** | hidden |
| 137px | **519.8px** | hidden |

The trail rendered at its full natural width regardless of the container.

**The fix now in the branch** (`fix(breadcrumbs): keep the first-item offset off the host`) puts the offset on `.kbq-breadcrumbs__list` as well as on `.kbq-overflow-items`. That wrapper only exists because the review added `role="list"` under BCR-A11Y-02, so the original objection — that `auto` was the only mode with an inner wrapper to hang the offset on — no longer holds. Both modes get the offset and the box the trail is measured against is left alone.

After the fix: 480px → `Main / ⋯ / Edit / Test`, 320px → the same, 137px → `⋯ / Test`, nothing clipped. `wrap` mode still gets the offset: `-6px` on `.kbq-breadcrumbs__list`, first item moving from 0 to −6px.

### Why the e2e added with BCR-CSS-01 did not catch it

Two independent reasons, both worth knowing before trusting those rows again:

- they cap the trail with its own `max-width`, which clamps it whatever the margin does — in that shape the bug does not reproduce at all (300px stays 300px);
- they set `max: 4`, so collapsing is driven by counting rather than by measuring, and they collapse either way.

A unit test cannot cover this: jsdom has no layout and `ResizeObserver` is stubbed.

So `E2eBreadcrumbsStateAndStyle` gains a row in the shape a consumer actually writes — width coming from a constrained parent (`display: flex; width: 300px; overflow: hidden`) and `[max]="null"`, so collapsing is decided by measurement alone. Verified against the regenerated baseline: the row collapses, and its left ink edge is at x=21, matching the `auto` row that carries the offset, against x≈31 for the rows without it.

The `#### Breadcrumbs` section of the migration guide claimed "It sits on the host now" in both languages. Corrected, and it now records why the host is the wrong place rather than just where the offset ended up.

lskramarov (approved):

PR #2064, opened 2026-09-17
fix(filter-bar): preserve pipe label width (#DS-5588)

rmnturov:
## Summary

Исправлена обрезка названия и значения пайпа в Filter Bar. Название и значение теперь занимают независимые сжимаемые grid-треки, а максимальная ширина применяется ко всему пайпу вместе с кнопкой очистки.

Для тултипа добавлена локальная директива `KbqPipeTitle`. Она проверяет фактическую обрезку текстовых частей пайпа. Общая `KbqTitleDirective` не изменялась, поэтому исправление не влияет на другие компоненты.

## List of notable changes:

- **updated** раскладка пайпа: название и значение обрезаются независимо, а короткая часть сохраняет естественную ширину;
- **added** локальная директива `KbqPipeTitle`, потому что общая `KbqTitleDirective` измеряет весь триггер и не определяет обрезку отдельного grid-трека;
- **updated** пайпы Select, MultiSelect, TreeSelect, TreeSelect Multiple, Text, Date и Datetime для использования локальной логики тултипа;
- **fixed** отображение одного выбранного значения в MultiTreeSelect;
- **covered** короткие, средние и длинные значения, Datetime, вложенный пользовательский триггер и тултипы unit- и browser-тестами;
- **added** dev-пример с основными комбинациями длины названия и значения.

- независимая обрезка названия и значения;
- отсутствие лишней обрезки короткой части пайпа;
- ограничение ширины всего пайпа вместе с кнопкой очистки;
- появление тултипа только при фактической обрезке названия или значения;
- локальность `KbqPipeTitle`: поведение общей `KbqTitleDirective` не изменено.

Задача: DS-5588.


packages/components/filter-bar/pipes/pipe-multi-select.html
```diff
@@ -24,9 +24,14 @@
         [kbqPipeState]="data"
         [class]="{ 'kbq-active': select.panelOpen }"
     >
-        <span #kbqTitleText class="kbq-pipe__name">{{ data.name }}</span>
+        <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
```
lskramarov: **Two labels under 20 characters are now hard-clipped, with no ellipsis and no tooltip.**

`kbqPipeMinWidth` pins every span shorter than 20 characters at `min-width: fit-content`. Meanwhile the label area of a capped pipe shrank from 320px to 266px with a clear button (about 296px without one), because the cap now includes the padding and the button.

I rendered a cleanable pipe `Ответственный отдел` (19) + `Отдел безопасности` (18) in Chromium, merge base vs this branch, same fixture:
- merge base: 295px of text in a 320px label, fully visible;
- this branch: `.kbq-button-text` is 266px wide for 295px of content, so the value is cut off mid-letter ("Отдел безопасн"). Neither span overflows its own box, so no ellipsis is drawn, and `kbq-title` shows no tooltip because it compares each span against the wrapper.

Same result for the select, multi-select, text, readonly and multi-tree-select pipes. This PR also adds the pin to the multi-select, date and datetime spans, which previously shrank with an ellipsis.

packages/components/filter-bar/pipes/pipe-date.html
```diff
@@ -15,8 +15,8 @@
     [kbqPopoverPlacementPriority]="[placements.BottomLeft, placements.TopLeft]"
     (kbqPopoverVisibleChange)="isListMode = true"
 >
-    <span #kbqTitleText class="kbq-pipe__name">{{ data.name }}</span>
-    <span #kbqTitleText class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
```
lskramarov: **`kbqPipeMinWidth` keeps a stale `fit-content` on date/datetime values when no `[filter]` is bound.**

The directive re-measures only after the first render and when `filterBar.filter()` changes. `KbqFilterBar` updates `filter` on `onChangePipe` only when a filter is bound (`if (current)` in `filter-bar.ts`). Without `[filter]`, as in the `filter-bar-date-limits` docs example, the empty value is measured once (`fit-content`) and never again.

To reproduce: take a cleanable date pipe `Дата обнаружения` and pick `За последние три календарных месяца`. The value keeps `min-width: fit-content`, and `.kbq-button-text` is 266px wide for 407px of content, so the value is clipped with no ellipsis. The same pipe in a bar with `[(filter)]` re-measures to `unset` and gets an ellipsis. At the merge base both bars show an ellipsis. `pipe-datetime.html` has the same span. The single-value span in `pipe-multi-select.html` should go stale the same way when new `pipeTemplates` relabel an option while it stays the only one selected.

packages/components/filter-bar/pipes/pipe-datetime.html
```diff
@@ -15,8 +15,8 @@
     [kbqPopoverPlacementPriority]="[placements.BottomLeft, placements.TopLeft]"
     (kbqPopoverVisibleChange)="isListMode = true"
 >
-    <span #kbqTitleText class="kbq-pipe__name">{{ data.name }}</span>
-    <span #kbqTitleText class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
```
lskramarov: **Ordinary datetime ranges now lose their end, and hovering doesn't show it.**

With the cap on the whole pipe, `12 дек 2025, 10:10 — 9 дек 2025, 10:10` (255px) gets 237px on a plain pipe and 198px on a cleanable one. At the merge base it got 255 and 253 of 255 (pipes #30/#31 in `E2eFilterBarStates`). Hovering either pipe opens no `kbq-title` tooltip: `isElementOverflown` compares each `#kbqTitleText` span's `scrollWidth` (255) with the wrapper width (296/266). A value truncated because of its sibling never counts as overflown, so users can only read the end of the range by opening the pipe.

lskramarov (commented):
I wrote these comments at 0b91663 and re-checked them at 56d85e5. That commit updates the `01-light`/`01-dark` baselines, which fixes the stale-baseline problem I had found, and merging `main` doesn't change the code involved.

**How I checked:** I served the e2e app at the merge base `b8e2a4889` and at this branch, then rendered the same fixtures in Chromium, plus some edge cases, and compared layout, screenshots and tooltips. I also ran the new unit tests against both commits. The pixel figures in the inline comments come from those renders.

**Summary**
- The targeted case, a short name next to a long multi-select value, is fixed.
- Capping the whole pipe takes 24–54px from every capped label. Two short labels are now cut off with no ellipsis, and ordinary datetime ranges truncate with no tooltip.
- The cap also limits the input pipe and custom pipes with a wrapped trigger.
- `kbqPipeMinWidth` on date/datetime values goes stale in bars without `[filter]`.
- The new unit tests can't fail on the width behaviour, and the e2e tests check the inline style rather than whether the short label is truncated.

The description says `kbqPipeMinWidth` is now applied to Select and TreeSelect, but those templates already had it; the date and datetime template changes aren't mentioned.

_Generated with Claude Code_

packages/components/filter-bar/pipes/base-pipe.scss
```diff
@@ -5,14 +5,24 @@
     display: flex;
     flex-direction: row;
 
+    max-width: var(--kbq-filter-bar-pipe-max-width, 320px);
```
lskramarov, line 8: **The host cap now also clamps the input pipe.**

`kbq-pipe-input` has the `kbq-pipe` host class and loads `base-pipe.scss`, so this `max-width` overrides the fixed `width` set in `pipe-input.scss`. The token comment calls that value "a width rather than a max-width". In a render, a bar with `--kbq-filter-bar-pipe-input-width: 400px` gets a 400px input at the merge base and a 320px one on this branch. Lowering `--kbq-filter-bar-pipe-max-width` below 240px now also shrinks the default input. Adding `max-width: none` in `pipe-input.scss`, or keeping the cap off the input pipe, would preserve the documented token.

packages/components/filter-bar/pipes/base-pipe.scss
```diff
@@ -5,14 +5,24 @@
...
     max-height: 32px;
+    min-width: 0;
+
+    & > .kbq-button,
```
lskramarov: **A custom pipe whose trigger isn't a direct child stops truncating.**

The old cap was a fixed-length descendant rule, so it limited a `kbq-button` at any depth. Now only direct `.kbq-button` / `.kbq-select` / `.kbq-tree-select` children get `min-width: 0`, and the remaining `max-width: 100%` doesn't limit their min-content width. I wrapped the trigger of `E2eFilterBarPipeTruncation` in a `<div>`, as a consumer pipe built on the public `KbqBasePipe` may do. At the merge base the pipe stays 344px and the value gets an ellipsis. On this branch the host is 320px and the button overflows it by 208px with nothing truncated. In the same render, replacing the class list with `& > :not(.kbq-pipe-button) { min-width: 0; }` restored truncation.

packages/components/filter-bar/pipes/pipe-tree-select.spec.ts
```diff
@@ -171,6 +171,33 @@ describe('KbqPipeTreeSelectComponent', () => {
...
+            fixture = TestBed.createComponent(TestComponent);
+            filterBarDebugElement = fixture.debugElement.query(By.directive(KbqFilterBar));
+            fixture.componentInstance.pipeTemplates = [
+                {
+                    ...fixture.componentInstance.pipeTemplates[0],
+                    values: kbqBuildTree({ ...DEV_DATA_OBJECT, [longValueName]: 'long' }, 0)
+                }
+            ];
+            fixture.componentInstance.activeFilter = createFilter([
+                createPipe({ name: 'Тип', value: { name: longValueName, value: 'long' } })
```
lskramarov: **This test never renders a value, and it passes without the fix.**

The tree-select pipe matches raw node values: `kbqBuildTree` stores `'long'` and `FlatTreeControl` compares with `===`. The pipe here gets the object `{ name: longValueName, value: 'long' }`, so nothing is selected. Probing the DOM in this test finds an empty `.kbq-pipe__value` with `kbq-pipe__value_empty`. `pipe-tree-select.html` isn't touched by the PR and already had `kbqPipeMinWidth` on both spans, so the spec passes against the merge base `b8e2a4889`. `value: 'long'` would at least select the node, but the assertions still couldn't see the threshold (see the jsdom comment).

packages/components/filter-bar/pipes/pipe-multi-select.spec.ts
```diff
@@ -198,6 +198,37 @@ describe('KbqPipeMultiSelectComponent', () => {
...
+                createPipe({ name: 'Тип', value: [longValue] })
+            ]);
+
+            fixture.detectChanges();
+            flush();
+            fixture.detectChanges();
+
+            const name = fixture.nativeElement.querySelector('.kbq-pipe__name') as HTMLElement;
+
+            expect(name.style.minWidth).toBe('fit-content');
```
lskramarov: **`style.minWidth === 'fit-content'` can't tell short text from long text under Jest.**

`KbqPipeMinWidth.textLength` reads `innerText`, which jsdom 26.1.0 doesn't implement. `innerText?.length || 0` is therefore always 0, and every span gets `fit-content`. I ran this setup with a 48-character name and it also read `fit-content`. The new assertions only prove that the directive is attached. None of the three tests can check that the long part gets `unset` (the same applies to `pipe-multi-tree-select.spec.ts:219` and `pipe-tree-select.spec.ts:196`). Only the Playwright spec covers the threshold.

packages/components/filter-bar/e2e.playwright-spec.ts
```diff
@@ -54,6 +54,54 @@ test.describe('KbqFilterBarModule', () => {
...
+            await page.goto('/E2eFilterBarMultiSelectPipeTruncation');
+
+            const pipe = getComponent(page).locator('.kbq-pipe').first();
+            const name = pipe.locator('.kbq-pipe__name');
+            const value = pipe.locator('.kbq-pipe__value');
+
+            await expect(name).toHaveText('Тип');
+            await expect(value).toHaveText('Исходный код и развернутое приложение из внешнего репозитория');
+            expect(await name.evaluate((element) => (element as HTMLElement).style.minWidth)).toBe('fit-content');
+            expect((await getWidths(name)).client).toBeGreaterThan(0);
```
lskramarov: **The e2e tests never check that the short part is whole.**

`client > 0` also passes on the squeezed layout from before the fix: at the merge base `Тип` measured 16 of its 25px. The bug is caught only by the inline `style.minWidth` string (line 73) and the hardcoded `321` (line 80), which duplicates the token default. So a stylesheet rule that squeezes the label while the directive still writes `fit-content` stays green. A CSS-only fix without the inline style fails, and raising the token default fails both tests. Asserting `scrollWidth <= clientWidth` on the short part (next to the existing `>` check on the long one), and reading the cap from computed style, would test what users see. Date and datetime get no truncation case, although their templates change here.

packages/components/filter-bar/pipes/pipe-multi-tree-select.html
```diff
@@ -42,7 +42,12 @@
         <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
 
         @if (select.selected?.length === 1) {
-            <span #kbqTitleText class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!select.triggerValue">
+            <span
+                #kbqTitleText
+                kbqPipeMinWidth
+                class="kbq-pipe__value"
+                [class.kbq-pipe__value_empty]="!select.triggerValue"
```
lskramarov: **Pre-existing, but in the span this PR edits: `select.triggerValue` is always empty for the multi tree select.**

In multiple mode `KbqTreeSelect.selected` is an array. The pipe's patched `treeControl.getViewValue` returns `node.name`, which is `undefined` for an array. This span therefore always gets `kbq-pipe__value_empty` (seen in the DOM with one leaf selected), and the tooltip line at 98 is blank: hovering a truncated multi-tree-select pipe with `Отдел безопасности` selected shows only the name. Deriving both from `select.selected`, as the span's content already does, would fix it.

packages/components-dev/filter-bar/pipe-truncation-example.ts
```diff
@@ -0,0 +1,224 @@
...
+            disabled: false,
+            changed: false,
+            saved: false,
+            pipes: [
+                {
+                    id,
+                    name,
+                    type,
+                    value,
+                    cleanable: hasLongText,
```
lskramarov: **The dev matrix never shows the combination this change breaks.**

`cleanable` comes from the text length (`hasLongText`). As a result, every "Короткое название + короткое значение" case renders without a clear button, while every other filled case has one. Each row varies two things at once, and short labels next to a clear button never appear, although that is the layout that now clips (see `pipe-multi-select.html`). `Тип`/`Файл` would be too short to show it anyway. A pair such as `Ответственный отдел` + `Отдел безопасности` with `cleanable: true` would show it. Also, `createTextPipeTruncationCases` duplicates `createPipeTruncationCase` for the Text group.

packages/components/filter-bar/pipes/pipe-date.html
```diff
@@ -15,8 +15,8 @@
     [kbqPopoverPlacementPriority]="[placements.BottomLeft, placements.TopLeft]"
     (kbqPopoverVisibleChange)="isListMode = true"
 >
-    <span #kbqTitleText class="kbq-pipe__name">{{ data.name }}</span>
-    <span #kbqTitleText class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
```
lskramarov: **Which part keeps its width still depends on each template opting in, and on a character count.**

Before this PR, 4 of the 8 built-in templates lacked `kbqPipeMinWidth`: date, datetime and multi-select since they were added, plus the multi-tree-select value. Custom pipes have to import it too. `innerText.length < 20` knows neither pixel widths nor the sibling's width, and it only re-measures on `filter()` changes. The clipping and staleness described in the other comments come from this.

A layout rule in `base-pipe.scss` would need neither. In a quick render, `.kbq-button-text { display: grid; grid-auto-flow: column; grid-auto-columns: minmax(0, max-content); }` kept `Тип` whole beside the long value, and truncated both halves of the 19+18 pair with an ellipsis instead of clipping. It splits space equally rather than using the 20-character rule, so a short part stays whole only while it fits in half the space. That makes it a design trade-off rather than a drop-in replacement, but it would remove both the measuring and the opt-in.

packages/components/filter-bar/pipes/base-pipe.scss
```diff
@@ -5,14 +5,24 @@
...
 
     & .kbq-select .kbq-select__trigger {
         height: unset;
     }
 
     .kbq-button-wrapper {
-        max-width: var(--kbq-filter-bar-pipe-max-width, 320px);
+        max-width: 100%;
+        min-width: 0;
     }
```
lskramarov: **Three of these declarations change nothing.**

`.kbq-button` already has `max-width: 100%` (`%kbq-button-clamped-to-container`), `.kbq-select` / `.kbq-tree-select` are `width: 100%` (`%kbq-select-base`), and `.kbq-button-wrapper` already gets `min-width: 0` from `_button-base.scss`. I removed the `max-width: 100%` here and the whole `.kbq-button-wrapper` block and confirmed in the CSSOM that the rules were gone. `E2eFilterBarStates`, `E2eFilterBarPipeTruncation`, the new fixture and several edge cases all rendered pixel-identical. Only `min-width: 0` on the direct children has an effect.

```suggestion
        min-width: 0;
    }

    & .kbq-select .kbq-select__trigger {
        height: unset;
    }
```

packages/components/filter-bar/e2e.playwright-spec.ts
```diff
@@ -54,6 +54,54 @@ test.describe('KbqFilterBarModule', () => {
         });
     });
 
+    test.describe('E2eFilterBarMultiSelectPipeTruncation', () => {
+        const getComponent = (page: Page) => page.getByTestId('e2eFilterBarMultiSelectPipeTruncation');
+
+        /** Width the text would need if nothing clipped it, versus the width it actually got. */
+        const getWidths = (locator: Locator) =>
```
lskramarov: **This helper and the two tests duplicate code already in this file.**

This `getWidths` is byte-identical to the one at line 23, and the two new tests mirror each other (same bounds block, same `321`). With `getNameWidths` from `main` doing the same read, the file now has three such helpers. Moving the helper up to the outer `describe` and looping over `{ index, short, long }` would keep future fixes in one place. `E2eFilterBarMultiSelectPipeTruncation` also mirrors `E2eFilterBarPipeTruncation` (same template minus the `kbq-filters` line). Its pipes could live in the existing fixture instead of a second component, route and `describe`, though the existing tests use `.first()`, so their indices would need adjusting.

packages/components/filter-bar/pipes/pipe-datetime.html
```diff
@@ -15,8 +15,8 @@
     [kbqPopoverPlacementPriority]="[placements.BottomLeft, placements.TopLeft]"
     (kbqPopoverVisibleChange)="isListMode = true"
 >
-    <span #kbqTitleText class="kbq-pipe__name">{{ data.name }}</span>
-    <span #kbqTitleText class="kbq-pipe__value" [class.kbq-pipe__value_empty]="!formattedValue">
+    <span #kbqTitleText kbqPipeMinWidth class="kbq-pipe__name">{{ data.name }}</span>
```
lskramarov: **Each added `kbqPipeMinWidth` re-measures on every filter change, even text that can't change.**

On every `filterBar.filter()` emission, each instance schedules its own `setTimeout`, reads `innerText` (which forces layout when the DOM is dirty) and calls `markForCheck()` unconditionally. Under zone.js that usually adds one change-detection pass per timer. This PR adds up to 7 instances per bar: the multi-select name and value, the multi-tree-select value, and the date and datetime names and values. `data.name` never changes while a pipe exists, yet every option toggle re-measures it. Measuring names once, or batching reads and writes (e.g. `afterRenderEffect` with `earlyRead`/`write`) and skipping `markForCheck` when nothing changed, would avoid this. The CSS approach from the `pipe-date.html` comment would remove it entirely.

packages/components/filter-bar/pipes/pipe-multi-select.html
rmnturov: Исправил в `81dce2f`: убрал порог по количеству символов и `kbqPipeMinWidth`. Название и значение теперь размещаются в отдельных shrinkable `max-content` grid-треках, поэтому короткая часть сохраняет ширину, а две более длинные части обрезаются каждая со своим ellipsis. Добавил Playwright-кейс `Ответственный отдел` + `Отдел безопасности`: обе части действительно переполняют свои треки, а тултип показывает оба полных значения.

packages/components/filter-bar/pipes/base-pipe.scss
rmnturov: Учёл вложенный custom trigger: непосредственный wrapper теперь получает `min-width: 0` через `& > :not(.kbq-pipe-button)`, а button/select/tree-select ограничиваются независимо от глубины. Grid-правило для `.kbq-button-text` также работает по descendant-селектору. Целевой Playwright-регресс обрезания проходит, но отдельный fixture с дополнительным wrapper в коммит не добавлен.

packages/components/filter-bar/pipes/pipe-date.html
rmnturov: Убрал `kbqPipeMinWidth` из date/datetime и остальных встроенных pipe. Раскладка теперь определяется CSS и больше не зависит от обновления `[filter]`, поэтому устаревшего `fit-content` не остаётся. Поведение обрезанного datetime проверяется в Playwright на обычном `E2eFilterBarStates`.

packages/components/filter-bar/pipes/pipe-multi-tree-select.html
rmnturov: Исправил заодно: при одном выбранном узле multi-tree-select теперь берёт текст из `select.selected[0].name` и для значения, и для тултипа, без ложного `kbq-pipe__value_empty`. Добавил unit-тест, который проверяет текст выбранного leaf и отсутствие empty-класса.

packages/components/filter-bar/pipes/pipe-datetime.html
rmnturov: Убрал `kbqPipeMinWidth` из встроенных pipe, поэтому таймеры, повторное чтение `innerText` и `markForCheck` на каждое изменение filter больше не создаются. Локальный `KbqPipeTitle` реагирует только на resize/content changes и непосредственно перед focus/hover; его состояния покрыты unit-тестами.

packages/components/filter-bar/pipes/pipe-multi-select.spec.ts
rmnturov: Удалил проверки inline `minWidth` и саму зависимость встроенных pipe от `kbqPipeMinWidth`. Реальная раскладка теперь проверяется в Chromium через `scrollWidth`/`clientWidth`, а unit-тест новой локальной tooltip-директивы использует явно заданную геометрию и проверяет оба состояния: текст обрезан и текст помещается.

packages/components/filter-bar/pipes/pipe-tree-select.spec.ts
rmnturov: Согласен, этот Jest-тест не проверял реальное значение и проходил без исправления. Удалил его вместе с `kbqPipeMinWidth`; проверку геометрии оставил в Playwright, где доступны настоящие `scrollWidth` и `clientWidth`.

packages/components/filter-bar/pipes/pipe-datetime.html
rmnturov: Добавил локальную для Filter Bar директиву `KbqPipeTitle`: она сравнивает `scrollWidth` и `clientWidth` каждого текстового grid-трека, а не текста со всей шириной триггера. Playwright-тест воспроизводит именно этот случай — значение уже шире своего трека, но ещё уже кнопки — и проверяет, что по hover появляется полный datetime-диапазон.

packages/components/filter-bar/pipes/base-pipe.scss
rmnturov: Удалил неработающие `max-width: 100%` и блок `.kbq-button-wrapper`. Оставил только влияющие на shrink правила `min-width: 0`, расширив их на wrapper и вложенные triggers для custom pipe. После упрощения прошли browser-сценарии обрезания, unit-тесты и сборка компонентов.

packages/components/filter-bar/pipes/pipe-date.html
rmnturov: Перешёл на CSS-подход: `.kbq-button-text` использует `grid-auto-columns: minmax(0, max-content)`. `kbqPipeMinWidth` удалён из всех встроенных шаблонов и custom-pipe-примера, поэтому больше нет opt-in по каждому template, порога в 20 символов и зависимости от обновления filter. Playwright проверяет обе комбинации short/long и случай, когда обе части обрезаны.

packages/components/filter-bar/e2e.playwright-spec.ts
rmnturov: Обновил e2e-проверки: для короткой части теперь утверждается `scrollWidth <= clientWidth`, для длинной — `scrollWidth > clientWidth` и `text-overflow: ellipsis`. Максимум читается из computed style вместо `321`. Дополнительно добавлен datetime-кейс с проверкой фактического переполнения и полного тултипа; отдельного аналогичного date-теста пока нет.

packages/components/filter-bar/pipes/base-pipe.scss
rmnturov: Добавил `max-width: none` для `.kbq-pipe.kbq-pipe__input`, поэтому общий cap pipe больше не переопределяет `--kbq-filter-bar-pipe-input-width`. Сборка компонентов и целевые тесты проходят; отдельного browser assertion на значение `400px` в этом коммите нет.

packages/components-dev/filter-bar/pipe-truncation-example.ts
rmnturov: Добавил в dev-матрицу cleanable-кейс `Ответственный отдел` + `Отдел безопасности` и отвязал `cleanable` от длины текста. Также убрал дублирующий генератор Text-кейсов: теперь все группы строятся через один helper. Этот же medium-сценарий покрыт Playwright-проверкой обрезания и тултипа.

packages/components/filter-bar/e2e.playwright-spec.ts
rmnturov: Вынес `getWidths` на уровень внешнего `describe` и убрал его дубликаты; максимум также читается общим helper из computed style. Отдельный multi-select fixture оставил: в нём три стабильных состояния для проверки short/long/medium и тултипа, тогда как существующий fixture проверяет общий overflow и saved-filter name.

packages/components-dev/filter-bar/pipe-truncation-example.html
NikGurev, line 1: это будет только локальный пример в дев? Не будем его в доке показывать? Предлагаю убрать в docs-examples/components/filter-bar

и показывать в разделе "примеры"

NikGurev (approved):

packages/components-dev/filter-bar/pipe-truncation-example.html
rmnturov: Такой взгляд: не надо его в доке показывать. Я его когда-то сделал для тестирования, чтобы проверить логику обрезания длинного текста.

lskramarov:
закрыто в пользу https://github.com/koobiq/angular-components/pull/2073 - изменения действительно нужно делать в kbq-title

PR #2067, opened 2026-09-17
feat(docs): open a live example on its own page (#DS-4873)

lskramarov:
## Summary

Every live example in the documentation can now be opened on a page of its own. The example viewer gets an "Open in new tab" link to `/examples/<id>`, a page that shows the example alone, in the reader's saved theme and examples language. The two ad-hoc pages behind the popover and select layering demos (`examples/popover` and `examples/select`, marked `todo DS-4873`) are replaced by regular examples framed through that page.

## List of notable changes:

- **added** an "Open in new tab" link to the live example viewer and the `/examples/<id>` page (`components/example-page`): no site navigation, the same frame as in the viewer, titled after the example, `noindex`. Unknown ids go to 404; relative links inside an example keep it on its page.
- **added** the generated `@koobiq/docs-examples/loader` entry point with `loadExampleComponent(id)`, kept apart from `EXAMPLE_COMPONENTS` so the pages that show live examples do not load it. `components-dev/ssr` uses it as well.
- **moved** the popover and select layering demos from docs-app components into the `popover-scrolling-and-layering-page` and `select-scrolling-and-layering-page` examples; their iframes point at the example page and load lazily.
- **fixed** the saved theme and interface/examples languages are applied as the app starts (`docsProvidePreferences`) instead of as a side effect of the navbar and footer rendering. The `/examples/*` pages used to get them only because the site navigation rendered once before being removed, which also created the routed page twice.
- **updated** the app shell keeps a single router outlet for both layouts.
- **updated** the frame of a live example is defined in `styles/_examples.scss`, and per-example modifiers set custom properties.
- **fixed** a saved language index past the end of the locale list no longer breaks the footer, and blocked `localStorage` no longer throws.

- **Visual change:** the `top-bar-overflow` (32px padding) and `ag-grid-and-actions-panel` (no padding) modifiers have had no effect since 1ba2f573e, because the viewer's frame rule won the cascade. They now apply as authored.
- **Generated files:** `packages/docs-examples/example-module.ts` no longer contains the loader; the new `packages/docs-examples/loader/index.ts` is generated by `build:docs-examples-module` (AGENTS.md updated).
- The old `/examples/popover` and `/examples/select` URLs now lead to 404. They were only used as iframe sources and were `noindex`.
- Known and left as is: the interface language switch in the footer goes stale after back/forward navigation (the logic moved from the footer unchanged).

Checked with `unit:koobiq-docs`, `unit:tools`, `e2e:docs` (new tests for the link, the standalone page with the saved theme and language and without site navigation, the 404 redirects and the framed layering demos), `ssr:build` and the linters.


rmnturov:
Все норм по дизайну

artembelik (approved):

NikGurev (approved):

PR #2068, opened 2026-09-17
feat(inline-edit): async save with progress and server error states (#DS-4384)

NikGurev:
## Summary

Обработка асинхронных запросов и ошибок

- [x] Примеры для select


packages/components/inline-edit/inline-edit.ts
```diff
@@ -343,14 +425,39 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+        if (this.saveStatus() !== 'error' || !saveHandler) return;
+
+        this.startSave(saveHandler);
+    }
+
+    /** Discards the value the server rejected, restoring the last saved one and clearing the failed state. */
+    rollback(): void {
+        if (this.saveStatus() !== 'error') return;
+
+        this.setValue(this.lastSavedValue);
```
lskramarov, line 470: **`rollback()` wipes the value when edit mode has no `kbq-form-field`.**

`lastSavedValue` is only ever assigned inside `onAttach()`'s `setTimeout`, *after* `if (!formFieldRef) return;` (line 491). Edit mode without a form field is a supported shape — `inline-edit-custom-handler-example` projects `<div kbqInlineEditEditMode>` with a `kbq-file-upload` and drives the value through `getValueHandler`/`setValueHandler`. Add `[saveHandler]` to that shape, let a save fail, and `rollback()` calls `setValue(undefined)`, which goes straight to `setValueHandler(undefined)` and destroys the value instead of restoring it.

`initialValue` has the same hole for `cancel()`, so both captures belong before the `formFieldRef` early return (or at save time).

packages/components/inline-edit/inline-edit.ts
```diff
@@ -411,10 +520,87 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+    }
+
+    private onSaveSucceeded(): void {
+        this.lastSavedValue = this.submittedValue;
+        this.saveStatusSource.set('idle');
+        this.saved.emit();
+    }
+
+    private onSaveFailed(error: unknown): void {
+        this.saveStatusSource.set('error');
```
lskramarov, line 635: **`cancel()` never clears `'error'`, so the row can get stuck in the failed look with nothing left to recover it.**

`saveStatus` only leaves `'error'` through `onSaveSucceeded()` or `rollback()`. Sequence: save fails → toast opens → user picks *Edit* (which closes the toast and, in `inline-edit-save-error-example`, deletes the `failedSaves` entry) → user presses Escape. `cancel()` restores `initialValue`, which `onAttach` pinned to the rejected value, and leaves `saveStatus` at `'error'`.

Result: the row keeps the error background and `aria-invalid`, the notification is gone, and no one holds the context needed to call `rollback()`/`retrySave()`. Either `cancel()` should clear the failed state (or roll back to `lastSavedValue`), or the docs should say the error survives a cancel.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -391,6 +494,12 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
 
             this.initialValue = this.getValue();
 
+            // Opening the editor over a value the server rejected must not turn that value into the one
+            // `rollback()` restores.
+            if (this.saveStatus() !== 'error') {
```
lskramarov, line 522: **After `rollback()` in edit mode, Escape silently re-applies the rejected value — and now without any error marking.**

`rollback()` writes `lastSavedValue` into the control and clears the status, but it leaves edit mode open and never refreshes `initialValue`, which this block pinned to the rejected value.

Sequence: save fails → *Edit* opens the editor → *Discard changes* calls `rollback()` → Escape → `cancel()` puts the rejected value back. Because `rollback()` already set the status to `'idle'`, the row now displays the value the server refused with no error indication at all.

`rollback()` should also reset `initialValue` (and probably close the editor).

packages/components/inline-edit/inline-edit.ts
```diff
@@ -343,14 +425,39 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+        this.setValue(this.lastSavedValue);
+        this.saveStatusSource.set('idle');
+    }
+
     /** @docs-private */
     protected onClick(event: Event): void {
-        if (this.disabled() || this.isEditMode() || this.isInteractiveElement(event.target)) return;
+        // Editing is closed while the request is in flight: reopening it would let the user edit a value that is
+        // already on its way to the server.
+        if (this.disabled() || this.isEditMode() || this.isSaving() || this.isInteractiveElement(event.target)) return;
```
lskramarov, line 483: **The in-flight guard lives only on `onClick`, so `toggleMode()` walks straight past it into a concurrent save.**

`toggleMode()` is public, has no `isSaving()` check, and is advertised as one of the three recovery actions in `KbqInlineEditSaveErrorContext`'s doc — both new examples use it that way, and `E2eInlineEditSaveStates.save()` calls it too. Calling it during `'pending'`/`'progress'` produces exactly what this comment says must not happen, and a `commit()` afterwards starts a second `startSave()`: two live subscriptions writing `saveStatusSource` and overwriting the single `submittedValue` field, so the older response can land last and leave `saveStatus` at `'idle'` while a request is still in flight.

The docs claim *“Edit mode cannot be reopened while the request is in flight”* — that only holds for pointer/keyboard entry. Guard `toggleMode()`, or make `startSave()` ignore (or cancel) a re-entrant call.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -411,10 +520,87 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
+            .pipe(
+                // Completes on the result, dropping the progress timer when the request settles first.
+                takeWhile((result) => result.status === 'progress', true),
+                // A result arriving while the progress state is shown is queued behind the minimum display time.
+                concatMap((result) =>
+                    result.status === 'progress'
+                        ? concat(of(result), timer(minimumTimeToDisplayLoading).pipe(ignoreElements()))
+                        : of(result)
+                ),
+                takeUntilDestroyed(this.destroyRef)
```
lskramarov: **Destroying the row silently aborts the save the user was already told about — and the comment above says the opposite.**

Three lines up: *“Nothing cancels the request.”* `takeUntilDestroyed(this.destroyRef)` does: destroying the inline edit unsubscribes the chain and, for an `HttpClient` observable, aborts the XHR.

That needs no user action to reach — a row inside `@for` that re-renders, a virtual-scroll viewport recycling the row, a tab switch. By then edit mode has already closed and the user has been shown the new value as if saved. Neither `saved` nor `saveError` fires, so the application cannot tell it lost the write.

Worth deciding explicitly whether a background save should outlive its row, and fixing the comment either way.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -556,6 +750,9 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
                 break;
             }
             case 'Enter': {
+                // A button in edit mode runs its own action on Enter — saving here as well would fire both.
+                if (event.target instanceof HTMLButtonElement) return;
```
lskramarov: **This drive-by guard pre-empts `canSaveOnEnter`, the input that exists to customise exactly this.**

It runs *before* `canSaveOnEnter(event)`, so a consumer whose edit mode contains a `<button>` (a `kbq-button-toggle` group, a file-upload browse button) loses Enter-to-save with no way to opt back in. `instanceof HTMLButtonElement` also misses `<input type=button>` and `[role=button]` anchors, so the double action it guards against still happens for those.

It is untested and unmentioned in the PR description, and the component's own Save/Cancel buttons were already safe: `save()` returns early on `!isEditMode()`, so the `setTimeout` call after a button click was always a no-op. The real case it fixes is a button in *projected* edit content — which is the case `canSaveOnEnter` should be deciding.

packages/components/inline-edit/inline-edit.scss
```diff
@@ -81,6 +81,27 @@
...
+ * state outlives the request itself. Applied after `_inline-edit-theme()` on purpose: the marking has to survive
+ * hover and active, which would otherwise paint their own background over it.
+ */
+@mixin _inline-edit-save-error-theme {
+    &.kbq-inline-edit_view.kbq-inline-edit_save-error {
+        --kbq-inline-edit-background: var(--kbq-inline-edit-error-background);
+
+        &:not(.kbq-inline-edit_disabled) {
+            &:hover {
+                --kbq-inline-edit-background: var(--kbq-states-background-error-less-hover);
```
lskramarov, line 96: **The error theme mirrors only part of `_inline-edit-theme`'s state block, and both omissions are visible.**

1. `--kbq-mask-background` is never re-pointed. Hovering an error row that has a `[kbqInlineEditMenu]` paints the menu mask in `--kbq-states-background-contrast-less-hover` (grey) on top of the pink background.
2. The base theme's `&:has(.kbq-inline-edit__menu.kbq-active)` is `(0,5,0)` — more specific than the error base rule's `(0,4,0)` — and has no error counterpart. Opening the row's action menu by keyboard (no `:hover` to break the tie at `(0,5,0)` in the error theme's favour) drops the background back to `--kbq-states-background-transparent-active`, so the failed state visually disappears.

Neither shows up in the new baselines because `E2eInlineEditSaveStates` has no menu row.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -154,6 +201,11 @@ export class KbqInlineEditMenu {
         '[class.kbq-inline-edit_disabled]': 'disabled()',
         '[class.kbq-inline-edit_anchor-focused]': 'anchorFocused()',
         '[class.kbq-inline-edit_select]': 'isSingleSelect()',
+        '[class.kbq-inline-edit_save-error]': 'saveStatus() === "error"',
+        // The background save is reported on the row itself, which by then is back in view mode.
+        '[class.kbq-progress]': 'saveStatus() === "progress"',
+        '[attr.aria-busy]': 'isSaving() || null',
```
lskramarov: **For assistive tech both new states are effectively colour-only.**

`aria-invalid` is surfaced on form controls; `<kbq-inline-edit>` carries no `role`, so it is dropped. `aria-busy` on a non-live region announces nothing either. So *“the save failed”* reaches a screen-reader user as the pink background alone (WCAG 1.4.1) with no text explaining it, and *“saving”* as the shimmer alone.

AGENTS.md requires component-provided ARIA text to come from `kbqInjectA11yLocaleConfiguration()`; no strings were added for either state. A visually-hidden live region, or `role=status` / `aria-describedby` on the row, with localised text would cover it.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -294,7 +370,7 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
     protected readonly anchorFocused = signal(false);
     /** @docs-private */
     protected readonly tabIndex = computed(() => {
-        if (this.isEditMode() || this.disabled() || this.hasInteractiveContent()) return -1;
+        if (this.isEditMode() || this.disabled() || this.hasInteractiveContent() || this.isSaving()) return -1;
```
lskramarov: **`isSaving()` covers `'pending'`, so the row leaves the tab order before anything is visible — and stays unreachable for the whole window `aria-busy` exists to announce.**

Every save, however fast, drops `tabindex` to `-1` for the first 101 ms with no visual cue. During `'progress'` the row cannot be focused at all, which is precisely when `aria-busy` is set, so the attribute can never be read.

It is also inconsistent: for rows with interactive content the `kbq-inline-edit__focus-anchor` in the template keeps `tabindex=0` throughout, so those rows stay tabbable while saving. Consider gating on `'progress'` only, or leaving the row focusable and relying on the `onClick` guard, which already handles both states.

tools/check-public-api-any/baseline.json
```diff
@@ -22,7 +22,7 @@
     "flag": 2,
     "form-field": 17,
     "icon": 2,
-    "inline-edit": 7,
+    "inline-edit": 9,
```
lskramarov: **Bumping the ratchet is the one thing this gate is built to catch.**

The tool's own header: *“it fixes the current amount per package and fails when it grows, so that the ongoing cleanup cannot be silently undone by the next feature.”* Its failure message says what to do instead: *“where the type is the consumer's to choose, use a generic with an `any` default rather than a concrete type.”*

Both new `unknown`s are that case — `KbqInlineEditSaveHandler = () => Observable<unknown>` and `KbqInlineEditSaveErrorContext.error: unknown`. As `KbqInlineEditSaveHandler<T = any>` / `KbqInlineEditSaveErrorContext<E = any>` the count stays at 7 *and* consumers lose the `context.error as ExampleSaveError` casts that both new examples need.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -44,10 +46,14 @@ import {
 import { KbqDropdownTrigger } from '@koobiq/components/dropdown';
 import { KbqFormField, KbqLabel } from '@koobiq/components/form-field';
 import { KbqIcon } from '@koobiq/components/icon';
-import { KbqSelect } from '@koobiq/components/select';
+import {
+    delayBeforeDisplayingResultWithoutOptions,
```
lskramarov: **Borrowing select's constants ties inline-edit's timing to an unrelated owner, and the combinator around them is now the fifth copy in this repo.**

`delayBeforeDisplayingResultWithoutOptions` is documented as *“Delay in milliseconds before displaying a result when there are no options”*. Reusing it as the delay before a save-progress indicator means a future change to how select's option loading feels silently retimes inline-edit, and the name tells a reader nothing about why it is here.

The *“wait D before showing loading, then hold it for at least M”* shape is also hand-rolled for the fifth time (the four `select-*-example`s do it with `race`, re-subscribing the request — this version is the first to get single-subscription right). Both the constants and the operator belong in `@koobiq/components/core`, same argument as the `TODO(DS-5420)` already in this file.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -307,6 +383,12 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
 
     private initialValue: unknown;
 
+    /** Value sent by the in-flight `saveHandler` request, which becomes `lastSavedValue` once it succeeds. */
+    private submittedValue: unknown;
+
+    /** Last value the server accepted — what `rollback()` restores after a failed save. */
+    private lastSavedValue: unknown;
```
lskramarov, line 407: **Three `unknown` fields now hold the same value, each captured at a different implicit moment — and every value bug on this PR falls out of that.**

`initialValue` and `lastSavedValue` are set inside `onAttach`'s `setTimeout`, and only past the `formFieldRef` early return; `submittedValue` is set at the top of `startSave`. From that layout: `rollback()` wiping the value when there is no form field, `cancel()` re-applying a rejected value after `rollback()`, and `submittedValue` being clobbered by a re-entrant save.

One record captured at one point — e.g. `{ value, status }` written where the save starts, with `getValue()` read once — retires all three instead of each being patched separately.

packages/components/inline-edit/inline-edit.spec.ts
```diff
@@ -999,6 +999,350 @@ describe('KbqInlineEdit', () => {
...
+            expect(componentInstance.update).toHaveBeenCalledTimes(1);
+        }));
+
+        it('should show the progress state after the delay and keep it for the minimum time', fakeAsync(() => {
+            const fixture = setup(TestWithSaveHandler);
+            const { componentInstance } = fixture;
+            const inlineEditDebugElement = openEditMode(fixture);
+
+            clickSave(fixture);
+            tick(100);
```
lskramarov: **The thresholds are magic numbers, so this test keeps passing while asserting nothing if the constants move.**

`tick(100)` / `tick(1)` / `tick(10)` / `tick(289)` / `tick(1)` only add up because `delayBeforeDisplayingResultWithoutOptions` is 101 and `minimumTimeToDisplayLoading` is 300 — the two values the component imports right at the top.

`select.component.spec.ts` imports them and ticks with them (`tick(delayBeforeDisplayingResultWithoutOptions)`); doing the same here — `tick(delayBeforeDisplayingResultWithoutOptions - 1)` then `tick(1)` — keeps the boundary under test.

packages/components/inline-edit/e2e.ts
```diff
@@ -254,6 +256,87 @@ export class E2eInlineEditActionButtons {
...
+    protected readonly failingRows = viewChildren<KbqInlineEdit>('failingRow');
+
+    /** Never settles, so the progress state stays on screen. */
+    protected readonly pendingSave: KbqInlineEditSaveHandler = () => NEVER;
+
+    protected readonly failingSave: KbqInlineEditSaveHandler = () =>
+        throwError(() => new Error('Couldn’t save the changes'));
+
+    /** Both states belong to view mode, so the rows are saved rather than opened for editing. */
+    protected save(inlineEditList: readonly KbqInlineEdit[]): void {
```
lskramarov: **The fixture reaches the new states through a path no user can take, and skips the row shape where the CSS is actually broken.**

`toggleMode(); commit()` in one synchronous block means `onAttach` never fires: the overlay never opens, `initialValue`/`lastSavedValue` are never captured, and `isInvalid()`/`getValue()` run against content children that were never focused. The screenshots prove the CSS renders, not that a user-driven save reaches these states.

The row selection is also incomplete for what the theme does: no `[kbqInlineEditMenu]` row in either state — exactly where the missing `--kbq-mask-background` / `:has(.kbq-active)` handling in `_inline-edit-save-error-theme` would have surfaced — and no hover/active shot of an error row.

packages/components/inline-edit/inline-edit.ts
```diff
@@ -630,10 +827,15 @@ export class KbqInlineEdit implements KbqConnectedOverlayOriginProvider {
...
-    private saveAndFocusNextInlineEdit(event: Event): void {
+    private saveAndFocusNextInlineEdit(event: KeyboardEvent): void {
         this.save(event);
+
         if (this.isInvalid()) return;
 
+        this.focusNextInlineEdit();
+    }
+
+    private focusNextInlineEdit(): void {
```
lskramarov: **Unrelated refactor churn riding along with the feature.**

`focusNextInlineEdit()` has exactly one caller (the line above), as does `hideValidationTooltip()` at line 720. `showValidationTooltipIfStillInvalid` was renamed to `showValidationTooltipIfNeeded` although the method still checks `isInvalid()` — the old name was the more precise one. And `saveAndFocusNextInlineEdit`'s parameter was narrowed from `Event` to `KeyboardEvent` even though it is only forwarded to `save(event: Event)`.

None of it relates to async save, and each one costs a reviewer a hunk to confirm as a no-op. Worth dropping from this PR or splitting out.

rmnturov:
Вот это или не растягивать на всю ширину, или растягивать так, чтобы 4 кнопки равномерно растягнулись
image

rmnturov:
Открыл на редактирование. ничего не поменял, сохранил. Вроде как можно не отправлять на сервер. Как считаешь?

rmnturov:
Описанию добавьть боковые паддинги, чтобы вровень с текстом ключ-значения было
image

У нас даже пример про это есть)
image

rmnturov:
Кнопки в тосте на такие нужны. Я имею в виду оформление
image

Макет
https://www.figma.com/design/sHcbGrJoOZmFA0DsJQIer9/%F0%9F%9F%A5-koobiq-%C2%B7-components-18?node-id=28453-6542&t=1TiKnetZu4YM6FCh-4

rmnturov:
Иконка рядом с кнопкой не нужна. Если что, то Retry в другую сторону стрелка
image

rmnturov:
image
Текст английский, а кавычки русские (елочки) — не гуд. Вместо кавычек используй strong оформление

rmnturov:
Вместо кавычек используй strong-стиль (text-normal-strong)
Кавычки православные не надо использовать в англ. тексте
image

rmnturov:
Закрытие тоста по крестику должно означать Discard — отмена несохранённых изменений и снятие статуса невалидности
Во всех примерах надо проверить
image

rmnturov:
Вот тут бы невалдиный тег тоже посвечивать, если можно
image

NikGurev:
> Открыл на редактирование. ничего не поменял, сохранил. Вроде как можно не отправлять на сервер. Как считаешь?

Да , тоже показалось, что излишне. поправлю

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Значение запоминается до проверки на наличие формы, поэтому откат возвращает прежнее, а не пустоту.

Заодно починилась отмена: в примере с загрузкой файла Escape стирал выбранный файл ровно по той же причине. Есть тест, он падает без правки.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Оставили как есть, намеренно.** В поле по-прежнему лежит значение, которое сервер отклонил, так что пометка должна держаться.

Описали это в документации на обоих языках: error-state уходит только после удачного сохранения или после отката.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Откат теперь обновляет значение, к которому возвращает отмена, и закрывает редактор. После него Escape уже не может подставить отклонённое.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Переключение режима ничего не делает, пока идёт запрос.

Теперь открыть редактор поверх летящего запроса нельзя ни мышью, ни с клавиатуры, ни из кода, поэтому двух одновременных сохранений не бывает.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено, причём поведение поменяли.** Решили, что обрывать запрос нельзя: это данные пользователя. Удаление строки его больше не отменяет.

События компонента после удаления молчат, а обработчик ошибки продолжает работать — он обычная функция, а не привязка к шаблону. В документацию добавили предложение о том, что объявлять его нужно в корневом инжекторе.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Проверка переехала внутрь `canSaveOnEnter`, так что приложение снова может её переопределить. Список расширили: кроме обычных кнопок учитываются `input` с типом button и элементы с ролью кнопки. Добавили два теста.

Одно уточнение: на кнопке «Отменить» двойное действие всё-таки было. Enter гасил нажатие самой кнопки и сохранял вместо отмены — проверили замером.

packages/components/inline-edit/inline-edit.scss
NikGurev: **Исправлено.** Добавили и состояние с открытым меню, и фон маски.

В e2e появилась строка с меню, снимок с открытым меню и снимок при наведении — раньше эти случаи действительно ничем не закрывались.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Добавили скрытую область, которая проговаривает оба состояния, со строками в пяти локалях.

`aria-invalid` и `aria-busy` убрали. Второй был не просто бесполезен: он стоял на строке, которая содержит это сообщение, и заставлял скринридер придержать его.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено.** Привели к `kbq-dropdown-item`: строка остаётся в обходе по Tab всё время сохранения, а нажатие блокируется.

Из обхода уходит только по-настоящему отключённое поле — там эта граница уже проведена именно так.

tools/check-public-api-any/baseline.json
NikGurev: **Дженерик сделали, но счётчик всё равно вырос — до 10.**

Проверили запуском самой проверки: слово `any` она считает наравне с `unknown`, поэтому замена ничего не экономит. А как только у типа появляется параметр, он дорисовывается в каждой ссылке — строка с событием стала `<any>`, это ещё плюс один.

Главное всё же получилось: касты из обоих примеров ушли. Если счётчик принципиальнее, вернём `unknown` и касты — скажите.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Исправлено, но частично.** Завели свои константы в компоненте и убрали импорт из select. Задержка стала ровно 100 мс, как и написано в документации, вместо унаследованных 101.

Выносить оператор в core не стали — договорились держать правку в границах inline edit.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Оставили три поля.** Два из них нужны по-настоящему: одно значение для отмены, другое для отката, и различаются они как раз когда редактор открывают поверх отклонённого.

Третье, отправленное, могло затираться при двух одновременных сохранениях — но их больше не бывает, так что переделка стала косметической.

packages/components/inline-edit/inline-edit.spec.ts
NikGurev: **Исправлено.** Все пороги считаются от констант, включая те, что не упоминались: `tick(150)` и три `tick(500)`. Границу по-прежнему проверяем с точностью до миллисекунды.

packages/components/inline-edit/e2e.ts
NikGurev: **Исправлено.** Кнопки-ярлыки убрали совсем. Строки теперь сохраняются так же, как это делает пользователь: клик по строке, ожидание фокуса, Enter, ожидание закрытия.

Добавили строку с меню и снимок при наведении.

packages/components/inline-edit/inline-edit.ts
NikGurev: **Откатили.** Вынесенные методы вернули на места вызова, имя `showValidationTooltipIfStillInvalid` восстановили, тип параметра вернули к `Event`.

rmnturov:
Здесь бы многострочное поле сделать
image

rmnturov (approved):

rmnturov (approved):

lskramarov (approved):

PR #2075, opened 2026-09-18
fix(table)!: errors following a full review of the component (#DS-5482)

NikGurev:
## Summary

A full review pass over `table`, in the same series as `checkbox`, `loader-overlay`, `markdown` and `progress-spinner`. The headline defect is a directive that matched far more than it was written for: `KbqTableCellContent` was declared `selector: 'kbq-table td'`, and Angular's directive selector grammar has element names, attributes, classes and `:not(…)` — but no combinators. The parser calls `setElement()` for every bare tag it meets, so the second one overwrote the first and the effective selector was plain `td`. Every template that imported `KbqTableModule` got a directive instance plus a live content query on **every cell it rendered**, whether or not the table was a `kbq-table` — and the `kbq-table` element the selector named has never existed, since the component is an attribute on a native `<table>`.

Beyond that: every color was hardcoded rather than tokenised, row hover and the pill radii stopped at `td` so a row led by a `<th>` was highlighted from its second cell on, `<tfoot>` was padded but never themed, padding and radii were physical properties so an RTL table stayed left-aligned, and a pinned header silently covered anything the browser scrolled to the top edge of the container.

Breaking, so the title carries `!`. `table-cell-content` ships with it as a **warn-only** migration: an entry in an `imports` array is deleted rather than replaced, and a stylesheet keyed on the removed modifier class has to be re-pointed at the cell, neither of which has a single mechanical translation. The `#### Table` subsection of the migration guide carries the full report in both languages.

## List of notable changes:

- **removed** `KbqTableCellContent` (TBL-BUG-01). The one rule it fed — a cell holding a button keeps the row at the button's own height — is CSS now, `.kbq-table > :is(tbody, tfoot) > tr > :is(td, th):has(.kbq-button, .kbq-button-icon)`, which also drops the `@koobiq/components/button` dependency from the table entry point
- **updated** every color to go through a `--kbq-table-*` token — `--kbq-table-header-text`, `--kbq-table-body-text`, `--kbq-table-border-color`, `--kbq-table-states-hover-background`, `--kbq-table-sticky-header-background` — and the row radius through `--kbq-table-size-row-border-radius`. No value changed — the point is that a table on a card, in a modal or in a sidepanel can now point `--kbq-table-sticky-header-background` at that surface instead of out-specifying `.kbq-table > thead > tr > th` (TBL-CSS-01)
- **fixed** row hover and the pill corner radii, widened from `td` to `:is(td, th)`, so a row led by a row header is highlighted and rounded as a unit (TBL-CSS-02)
- **added** `--kbq-table-size-sticky-header-height`, the pinned header's measured height, applied as `scroll-margin-block-start` to body cells **and to the focusable content inside them** — `scroll-margin` is read from the element the browser scrolls to, not from its cell, so a control scrolled to the top edge no longer lands under the header (WCAG 2.2 SC 2.4.11) (TBL-A11Y-01)
- **updated** cell padding and radii to logical properties and put alignment behind `--kbq-table-cell-text-align` (default `start`), so a table under `dir="rtl"` mirrors instead of staying left-aligned; `<tfoot>` gets the library color and typography it was already being padded for (TBL-STYLE-01)
- **removed** the unused `A11yModule` and `PlatformModule` imports from `KbqTableModule` (TBL-ARCH-01)
- **added** `table.spec.ts`: the modifier classes as bindings and as bare attributes, cell scoping inside and outside a `kbq-table`, sticky-header measurement and teardown, a late-mounting `<thead>`, and axe. `e2e.ts` gains a button cell, a row header, a footer, RTL and a sticky header on a card surface (TBL-TEST-01)
- **updated** documentation for the table component: `KbqTable`, `KbqTableModule` and the three inputs documented, `border` named in prose, and the sticky-header caveats written out in both languages (TBL-DOC-01); four examples de-localised and the unused `colors` field dropped from the overview example (TBL-DOC-02)

**The sticky-header background default was changed, then put back.** TBL-CSS-01 moved a pinned header from `--kbq-background-bg` to `--kbq-background-card`, reasoning that a table is more often on a card than on the page. That value was fixed by a designer, so `fix(table): keep the designer's sticky-header background` restores it and keeps only the token indirection — which is the part that was actually missing, since a table on another surface previously had to out-specify `.kbq-table > thead > tr > th`. The two tokens resolve to the same color in light theme, so the difference the original reasoning described existed in dark theme only. Note that the bodies of the first two commits still describe the old default and a warn pattern that no longer exists; the last commit names both supersessions.

**A `MutationObserver` is what makes the sticky-header measurement survive a late `<thead>`.** The effect's only tracked input is `stickyHeader()`; `elementRef.nativeElement.tHead` is a plain DOM read, so a `<thead>` mounting after first render — behind `@if`, async columns, `@defer` — would never be measured. A `childList` observer on the table's own children (not its subtree, so row insertions inside `<tbody>` don't retrigger it) bumps a counter the effect reads. The assumption is that `<thead>` is always a direct child of `<table>`, which the HTML parser guarantees. Covered by `should observe a thead that mounts after the table has already rendered`.

**What the migration deliberately does not rewrite, and what it cannot reach.** It never writes — every finding is reported. Deleting `KbqTableCellContent` from an `imports` array is not a replacement, and re-pointing a stylesheet keyed on `.kbq-table-cell_has-button` at the cell is a decision about specificity that no rewrite can make. The `--kbq-background-bg` pattern is bounded with `(?![\w-])` so it does not fire on `--kbq-background-bg-secondary`/`-tertiary`.

**Deferred: TBL-ARCH-02.** The `tools/api-gen` templates and `apps/docs` changes it calls for are outside this package, which is why the finding numbers skip it here.


packages/components/table/table.component.ts
```diff
@@ -19,22 +29,80 @@ import { KbqButton } from '@koobiq/components/button';
...
+    protected readonly stickyHeaderHeight = signal<number | null>(null);
+
+    /**
+     * Bumped whenever a direct child of the table is added, removed or reordered, so the effect below
+     * re-reads `tHead` — an otherwise untracked DOM property — when a `<thead>` mounts or is replaced
+     * after the table's first render (behind `@if`, async columns, `@defer`...). `<thead>` is always a
+     * direct child of `<table>`, so watching only the host's own child list, not its subtree, is enough
+     * and avoids rerunning on every row insertion inside `<tbody>`.
+     */
+    private readonly structuralChanges = signal(0);
```
artembelik: а какая проблема здесь решается?

NikGurev: это сайд эффект решения проблемы отображения фокусируемого контента под стики хедером

Если thead обернут в if / defer и появляется ПОСЛЕ первого рендера, то работать фокусирование  и скролл элемента в ячейках не будет, stickyHeaderHeight бесполезен

NikGurev: Это было основной фикс по ревью

NikGurev: убрал, оставил только resizeobserver
Кажется, что if / defer для thead редкий кейс и грузить этим каждую таблицу сейчас не нужно
 если что добавим

